PR #60, opened 2015-02-26
Use XMEMCMP_NEQ instead of mem_neq

pascal-brand-st-dev:
mem_neq is no more used directly. XMEMCMP_NEQ is used instead,
in the same way XMEMCMP, XMEMCPY,... are.

Signed-off-by: Pascal Brand <pascal.brand@st.com>

### DIFF
--- a/src/headers/tomcrypt_custom.h
+++ b/src/headers/tomcrypt_custom.h
@@ -45,6 +45,9 @@
    #endif
 #define XMEMCMP  memcmp
 #endif
+#ifndef XMEM_NEQ
+#define XMEM_NEQ  mem_neq
+#endif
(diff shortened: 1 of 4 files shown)


sjaeckel:
should we probably rename the macro to match the function name to avoid confusion?
XMEMCMP_NEQ sounds to me like a weird macro that uses memcmp and does some magic around it

pascal-brand-st-dev:
Sure. What name do you propose (I have no preference in mind)?

sjaeckel:
Just to match the macro XMEM_NEQ?!

pascal-brand-st-dev:
Done.

PR #63, opened 2015-04-14
Statically allocated key in hmac_state

pascal-brand38:
Signed-off-by: Pascal Brand pascal.brand@st.com

### DIFF
--- a/src/headers/tomcrypt_mac.h
+++ b/src/headers/tomcrypt_mac.h
@@ -3,7 +3,7 @@ typedef struct Hmac_state {
      hash_state     md;
      int            hash;
      hash_state     hashstate;
-     unsigned char  *key;
+     unsigned char  key[MAXBLOCKSIZE];
 } hmac_state;
 
 int hmac_init(hmac_state *hmac, int hash, const unsigned char *key, unsigned long keylen);
--- a/src/mac/hmac/hmac_done.c
+++ b/src/mac/hmac/hmac_done.c
@@ -89,7 +89,6 @@ int hmac_done(hmac_state *hmac, unsigned char *out, unsigned long *outlen)
 
     err = CRYPT_OK;
 LBL_ERR:
-    XFREE(hmac->key);
 #ifdef LTC_CLEAN_STACK
     zeromem(isha, hashsize);
     zeromem(buf,  hashsize);
--- a/src/mac/hmac/hmac_init.c
+++ b/src/mac/hmac/hmac_init.c
@@ -55,13 +55,6 @@ int hmac_init(hmac_state *hmac, int hash, const unsigned char *key, unsigned lon
        return CRYPT_MEM;
     }
 
-    /* allocate memory for key */
-    hmac->key = XMALLOC(LTC_HMAC_BLOCKSIZE);
-    if (hmac->key == NULL) {
-       XFREE(buf);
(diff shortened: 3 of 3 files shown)


tomstdenis:
The downside though is you increase the stack usage of anyone who put hmac_state on their stack.  They could malloc/free it but that was handled internally originally (which is more ideal).

If you're worried about the malloc/free cycle with the init/done you could add a done_and_reset function which done calls with a flag saying "free" meanwhile others can call it with "don't free" and then immediately jump back to "process" a new message.

pascal-brand-st-dev:
On one hand, I agree.

But on the other hand, if you look at omac_state (and others), we have the same scheme: omac_state contains several static arrays. So the current pull-request makes the code more unified.

Indeed, we try to minimize the heap. So minimizing malloc/free is fine for us. The solution you proposed would not fit in our case I think. However, what could be ok for us is to compile one solution or another based on a new LTC_xxx definition. I can propose something that could limit the impact on the code if you're interested.

tomstdenis:
That makes sense (to be honest I haven't look at LTC code in a really long time).  The downside to having static arrays though is that if a hash comes around with a larger block size it could cause problems requiring a LTC recompile.  Which isn't the end of the world but isn't a design goal of the project [at least originally].

Given that the other MACs used static arrays I think it's appropriate to change the hmac code over.

sjaeckel:
> The downside though is you increase the stack usage of anyone who put `hmac_state`

TBH that was also my first concern.
But as `hmac_state` already contains two instances of `hash_state` it's not that big of a difference to have another 124 resp. 120 bytes on the stack...

> The downside to having static arrays though is that if a hash comes around with a larger block size it could cause problems requiring a LTC recompile.

Valid reason why we shouldn't do that, but as the allocation uses a Macro for its size a recompile would be required in any case.

> Which isn't the end of the world but isn't a design goal of the project [at least originally].

Well, the layout with static arrays was always there and has not been touched.
The only MAC that used dynamic allocation was HMAC...

So there would be two approaches to have the same scheme for all MAC's:

1. Apply that patch
2. Modify all other MAC's to use dynamic allocation

In regards to ABI breakage, for the case of a hash with larger block size coming around we would have to do approach 2...

pascal-brand-st-dev:
Another implementation can be found in #68. This allows to choose, at compile time, between static and dynamic version.

karel-m:
I am perhaps a bit late to this discussion but looking at #63 (this PR) and alternative #64 I support #63 
* consistency - as mentioned above it is consistent with other mac algs
* it is simple and clear - in fact the final code is even simpler (compared to the newly introduced set of macros in #64)
* I am afraid of future maintenance costs if we would apply #64 (and you know, we are not that good at maintenance)
* Also properly testing both new branches (static/dynamic) is an extra complexity that will consume our limited resources
* I understand that wasting the stack space is not good. But frankly speaking I do not see so many request/reports/bugs/pulls demanding lower stack usage. Once they appear, let us solve them.

jforissier:
FYI, OP-TEE also went the "static" route (see https://github.com/OP-TEE/optee_os/commit/947ebd697b79) so I suppose https://github.com/libtom/libtomcrypt/pull/64 can be closed (at least, OP-TEE won't be pushing for it to be merged).
Thanks!

PR #64, opened 2015-04-15
LTC_MAC_STATIC introduction to choose between static / dynamic

pascal-brand38:
LTC_MAC_STATIC:
- When defined, static arrays are defined in mac structures.
- When not defined, pointers are declared. Dynamic allocation is then
  performed in the related xxx_init() and xxx_done() functions

This leverages comments made on #63

Signed-off-by: Pascal Brand pascal.brand@st.com


sjaeckel:
looks good to me
+ can easily be used for all the other MAC's
+ uses the dynamic version by default

pascal-brand-st-dev:
Comments have been taken into account.

tomstdenis:
Are there plans to circle back to at least the xcbc/cmac implementations to support this?

sjaeckel:
> Are there plans to circle back to at least the xcbc/cmac implementations to support this?

Should we probably apply that approach in all places where statically sized arrays are used? That would be prng, mac, hash, cipher... or is that too much?

tomstdenis:
Normally inside functions scratch buffers shouldn't be considered unless the total size of them exceeds a couple 100 or so bytes.  So a 16 byte buffer in a cipher is fine.  Outside functions (say state structures) the threshold can be smaller because in theory they're going to be malloc/freed less often.

The flaw in the API here is that we can't reuse [say] an hmac_state structure once it's been initialized since the call to done will free memory.  So any sort of packet processing is going to be performing needless heap operations.  

So the solutions could be 

1.  Always use static and force the user to static/malloc the state themselves
2.  Allow build-time malloc and then the user can do whatever they want with the state
3.  Always do malloc

#2 is diplomatic but causes more complexity inside LTC which may increase the test space

#3 is a performance hazard but avoids stack issues completely.

Giving this some though over the last couple of days I'd prefer #1 and then instruct users they could malloc the state if their stack size is limited.  always doing static also means you don't need any API upgrades to have "reset" type functionality.  The user can simply re-use the struct over and over.  the only performance hazard is the memcpy() of keying material.

sjaeckel:
Thanks, that's a nice summary.

> The flaw in the API here is that we can't reuse [say] an hmac_state structure once it's been initialized since the call to done will free memory. So any sort of packet processing is going to be performing needless heap operations.

That makes me immediately think about your first idea:

> If you're worried about the malloc/free cycle with the init/done you could add a done_and_reset function which done calls with a flag saying "free" meanwhile others can call it with "don't free" and then immediately jump back to "process" a new message.

How about introducing something like that? say, we keep the existing behavior, we someone calls `done()` it free's the memory, but we provide a new API function `reset()` (or something similar) ... that basically does everything like `init()` besides the allocation?

Personally I would also prefer to have everything statically allocated and the user decides himself if he wants to malloc the struct, put it on the stack or static somewhere... but the main reason against it is the ABI breakage in case "a {prng, mac, hash, cipher} comes around with a larger block size".

I would also like to keep the compile time option to just allocate everything statically in the struct, but by default (e.g. for the version of ltc in OS distributions) use the dynamic version.

pascal-brand-st-dev:
@sjaeckel Should the current PR be merged?  Or do you want me to introduce any other change?

pascal-brand-st-dev:
Any feedback?

karel-m:
as mentioned in #63 I am against this PR introducing more complexity 👎

sjaeckel:
Closed because it's too complex

Issue #34, opened 2014-01-15
ECC improvements


karel-m:
Hi,

I would like to propose a set of changes to ECC part of libtomcrypt. I have already implemented all proposed changes in my something-like-fork of libtomcrypt which is available https://github.com/DCIT/perl-CryptX/tree/master/src/ltc
## 1/ Support for curves y^2 = x^3 + a*x + b

Currently libtomcrypt supports only y^2 = x^3 - 3*x + b. 

Support for general case required changes to ltc_ecc_projective_dbl_point(), ltc_ecc_projective_add_point() and ltc_ecc_is_point() as you can see in my repo (please note that I have moved is_point function into a separate file ltc_ecc_is_point.c). Extending ltc_ecc_set_type + other small changes were also necessary.
## 2/ Support for compressed public keys

Already existing function ecc_ansi_x963_export() allows you to export EC public key in a format defined by X9.63 but only supports 'uncompressed' form.

I have added support for exporting compressed EC public keys (which was easy) as well as for import. The import part was a bit tricky as it was necessary to add mp_sqrtmod_prime to libtommath.

As I wanted to keep old interface I have created new functions:
- https://github.com/DCIT/perl-CryptX/blob/master/src/ltc/pk/ecc/ecc_export_raw.c
- https://github.com/DCIT/perl-CryptX/blob/master/src/ltc/pk/ecc/ecc_import_raw.c

NOTE: The new functions ecc_import_raw() / ecc_export_raw() also support import/export of private key (raw octects)

My implementation of mp_sqrtmod_prime (based on Tonelli–Shanks algorithm) is available here:
- https://github.com/DCIT/perl-CryptX/blob/master/src/ltm/bn_mp_sqrtmod_prime.c
## 3/ DER format compatible with openssl

To support importing/exporting keys in DER format compatible with openssl we need to support the following:
- private key format: http://tools.ietf.org/html/rfc5915
- public key format: http://tools.ietf.org/html/rfc5480

Again, I wanted to keep old interface so I have created new functions:
- https://github.com/DCIT/perl-CryptX/blob/master/src/ltc/pk/ecc/ecc_import_full.c
- https://github.com/DCIT/perl-CryptX/blob/master/src/ltc/pk/ecc/ecc_export_full.c

[...]

PR #65, opened 2015-04-17
ECC enhancements

karel-m:
This PR is related to https://github.com/libtom/libtomcrypt/issues/34


sjaeckel:
I think the build error can be fixed in several ways:

1. by making point compression optional and disabling it by default
2. creating a new release of libtommath with the required additional algorithm, waiting until it's published
3. building libtommath from the git repositories
4. combining 1&3 and creating a special recipe with point compression enabled and build ltm from git

I would prefer going for 2 or 4

karel-m:
Or we can introduce in libtommath something like LTM_VERSION (e.g. 0x0042 for version 0.42) and then in libtomcrypt's `ltm_desc.c`:

```c
/* sqrtmod_prime */
static int sqrtmod_prime(void *a, void *b, void *c)
{
   LTC_ARGCHK(a != NULL);
   LTC_ARGCHK(b != NULL);
   LTC_ARGCHK(c != NULL);
#if !defined(LTM_VERSION) || LTM_VERSION < 0x0042
   return CRYPT_LTM_TOO_OLD; 
#else
   return mpi_to_ltc_error(mp_sqrtmod_prime(a, b, c));
#endif
}
```

karel-m:
@fperrad may I ask you for running your linting machine on `miko-ecc-enhancements` branch? Thanks.

fperrad:
@karel-m linting issues:

missing prototypes in tomcrypt.h : ecc_sign_hash_ex & ecc_verify_hash_ex
```
Info 765: external 'ecc_sign_hash_ex(const unsigned char *, unsigned long, unsigned char *, unsigned long *, union Prng_state *, int, struct ecc_key *, int)' (line 24, file src/pk/ecc/ecc_sign_hash.c) could be made static
Info 765: external 'ecc_verify_hash_ex(const unsigned char *, unsigned long, const unsigned char *, unsigned long, int *, struct ecc_key *, int)' (line 34, file src/pk/ecc/ecc_verify_hash.c) could be made static
```
bad indentation line 30 in src/pk/ecc/ecc_export_raw.c

bad indentation line 62 in src/pk/ecc/ecc_sign_hash.c

bad indentation line 72, 74 & 75 in src/pk/ecc/ecc_import_raw.c

sjaeckel:
can we somehow bring the coverage up again?

karel-m:
> can we somehow bring the coverage up again?

What has to be done to bring it up?

sjaeckel:
1. These aren't tested at all

        ecc_sign_hash_rfc7518()
        ecc_verify_hash_rfc7518()

1. There are no tests for all the added functionality

        pk_get_oid()

1. The new `tag` field in DER/`ltc_asn1_list`

1. The case `a==NULL` in some of the code paths that are handled specially

karel-m:
Oh yes, tests, I know.

src/pk/asn1/der/sequence/der_encode_sequence_ex.c
```diff
@@ -207,141 +88,145 @@ int der_encode_sequence_ex(ltc_asn1_list *list, unsigned long inlen,
...
-               x       += z;
-               *outlen -= z;
                break;
 
            case LTC_ASN1_CHOICE:
            case LTC_ASN1_CONSTRUCTED:
            case LTC_ASN1_CONTEXT_SPECIFIC:
            case LTC_ASN1_EOL:
            case LTC_ASN1_TELETEX_STRING:
+           default:
```
sjaeckel: I never use a default case when I switch on an enum as the compiler then warns me if I add elements to the enum

sjaeckel: I prefer not to use a default case when I switch on an enum as the compiler then warns me if I add elements to the enum and forget to modify the switch-case

karel-m:
@sjaeckel this PR is too complex (rebasing is quite often a pain), I am considering the idea of splitting this PR to two parts: 1/ ASN.1 changes + 2/ EC-only stuf. 

The part 1/ would be a new PR (which I want to merge to develop before continuing on the remaining EC-only part) comprising of approx. ede958b + subset of 01c319d 

What do you think?

sjaeckel:
> ...splitting this PR into two parts: 1/ ASN.1 changes + 2/ EC-only stuf.

Very good idea, that also makes review easier. Can the EC part probably be split even more?

karel-m:
> Can the EC part probably be split even more?

We'll see, let's start with ASN.1 part

sjaeckel:
We need a ltm-1.0 package backported to trusty to fix these build failures.
I had to remove the ltm from sid in #217 as I couldn't update clang otherwise...

karel-m:
I have split this PR and my branch `miko-ecc-enhancements` to:
* #236 ECC enhancements part 1 (curves y^2 = x^3 + ax + b) - branch `pr/ecc-non-asn1-part`
* #187 ECC enhancements part 2 (ASN.1 changes) - branch `pr/ecc-asn1-part` (based on `pr/ecc-non-asn1-part`)

So basically current `miko-ecc-enhancement` is the same as `pr/ecc-asn1-part`

There fore I am closing this PR, please move futher discussion to #236 + #187

PR #67, opened 2015-06-28
fix crypt.tex with newer LaTeX: s/here/h/g

stapelberg:
See https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=790321 for context


karel-m:
merged into https://github.com/libtom/libtomcrypt/tree/feature/doc

closing

PR #70, opened 2015-07-08
Add function pointers for accelerated XTS to ltc_cipher_descriptor

jforissier:
Similar to what already exists for other modes.

Signed-off-by: Jerome Forissier <jerome.forissier@linaro.org>

### DIFF
--- a/src/modes/xts/xts_decrypt.c
+++ b/src/modes/xts/xts_decrypt.c
@@ -87,22 +87,36 @@ static int tweak_uncrypt(const unsigned char *C, unsigned char *P, unsigned char
       return CRYPT_INVALID_ARG;
    }
 
-   /* encrypt the tweak */
-   if ((err = cipher_descriptor[xts->cipher].ecb_encrypt(tweak, T, &xts->key2)) != CRYPT_OK) {
-      return err;
-   }
-
-   /* for i = 0 to m-2 do */
    if (mo == 0) {
       lim = m;
    } else {
       lim = m - 1;
    }
 
-   for (i = 0; i < lim; i++) {
-      err = tweak_uncrypt(ct, pt, T, xts);
-      ct += 16;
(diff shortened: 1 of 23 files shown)


src/modes/xts/xts_decrypt.c
```diff
@@ -87,24 +88,39 @@ static int tweak_uncrypt(const unsigned char *C, unsigned char *P, unsigned char
...
+	   /* use accelerated decryption for whole blocks */
+	   if ((err = desc->accel_xts_decrypt(ct, pt, lim, tweak, &xts->key1,
+					      &xts->key2) != CRYPT_OK)) {
+	      return err;
+	   }
+	   ct += lim * 16;
+	   pt += lim * 16;
+
+	   /* tweak is encrypted on output */
+	   memcpy(T, tweak, sizeof(T));
```
sjaeckel: XMEMCPY

src/modes/xts/xts_decrypt.c
```diff
@@ -87,24 +88,39 @@ static int tweak_uncrypt(const unsigned char *C, unsigned char *P, unsigned char
...
       lim = m;
    } else {
       lim = m - 1;
    }
 
-   for (i = 0; i < lim; i++) {
-      err = tweak_uncrypt(ct, pt, T, xts);
-      ct += 16;
-      pt += 16;
+   desc = &cipher_descriptor[xts->cipher];
```
sjaeckel: Ah I don't know what I should say... I generally also like it that way, but somehow I don't think we should change the code style only in one file... so please change the pointer back to how it was before :)

sjaeckel:
Thanks for the PR, generally looks fine for me.
Comments are okay I hope.

jforissier:
@sjaeckel thanks for reviewing the patch. New version pushed.

Issue #73, opened 2015-08-19
ccm_memory


sebastianv89:
About the libtom [implementation](https://github.com/libtom/libtomcrypt/blob/master/src/encauth/ccm/ccm_memory.c) of authenticated encryption mode CCM:

According to the [NIST specification of CCM](http://csrc.nist.gov/publications/nistpubs/800-38C/SP800-38C_updated-July20_2007.pdf), the authentication tag is part of the ciphertext. In order to decrypt, this full ciphertext must be decrypted, resulting in a "plaintext" tag. The tag must then be recomputed upon the plaintext and compared with the decrypted value. However, upon decryption in the libtom implementation, the ciphertext is decrypted, and a tag is computed upon the header and (decrypted) plaintext. This is then re-encrypted, so that the caller of the function must compare the resulting (encryped) tag with the received (encrypted) tag.

The NIST specification specifies that "only the error message INVALID is returned" when the decryption-verification fails. In that case "the payload P and the MAC T shall not be revealed" and that "the implementation shall ensure that an unauthorized party cannot distinguish whether the error message results from [invalid message format] or from [authentication failure], for example, from the timing of the error message."

There are several fixes (I think), for example removing the tag parameter from the function and let it be part of the ciphertext parameter (as it should be according to the specification). Verification of authentication can then be done inside the function ccm_memory. Of course, this would break compatibility with the existing implementation, so I'm not sure if that is what you want.

sebastianv89:
I proposed a fix, but unfortunately it is impossible to write a correct implementation according to NIST that is still compatible with the previous implementation. The reason is that the tag parameter is an output parameter, whereas it should be an input parameter in case of decryption.

sjaeckel:
I've pushed [fix/ccm_constant_time](https://github.com/libtom/libtomcrypt/tree/fix/ccm_constant_time)
What do you think?
It'd be cool if you could set your branch to that commit and force-push, so it will also be visible in the PR.

PR #75, opened 2015-08-21
Quickfix for issue #73

sebastianv89:
The API of the function is changed (for decryption, tag is now an input
parameter). With the old API it is impossible to confirm to the NIST
specification and a timing sidechannel leak is inevitable.


sebastianv89:
See pull request #76

PR #76, opened 2015-08-21
ccm_memory timing issues

sebastianv89:


sebastianv89:
I looked at the fix I provided in 3120e32, but it was not necessary. The only thing that the was leaking in commit 6c7a393 was the value of `err`, which was not secret anyway. If your code is faster (which I believe it is), you might consider rolling back to that commit. Modulo the fixes to the testing library of course.

karel-m:
What needs to be done here?

sjaeckel:
> What needs to be done here?

review if we prefer the version of 3120e32b48e2b7c714dea21ff1a6a0c1825e7bab or 6c7a3931ef3c3897d31530d2dbe9d6a1a3f332ba

I'd say it doesn't really matter... IIUC it boils down to speed vs. space... 6c7a3931ef3c3897d31530d2dbe9d6a1a3f332ba should be faster, but requires twice the memory on the heap.

...we could keep both implementations and switch depending on `LTC_SMALL`...

rofl0r:
did you test if  6c7a393 is really a lot faster ? if it's less than 50% faster i'd say it's not worth keeping it, if it has twice the memory cost (also: the more malloc, the more possible memory leaks, free-after-use, etc)

PR #77, opened 2015-09-08
Implement RSA-CRT hardening

sjaeckel:
Please implement RSA-CRT hardening, to avoid leaking RSA private keys in case there is a computation error (hardware fault or bignum library implementation problem) during a signature computation involving the RSA-CRT optimization:

* https://securityblog.redhat.com/2015/09/02/factoring-rsa-keys-with-tls-perfect-forward-secrecy/
* https://people.redhat.com/~fweimer/rsa-crt-leaks.pdf

### DIFF
--- a/src/headers/tomcrypt_custom.h
+++ b/src/headers/tomcrypt_custom.h
@@ -402,6 +402,11 @@
 #define LTC_RSA_BLINDING
 #endif  /* LTC_NO_RSA_BLINDING */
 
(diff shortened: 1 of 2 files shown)


sjaeckel:
Something like that?

fweimer:
Yes, looks good.

sjaeckel:
Cool, thanks for notify and review.

PR #78, opened 2015-09-08
add ltc_rng function pointer

sjaeckel:
the idea is to be able to easily provide a plug-in rng for a specific
platform without the need to touch the library.

any comments/ideas?

### DIFF
--- a/libtomcrypt_VS2005.vcproj
+++ b/libtomcrypt_VS2005.vcproj
@@ -1098,6 +1098,10 @@
 					>
 				</File>
 				<File
+					RelativePath="src\misc\crypt\crypt_prng_rng_descriptor.c"
+					>
+				</File>
+				<File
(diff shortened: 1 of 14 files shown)


karel-m:
What is the use case for this?

IMO introducing the global `ltc_rng` is asking for thread safeness troubles.

sjaeckel:
> What is the use case for this?

I had embedded devices in mind (and used it for that purpose)

> IMO introducing the global ltc_rng is asking for thread safeness troubles.

Yep you're right, it's up to the implementation of the ltc_rng to handle this kind of issue... probably we should disable it by default...

karel-m:
Introducing something like `-DLTC_ENABLE_RNG_HOOK` which will enable this feature would be great as I expect that vast majority of users will not need it.

PR #83, opened 2015-12-03
Add OpenSSL-Compatible PKCSv5#1 Variant, Symmetric Crypto Demo, Docs

bblack-cuda:
Hi all!

Big fan of libtomcrypt here :-)

The PR below adds the following capabilities to libtomcrypt:
1. A key-derivation function compatible with OpenSSL's variable-length PKCS#5v1 variant (i.e. it creates keys compatible with EVP_BytesToKey()).
2. Minor header and doc changes for the new function.
3. A fairly comprehensive demo of a command-line crypt program that's compatible with "openssl enc -aes-256-cbc", including header/salt handling, PKCS#7-type padding, and the KDF I added above.

Testing has included a bunch of encrypt/decrypts of files ala:

```
# ./openssl-enc enc /bin/ls foo.bin foobar
...
# openssl enc -aes-256-cbc -d -pass pass:foobar -in foo.bin -out foo.out
```

I've tried a bunch of files of various kinds and with various sizes (some even multiples of blocksize, some not), both back and forth from openssl and it seems to work well.  Every time the encrypted files came out byte-perfect (assuming the same salt) between the two and the decrypted files came out byte-perfect and sized correctly.

I tried to tread lightly and keep most of the code close to the style and idioms observed, but feedback is very welcome :-)

I assume that my changes will be licensed the same as ltc (IIRC that's WTFPL + public domain) and hereby grant permission for the same.

### DIFF
--- a/src/misc/pkcs5/pkcs_5_1.c
+++ b/src/misc/pkcs5/pkcs_5_1.c
@@ -16,7 +16,17 @@
 */
 #ifdef LTC_PKCS_5
 /**
-   Execute PKCS #5 v1
+   Execute PKCS #5 v1 in strict or OpenSSL EVP_BytesToKey()-compat mode.
+
+   PKCS#5 v1 specifies that the output key length can be no larger than
+   the hash output length.  OpenSSL unilaterally extended that by repeating
+   the hash process on a block-by-block basis for as long as needed to make
+   bigger keys.  If you want to be compatible with KDF for e.g. "openssl enc",
+   you'll want that.
+
+   If you want strict PKCS behavior, turn openssl_compat off.  Or (more
+   likely), use one of the convenience functions below.
+
    @param password         The password (or key)
    @param password_len     The length of the password (octet)
    @param salt             The salt (or nonce) which is 8 octets long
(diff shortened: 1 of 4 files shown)


sjaeckel:
cool! looks basically fine for me

some changes I'd like to have please:
- don't use camelCase, but snake_case
- :-)
```
function signatures(look
                    like this);
  not like(
    this
  );
```
- squash your fixup commits in the original ones and rebase on develop afterwards

bblack-cuda:
OK, I think that did it.  I noticed that Travis CI is failing because it can't figure out how to apt-get install libtommath-dev, but I think it's otherwise OK.

src/misc/pkcs5/pkcs_5_1.c
```diff
@@ -16,25 +16,42 @@
...
    @param out              [out] The destination for this algorithm
    @param outlen           [in/out] The max size and resulting size of the algorithm output
+   @param openssl_compat   [in] Whether or not to grow the key to the buffer size ala OpenSSL
    @return CRYPT_OK if successful
 */
-int pkcs_5_alg1(const unsigned char *password, unsigned long password_len,
-                const unsigned char *salt,
-                int iteration_count,  int hash_idx,
-                unsigned char *out,   unsigned long *outlen)
+int pkcs_5_alg1_common(const unsigned char *password,
```
sjaeckel: guessing in the wild that this could/should be static

karel-m:
@sjaeckel is right about `static` - I suggest changing:
```
- int pkcs_5_alg1_common
+ static int _pkcs_5_alg1_internal
```

The style `static sometype _some_func_internal` we already use at other places.

Otherwise I agree with this patch.

sjaeckel:
> The style static sometype _some_func_internal we already use at other places.

We use this exact format in b64, but nowhere else... I like to prefix static functions with `_` and that's it. In general I go with what the author likes, that's why I didn't say anything of the `_some_func_internal` format in b64 :-)

...just make it static, rename it as you wish...

karel-m:
closing this one, for extra comments please use #161

PR #88, opened 2016-01-02
HP-UX related fixes

karel-m:
Hi,

please find enclosed patch that is neccessary to compile libtomcrypt on HP-UX.

There are two parts:
- some more or less expected stuff in `tomcrypt_cfg.h`
- a workaround for HP C compiler that does not like `static inline int func_name(...)` (the `inline` part is the problem)

--Karel

### DIFF
--- a/src/headers/tomcrypt_cfg.h
+++ b/src/headers/tomcrypt_cfg.h
@@ -43,6 +43,15 @@ LTC_EXPORT int   LTC_CALL XSTRCMP(const char *s1, const char *s2);
 
 #endif
 
+/* some compilers do not like "inline" */
+#if defined(__HP_cc)
+   #define LTC_INLINE
+#elif defined(_MSC_VER)
+   #define LTC_INLINE __inline
+#else
+   #define LTC_INLINE inline
+#endif
+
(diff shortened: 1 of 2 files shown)


sjaeckel:
I'd prefer to use something like that:
```
#if defined(__hpux) || defined(__hpux__)
#define LTC_INLINE
#else
#define LTC_INLINE inline
#endif
```
and then changing the `inline` keywords in the function definitions to `LTC_INLINE`

karel-m:
The part of this patch related to `tomcrypt_cfg.h` was superseded by https://github.com/libtom/libtomcrypt/tree/fix/109-alternative-approach

I'll update this PR to only handle `static inline` issue

```
/* some compilers do not handle "static inline" */
#if defined(__HP_cc)
#define LTC_STATIC_INLINE static
#else
#define LTC_STATIC_INLINE static inline
#endif
```

karel-m:
So now simplified to "static inline" only fix

Ready for review/merge

sjaeckel:
why not the way I proposed above?

it's only about the inline which is specific so I think we should only pull that into the define.

karel-m:
You are right, I'll fix it

Issue #92, opened 2016-01-03
camellia.c const is meaningless on cast type


karel-m:
I have experienced the following compiler warning:
```
"src/ltc/ciphers/camellia.c", line 193: warning #2191-D: type qualifier is
          meaningless on cast type
     U = D ^ ROR(U, (const int)8);
                     ^
```

Why not just use: `U = D ^ ROR(U, 8);` ?

sjaeckel:
well we could use RORc at that place as it's a rotate by a constant amount of bits, but the cast doesn't make sense.
can you try if `RORc(U, 8)` works for you? I just tested on x84_64 and it's fine

sjaeckel:
that just made me think of 62878de0c5dbb9f89474590d953bbdb339bd2f76 when I saw the PPC `ROXc()` functions

karel-m:
changing to `RORc(U, 8)` works

PR #96, opened 2016-01-10
fix for issue #92 - const is meaningless on cast type

karel-m:


sjaeckel:
Can't we use the `RORc` macro instead of `ROR`?

karel-m:
RORc is fine

PR #106, opened 2016-01-23
_base64_decode_internal not compliant with RFC4648

sjaeckel:
The _base64_decode_internal function is not compliant with RFC4648. RFC4648 states in 3.3:

Implementations MUST reject the encoded data if it contains
characters outside the base alphabet when interpreting base-encoded
data, unless the specification referring to this document explicitly
states otherwise.

Please find attached a version which is compliant with 3.3 of RFC4648. If there is a need for this behaviour as part of other referencing standards then an appropriate pre-processor directive or configuration parameter should be added.

[base64_decode.txt](https://github.com/libtom/libtomcrypt/files/99008/base64_decode.txt)

### DIFF
--- a/src/headers/tomcrypt_custom.h
+++ b/src/headers/tomcrypt_custom.h
@@ -476,6 +476,11 @@
    #define LTC_PKCS_1
 #endif
 
+#if (defined(LTC_BASE64) || defined(LTC_BASE64_URL)) && !defined(LTC_BASE64_STRICT)
+   /* By default we're doing strict decoding now */
+   #define LTC_BASE64_STRICT 1
+#endif
+
 #if defined(TFM_DESC) && defined(LTC_RSA_BLINDING)
     #warning RSA blinding currently not supported in combination with TFM
     #undef LTC_RSA_BLINDING
--- a/src/headers/tomcrypt_custom.h
+++ b/src/headers/tomcrypt_custom.h
@@ -476,11 +476,6 @@
    #define LTC_PKCS_1
 #endif
 
-#if (defined(LTC_BASE64) || defined(LTC_BASE64_URL)) && !defined(LTC_BASE64_STRICT)
-   /* By default we're doing strict decoding now */
-   #define LTC_BASE64_STRICT 1
-#endif
-
 #if defined(TFM_DESC) && defined(LTC_RSA_BLINDING)
     #warning RSA blinding currently not supported in combination with TFM
     #undef LTC_RSA_BLINDING
--- a/src/headers/tomcrypt_misc.h
+++ b/src/headers/tomcrypt_misc.h
@@ -3,18 +3,20 @@
 int base64_encode(const unsigned char *in,  unsigned long len,
                         unsigned char *out, unsigned long *outlen);
 
-#define base64_decode(i, il, o, ol) base64_decode_ex(i, il, o, ol, LTC_BASE64_STRICT)
-int base64_decode_ex(const unsigned char *in,  unsigned long len,
-                        unsigned char *out, unsigned long *outlen, int strict);
+int base64_decode(const unsigned char *in,  unsigned long len,
+                        unsigned char *out, unsigned long *outlen);
+int base64_strict_decode(const unsigned char *in,  unsigned long len,
+                        unsigned char *out, unsigned long *outlen);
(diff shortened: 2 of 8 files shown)


sjaeckel:
fine like that?

iainmeikle:
Looks good to me after a very quick glance. Thank you very much for your time and effort in resolving this. Do you want me to close the issue?

karel-m:
Sorry for my late response.

I did try the changes proposed in #106 but frankly, I had quite hard times with them.

1/ I am pretty sure that having a new build option LTC_BASE64_STRICT is not a good idea. It will IMO inevitably end up with different libtomcrypt builds around with decode_base64 behaving differently so that if you want to have control over what it is actually doing you will have to use decode_base64_ex anyway. I think that build options/defines should be used only for including/excluding some parts of libtomcrypt not changing the behaviour.

2/ providing that the old `decode_base64` was sort of relaxed implementation it would be less surprising to introduce `decode_base64_strict` 

3/ according to my testing current #106 implementation does not accept white spaces `\n`, `\r`, `\t` and ` ` if the strictness is turned off. Which is probably fine if we want to strictly follow RFC but the old/current `decode_base64` happily ignores whitespaces (which is handy).

sjaeckel:
> Sorry for my late response.

No worries.

> 1/ ... I think that build options/defines should be used only for including/excluding some parts of libtomcrypt not changing the behaviour.

True, I'll change that.

> 2/ providing that the old decode_base64 was sort of relaxed implementation it would be less surprising to introduce decode_base64_strict

Fine by me. (this also solves 3)

It's `base64_decode()` so `base64_strict_decode()` okay?

karel-m:
`base64_strict_decode()` is absolutely clear and ok

karel-m:
The latest changes looks fine.

I only think that this line should be implemented as more forgiving in relaxed mode:
```c
if (y == 1 || map != map_base64url || is_strict == 1) return CRYPT_INVALID_PACKET;
```

My interpretation on "relaxed" mode is that the trailing "=" are completely optional for both `base64url` and `base64`.

For example:
```c
if ((y == 1 || map != map_base64url) && is_strict == 1) return CRYPT_INVALID_PACKET;
```

sjaeckel:
> My interpretation of "relaxed" mode is that the trailing "=" are completely optional for both base64url and base64.

May I ask you to provide a patch that implements your interpretation?
And also supports these new testcases? (At least the first 3 would be mandatory)

```
diff --git a/testprof/base64_test.c b/testprof/base64_test.c
index 40c3ea9..ddefba8 100644
--- a/testprof/base64_test.c
+++ b/testprof/base64_test.c
@@ -40,6 +40,12 @@ int base64_test(void)
          {"vuiSPKIl8PiR5O-rC4z9_xTQKZ0=", 1},
          {"vuiS*PKIl8P*iR5O-rC4*z9_xTQKZ0", 0},
          {"vuiS*PKIl8P*iR5O-rC4*z9_xTQKZ0=", 0},
+         {"vuiS*PKIl8P*iR5O-rC4*z9_xTQKZ0==", 0},
+         {"vuiS*PKIl8P*iR5O-rC4*z9_xTQKZ0===", 0},
+         {"vuiS*PKIl8P*iR5O-rC4*z9_xTQKZ0====", 0},
+         {"vuiS*=PKIl8P*iR5O-rC4*z9_xTQKZ0=", 0},
+         {"vuiS*==PKIl8P*iR5O-rC4*z9_xTQKZ0=", 0},
+         {"vuiS*===PKIl8P*iR5O-rC4*z9_xTQKZ0=", 0},
    };
 
    for (x = 0; x < sizeof(cases)/sizeof(cases[0]); ++x) {
```
 
Thanks

karel-m:
Trouble seems to be with `base64_decode` and for example `Zm9vYmE=` vs. `Zm9vYmE` which I expect both to decode in relaxed mode to `fooba`.

I'll try to prepare add my fix to `fix/106`

karel-m:
My last commit b16e5dab43 should be correct (at least it passes all my relaxed scenarios).

karel-m:
Providing that we more or less keep the old relaxed `base64_decode` there is IMO no need to wait with this fix for 2.0 milestone.

PR #124, opened 2016-07-10
fix tiny compile error in tomcrypt_pk.h macro

zeromus:
An ARM compiler gives me this: 

libtomcrypt\pk\asn1\der\sequence\der_decode_subject_public_key_info.c(65,4): error #188-D: enumerated type mixed with another type

Since der_decode_subject_public_key_info's parameters_type is of type 'unsigned long', an attempt to assign it to ltc_asn1_list's member 'ltc_asn1_type type' fails.

My fix solves this in a simple way by casting it at the point of assignment.

But while studying this I noticed there's no use of enum in the codebase other than a few PK-related things.  Perhaps a more appropriate solution would be to remove these enums. I mean, enums seem like an OK enough idea, but I don't know anything about the practicality of using enums in archaic C dialects like libtomcrypt conforms (thankfully!) to...

### DIFF
--- a/src/headers/tomcrypt_pk.h
+++ b/src/headers/tomcrypt_pk.h
@@ -487,7 +487,7 @@ typedef struct ltc_asn1_list_ {
    do {                                              \
       int LTC_MACRO_temp            = (index);       \
       ltc_asn1_list *LTC_MACRO_list = (list);        \
-      LTC_MACRO_list[LTC_MACRO_temp].type = (Type);  \
+      LTC_MACRO_list[LTC_MACRO_temp].type = (ltc_asn1_type)(Type);  \
       LTC_MACRO_list[LTC_MACRO_temp].data = (void*)(Data);  \
       LTC_MACRO_list[LTC_MACRO_temp].size = (Size);  \
       LTC_MACRO_list[LTC_MACRO_temp].used = 0;       \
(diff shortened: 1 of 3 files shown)


sjaeckel:
The error is somehow correct, but I don't like the patch.

The correct way would be by casting `parameters_type` to `ltc_asn1_type` in pk\asn1\der\sequence\der_decode_subject_public_key_info.c:65

The same in der_encode_subject_public_key_info.c:53

zeromus:
My way, it's at least recorded somewhere that the expected type of `Type` in the macro is an ltc_asn1_type.  I still think it would be better to ditch this one enum when the enums arent used anywhere else. But as you wish, here it is done your way

sjaeckel:
> My way, it's at least recorded somewhere that the expected type of `Type` in the macro is an ltc_asn1_type.

My way the compiler reminds you and not some deeply hidden typecast...

> I still think it would be better to ditch this one enum when the enums arent used anywhere else. But as you wish, here it is done your way

I think we should use the enum value as well in the API Functions.

zeromus:
OK, I'll do it yet again through the API. I still think it's weird to use enums in this one place and nowhere else. Does it just so happen it doesnt make sense anywhere else, or is this part of a new masterplan to use enums all over?

sjaeckel:
Noooo :-D Don't do that now, I only meant that we should do it, not that we do it now.

And if we do it, we do it all-over.... and also it'd make sense before having a 2.0 out...

But that should be another PR

PR #125, opened 2016-07-14
Fix to make xcode toolchain happy.

MMI:
[Housekeeping from open-sourcing BitKeeper]

BitKeeper developers on Macs using current xcode compilers would see things like:

```
/usr/include/tomcrypt_macros.h: In function 'RORc':
/usr/include/tomcrypt_macros.h:275: error impossible contraint in 'asm' 
```

In the BitKeeper copy of the code, we have been carrying around a change c/o wscott@bitkeeper.com, with the following comment:

   x86 only allows variable shifts and rotates with the CL register.
  Change the inline asm to use CL explicitly and update the constraints to
  reflect that.

N.B. This problem seems to have been fixed in a different way in the develop branch.

### DIFF
--- a/src/headers/tomcrypt_macros.h
+++ b/src/headers/tomcrypt_macros.h
@@ -264,17 +264,17 @@ static inline unsigned ROR(unsigned word, int i)
 
 static inline unsigned ROLc(unsigned word, const int i)
 {
-   asm ("roll %2,%0"
+   asm ("roll %%cl,%0"
       :"=r" (word)
-      :"0" (word),"I" (i));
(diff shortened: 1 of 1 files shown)


sjaeckel:
your patch just makes the implementation of RORc equal to the one of ROR... that's not a good patch... you can have the same behavior by defining `LTC_NO_ROLC`

MMI:
Sweet, thanks! I have verified that defining LTC_NO_ROLC allows me to build bitkeeper in MacOS-land without our patched version of tomcrypt.

-g

PR #127, opened 2016-09-16
Add function declarations to remove compile warning

matswebjorn:
Current tomcrypt generates the following compile warning

`/xxx/tomcrypt/src/modes/xts/xts_test.c:27:17: warning: ISO C forbids initialization between function pointer and ‘void *’ [-Wpedantic]
    void *orig = cipher_descriptor[xts.cipher].accel_xts_encrypt;
                 ^
/xxx/tomcrypt/src/modes/xts/xts_test.c:34:52: warning: ISO C forbids assignment between function pointer and ‘void *’ [-Wpedantic]
    cipher_descriptor[xts.cipher].accel_xts_encrypt = orig;
                                                    ^
/xxx/tomcrypt/src/modes/xts/xts_test.c: In function ‘_xts_test_accel_xts_decrypt’:
/xxx/tomcrypt/src/modes/xts/xts_test.c:51:17: warning: ISO C forbids initialization between function pointer and ‘void *’ [-Wpedantic]
    void *orig = cipher_descriptor[xts.cipher].accel_xts_decrypt;
                 ^
/xxx/tomcrypt/src/modes/xts/xts_test.c:58:52: warning: ISO C forbids assignment between function pointer and ‘void *’ [-Wpedantic]
    cipher_descriptor[xts.cipher].accel_xts_decrypt = orig;
                                                    ^`

accel_xts_encrypt and accel_xts_decrypt needs a proper function definition to eliminate this warning which is needed to get a clean compile

### DIFF
--- a/src/headers/tomcrypt_cipher.h
+++ b/src/headers/tomcrypt_cipher.h
@@ -336,6 +336,35 @@ typedef struct {
 } symmetric_F8;
 #endif
 
+/** Accelerated XTS encryption definition
+    @param pt      Plaintext
+    @param ct      Ciphertext
+    @param blocks  The number of complete blocks to process
+    @param tweak   The 128-bit encryption tweak (input/output).
+                   The tweak should not be encrypted on input, but
+                   next tweak will be copied encrypted on output.
+    @param skey1   The first scheduled key context
+    @param skey2   The second scheduled key context
+    @return CRYPT_OK if successful
+ */
+typedef int (*accel_xts_encrypt_func_)(const unsigned char *pt, unsigned char *ct,
+        unsigned long blocks, unsigned char *tweak, symmetric_key *skey1,
+        symmetric_key *skey2);
+
+/** Accelerated XTS decryption definition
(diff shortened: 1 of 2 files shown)


sjaeckel:
Srsly `-Wpedantic`?

matswebjorn:
Yes, srsly

karel-m:
Hi,

I see your point but your patch is dealing with `-Wpedantic` warnings only in `accel_xts_encrypt` + `accel_xts_decrypt`; which probably solves your use case, however, there are other places that throw `ISO C forbids ...` when compiled with `-Wpedantic`.

Therefore  :thumbsdown: from me to this patch.

First we should decide whether we want a clean build with `-Wpedantic` and if yes then fix **all** warnings not just some of them. Anyway it would be better to open a separated issue named like `Avoiding -Wpedantic warnings` or so.

-- Karel

sjaeckel:
Thanks @karel-m for it pointing out correctly.

PR #103, opened 2016-09-25
Serious bug in rng_get_bytes @ MS Windows

sjaeckel:
In `rng_get_bytes.c` we have:

``` c
#if defined(LTC_DEVRANDOM)
   x = rng_nix(out, outlen, callback);   if (x != 0) { return x; }
#endif
#if defined(WIN32) || defined(_WIN32) || defined(WINCE)
   x = rng_win32(out, outlen, callback); if (x != 0) { return x; }
#endif
#ifdef ANSI_RNG
   x = rng_ansic(out, outlen, callback); if (x != 0) { return x; }
#endif
   return 0;
```

On MS Windows unfortunately `LTC_DEVRANDOM` is defined which means that even if we are on MS Windows we will try to open/read `/dev/random` file first and after it fails we will try Windows' `CryptGenRandom` call.

It seems OK unless an evil adversary creates on your disk a file `c:\dev\random` (e.g. full of zeroes). I have tested libtomcrypt built with mingw-w64 compiler and it happily opens/reads from `c:\dev\random` - which, well, is not that random as one would expect.

My proposal is changing the above mentioned section like this:

``` c
#if defined(WIN32) || defined(_WIN32) || defined(WINCE)
   x = rng_win32(out, outlen, callback); if (x != 0) { return x; }
#elif defined(LTC_DEVRANDOM)
   x = rng_nix(out, outlen, callback);   if (x != 0) { return x; }
#endif
#ifdef ANSI_RNG
   x = rng_ansic(out, outlen, callback); if (x != 0) { return x; }
#endif
   return 0;
```

Other options are forcefully undefine `LTC_DEVRANDOM` (perhaps somewhere in `tomcrypt_custom.h`) when on MS Windows. Or maybe somebody might have another proposal.

### DIFF
--- a/src/prngs/rng_get_bytes.c
+++ b/src/prngs/rng_get_bytes.c
@@ -141,11 +141,10 @@ unsigned long rng_get_bytes(unsigned char *out, unsigned long outlen,
 
    LTC_ARGCHK(out != NULL);
 
-#if defined(LTC_DEVRANDOM)
-   x = rng_nix(out, outlen, callback);   if (x != 0) { return x; }
-#endif
 #if defined(_WIN32) || defined(_WIN32_WCE)
    x = rng_win32(out, outlen, callback); if (x != 0) { return x; }
+#elif defined(LTC_DEVRANDOM)
+   x = rng_nix(out, outlen, callback);   if (x != 0) { return x; }
 #endif
 #ifdef ANSI_RNG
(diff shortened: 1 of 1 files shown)


karel-m:
This PR looks fine. 

I do no think it should wait till 2.0 milestone.

sjaeckel:
True! please rebase, merge & close

PR #132, opened 2016-10-07
Fix compile warning on free build

matswebjorn:
The definition of LTC_NO_PROTOTYPES in tomcrypt_custom.h will generate warnings if more than one of XMALLOC, XREALLOC, ... is defined. LTC_NO_PROTOTYPES must be undefined before it is redefined.

### DIFF
--- a/src/headers/tomcrypt_custom.h
+++ b/src/headers/tomcrypt_custom.h
@@ -4,44 +4,51 @@
 /* macros for various libc functions you can change for embedded targets */
 #ifndef XMALLOC
    #ifdef malloc
-   #define LTC_NO_PROTOTYPES
+      #undef  LTC_NO_PROTOTYPES
+      #define LTC_NO_PROTOTYPES
    #endif
 #define XMALLOC  malloc
 #endif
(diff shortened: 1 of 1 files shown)


src/headers/tomcrypt_custom.h
```diff
@@ -64,7 +72,9 @@
 
 #ifndef XQSORT
    #ifdef qsort
-   #define LTC_NO_PROTOTYPES
+      #ifndef LTC_NO_PROTOTYPES
```
rofl0r: why you use ifdef here, unlike in all other places ?

sjaeckel:
and please don't create a new PR this time, update this one.

sjaeckel:
fine like that?

karel-m:
I only suggest this little improvement to fix/132 branch

https://github.com/libtom/libtomcrypt/commit/c046b780cb214444b9a40586f16a7261e93ac2b5

sjaeckel:
> I only suggest this little improvement to fix/132 branch
> 
> c046b78

That looks good, I added another comment to the comment of @rofl0r

PR #134, opened 2016-10-28
Remove reference to public domain software

mpherg:
"Public domain" [isn't a license](http://www.linuxjournal.com/article/6225), and it causes all sorts of problems for
lawyers if we want to use it. Remove public domain and just leave the WTF
license, which I believe is what the author is really concerned about.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,17 +1,3 @@
-LibTomCrypt is licensed under DUAL licensing terms.
-
-Choose and use the license of your needs.
-
-[LICENSE #1]
-
-LibTomCrypt is public domain.  As should all quality software be.
-
-Tom St Denis
-
-[/LICENSE #1]
-
-[LICENSE #2]
-
             DO WHAT THE FUCK YOU WANT TO PUBLIC LICENSE
                     Version 2, December 2004
 
@@ -24,6 +10,4 @@ Tom St Denis
             DO WHAT THE FUCK YOU WANT TO PUBLIC LICENSE
    TERMS AND CONDITIONS FOR COPYING, DISTRIBUTION AND MODIFICATION
 
-  0. You just DO WHAT THE FUCK YOU WANT TO. 
-
-[/LICENSE #2]
+  0. You just DO WHAT THE FUCK YOU WANT TO.


mpherg:
Or you could simply use the [unlicense](http://choosealicense.com/licenses/unlicense/).

Scarletts:
@mpherg The author of that article has changed his mind:

> For the record, I have already voted +1 to approve the CC0 public domain
dedication and fallback license as OSD compliant. I admit that I have argued
for years against the "public domain" as an open source license, but in
retrospect, considering the minimal risk to developers and users relying on
such software and the evident popularity of that "license", I changed my
mind. One can't stand in the way of a fire hose of free public domain
software, even if it doesn't come with a better FOSS license that I trust
more.

[Source](https://lists.opensource.org/pipermail/license-review/2012-March/001679.html)

from [djb's page on public domain](https://cr.yp.to/publicdomain.html)

sjaeckel:
> If you trust the WTFPL, the public domain declaration is irrelevant anyway, so this is a non-issue.

therefor closing this as non-issue

mpherg:
Unfortunately our lawyers disagree with this interpretation of public domain. I guess we will have to find some other library.

sjaeckel:
Why can't you just use the WTFPL license? Therefor it's dual licensed...

...or even just take it and re-license it under whatever license you want...

tomstdenis:
This reminds me of dealing with Sandisk about 10 years ago.  They also didn't like the public domain license.  So I told them for a $5000 fee I'd license it to them under a proprietary license.  They surprisingly didn't accept that.

The "public domain" really is a thing in the USA.  Works that have aged out their copyright terms are license free and within the public's domain.

Or do you argue that Hamlet is a copyrighted text?

mpherg:
Our lawyers argue that if the work is in the public domain (even with a dual licensing model), the author (i.e. copyright holder) may at some future date choose to apply a restrictive or proprietary license, and thus leave users (my company, for instance) at risk. This risk exists until a license is applied.

In the end this is your work and you can do whatever you want with it, and I respect that. Thanks for the consideration.

mpherg:
And your copyright hasn't expired yet, unless you have been dead for 70 years (https://www.copyright.gov/help/faq/faq-duration.html). Hamlet is a bit older.

PR #104, opened 2017-02-23
rng_ansic - why do we require CLOCKS_PER_SEC <= 10000?

sjaeckel:
According to `man 3 clock`

```
CONFORMING TO
       C89, C99, POSIX.1-2001.  POSIX requires that CLOCKS_PER_SEC 
       equals 1000000 independent of the actual resolution.
```

However in `rng_get_bytes.c` we require `100 <= CLOCKS_PER_SEC <= 10000` which is not gonna match. 

What would break when we change it to `100 <= CLOCKS_PER_SEC <= 1000000`?

### DIFF
--- a/src/headers/tomcrypt_custom.h
+++ b/src/headers/tomcrypt_custom.h
@@ -34,9 +34,6 @@
 #ifndef XCLOCK
 #define XCLOCK   clock
 #endif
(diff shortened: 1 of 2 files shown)


karel-m:
great :+1:

sjaeckel:
you have to select "approve" in the radio buttons :)

PR #121, opened 2016-06-28
Fix all warnings from -Wcast-align

mattkelly:
This fixes a very large amount of annoying warnings generated by -Wcast-align. Building with clang-207.0.31 on OS X.

Example of warnings that are fixed in ccm_memory.c:

```
src/encauth/ccm/ccm_memory.c:259:23: warning: cast from 'unsigned char *' to 'LTC_FAST_TYPE *' (aka 'unsigned long long *') increases required
      alignment from 1 to 8 [-Wcast-align]
                    *((LTC_FAST_TYPE*)(&PAD[z]))  ^= *((LTC_FAST_TYPE*)(&pt[y+z]));
                      ^~~~~~~~~~~~~~~~~~~~~~~~~
src/encauth/ccm/ccm_memory.c:259:56: warning: cast from 'unsigned char *' to 'LTC_FAST_TYPE *' (aka 'unsigned long long *') increases required
      alignment from 1 to 8 [-Wcast-align]
                    *((LTC_FAST_TYPE*)(&PAD[z]))  ^= *((LTC_FAST_TYPE*)(&pt[y+z]));
                                                       ^~~~~~~~~~~~~~~~~~~~~~~~~~
src/encauth/ccm/ccm_memory.c:260:23: warning: cast from 'unsigned char *' to 'LTC_FAST_TYPE *' (aka 'unsigned long long *') increases required
      alignment from 1 to 8 [-Wcast-align]
                    *((LTC_FAST_TYPE*)(&ct[y+z])) = *((LTC_FAST_TYPE*)(&pt[y+z])) ^ *((LTC_FAST_TYPE*)(&CTRPAD[z]));
                      ^~~~~~~~~~~~~~~~~~~~~~~~~~
src/encauth/ccm/ccm_memory.c:260:55: warning: cast from 'unsigned char *' to 'LTC_FAST_TYPE *' (aka 'unsigned long long *') increases required
      alignment from 1 to 8 [-Wcast-align]
                    *((LTC_FAST_TYPE*)(&ct[y+z])) = *((LTC_FAST_TYPE*)(&pt[y+z])) ^ *((LTC_FAST_TYPE*)(&CTRPAD[z]));
                                                      ^~~~~~~~~~~~~~~~~~~~~~~~~~
src/encauth/ccm/ccm_memory.c:260:87: warning: cast from 'unsigned char *' to 'LTC_FAST_TYPE *' (aka 'unsigned long long *') increases required
      alignment from 1 to 8 [-Wcast-align]
[...]

PR #144, opened 2017-02-24
fix coverity finding: dsa_import double free

karel-m:
@sjaeckel agree?

### DIFF
--- a/src/pk/dsa/dsa_import.c
+++ b/src/pk/dsa/dsa_import.c
@@ -99,9 +99,9 @@ int dsa_import(const unsigned char *in, unsigned long inlen, dsa_key *key)
          goto LBL_ERR;
       }
 
-      err = der_decode_subject_public_key_info(in, inlen,
-        PKA_DSA, tmpbuf, &tmpbuf_len,
-        LTC_ASN1_SEQUENCE, params, 3);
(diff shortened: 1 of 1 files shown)


src/pk/dsa/dsa_import.c
```diff
@@ -103,10 +103,12 @@ int dsa_import(const unsigned char *in, unsigned long inlen, dsa_key *key)
         PKA_DSA, tmpbuf, &tmpbuf_len,
         LTC_ASN1_SEQUENCE, params, 3);
       if (err != CRYPT_OK) {
```
rofl0r: whitespace seems to be misaligned to the above statements here

karel-m: indentation should be fixed now - bebbf0a

karel-m:
just FYI: while playing with clang when checking #121 I have also tried clang's static analyzer which also identified this issue:
```
src/pk/dsa/dsa_import.c:128:4: warning: Attempt to free released memory
   XFREE(tmpbuf);
   ^~~~~~~~~~~~~
./src/headers/tomcrypt_custom.h:15:18: note: expanded from macro 'XFREE'
#define XFREE    free
                 ^
1 warning generated.
```

PR #145, opened 2017-02-24
fix coverity finding: cbc_decrypt out-of-bound read

karel-m:
@sjaeckel agree?

### DIFF
--- a/src/modes/cbc/cbc_decrypt.c
+++ b/src/modes/cbc/cbc_decrypt.c
@@ -45,7 +45,7 @@ int cbc_decrypt(const unsigned char *ct, unsigned char *pt, unsigned long len, s
    }
 
    /* is blocklen valid? */
-   if (cbc->blocklen < 1 || cbc->blocklen > (int)sizeof(cbc->IV)) {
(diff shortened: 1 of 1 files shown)


sjaeckel:
IIUC that's not really a false positive, it's something that could happen in case that code is hit by a cipher with a bigger block length than 16 which:

1. is currently not implemented
2. I'm not aware of to even exist as publicly known cipher

I'm somehow against that patch as it increases the used stack without any benefit.

I'd prefer the approach to add a check [here](https://github.com/libtom/libtomcrypt/blob/develop/src/modes/cbc/cbc_decrypt.c#L48)

karel-m:
Like this?
```
- if (cbc->blocklen < 1 || cbc->blocklen > (int)sizeof(cbc->IV)) {
+ if (cbc->blocklen < 1 || cbc->blocklen > (int)sizeof(cbc->IV)  || cbc->blocklen > (int)sizeof(tmp)) {
```

karel-m:
changed, I'll wait for travis-ci results and then merge

PR #147, opened 2017-02-25
sanitize some Perl scripts

fperrad:
(Perl4 is gone)


karel-m:
👍 from me

I would consistently use in all `*.pl` files the first script line:
```
#!/usr/bin/env perl
```

karel-m:
rebased + merged into develop

PR #148, opened 2017-02-25
use static functions where appropriate

fperrad:


karel-m:
I agree with fixing `chc.c` as the `static` is clearly missing here.

The situation with `dsa_make_key.c` is interesting:
* `dsa_make_params` and `dsa_make_key_ex` are really used only in this `.c` file
* but the comments before those functions indicate that they were designed to be a part of API
* looking at `git blame` I have to admit that the current code in `dsa_make_key.c` is mostly mime (just to make it clear whom to blame)
* IMO `dsa_make_params` and `dsa_make_key_ex`  should be added to `tomcrypt_pk.h` therefore I am not for making them static
* anyway, good catch

sjaeckel:
@fperrad can you do the changes as proposed by @karel-m or should we?

src/pk/dsa/dsa_make_key.c
```diff
@@ -28,7 +28,7 @@
   @param g             [out] bignum where generated 'g' is stored (must be initialized by caller)
   @return CRYPT_OK if successful, upon error this function will free all allocated memory
 */
-int dsa_make_params(prng_state *prng, int wprng, int group_size, int modulus_size, void *p, void *q, void *g)
+static int dsa_make_params(prng_state *prng, int wprng, int group_size, int modulus_size, void *p, void *q, void *g)
```
sjaeckel: Instead of making this static, please add the prototype to tomcrypt_pk.h

src/pk/dsa/dsa_make_key.c
```diff
@@ -204,7 +204,7 @@ int dsa_make_params(prng_state *prng, int wprng, int group_size, int modulus_siz
   @param g_hex         Hexadecimal string 'g'
   @return CRYPT_OK if successful, upon error this function will free all allocated memory
 */
-int dsa_make_key_ex(prng_state *prng, int wprng, int group_size, int modulus_size, dsa_key *key, char* p_hex, char* q_hex, char* g_hex)
+static int dsa_make_key_ex(prng_state *prng, int wprng, int group_size, int modulus_size, dsa_key *key, char* p_hex, char* q_hex, char* g_hex)
```
sjaeckel: the same as above

fperrad:
done

src/headers/tomcrypt_pk.h
```diff
@@ -404,6 +404,11 @@ typedef struct {
 } dsa_key;
 
 int dsa_make_key(prng_state *prng, int wprng, int group_size, int modulus_size, dsa_key *key);
+
+int dsa_make_key_ex(prng_state *prng, int wprng, int group_size, int modulus_size, dsa_key *key, char* p_hex, char* q_hex, char* g_hex);
```
sjaeckel: @karel-m while asking myself why these two functions' arguments `p, q, g` don't have the same type I realized that I've no clue how this API is intended to be used, resp. why someone would want to use `dsa_make_params()` itself!?

karel-m: `dsa_make_params` is basically the same what you do by e.g. `openssl dsaparam 1024 -C`

Looking at my `dsa_make_params` code I am not sure whether the calling interface/API is good (perhaps worth a separate issue "Polish dsa_* API" or so)

karel-m:
I have created #154 where I want to move further discussion about `dsa_*` functions.

I vote for merging this PR as it is now.

karel-m:
rebased-merged-pushed closing

PR #135, opened 2017-02-27
RC2 does not support 40bit encryption

sjaeckel:
Originally reported here: https://github.com/DCIT/perl-CryptX/issues/25 by @nlitsme (a user of my perl binding to libtomcrypt).

The fix looks like this https://github.com/DCIT/perl-CryptX/commit/d5a49f8dbec01e020e4a798982acffb22fb2945d + https://github.com/DCIT/perl-CryptX/commit/0b0d21bfe70d0465792ba2e5f6a8187a27222fa1

Currently in libtomcrypt we support RC2 key sizes 64-1024bits, in theory RC2 can handle 8-1024bits.


karel-m:
Yes (except the crypt.tex)

I have this patch in my perl bindings some time - here is the original report https://github.com/DCIT/perl-CryptX/issues/25 with a short discussion

sjaeckel:
> Yes (except the crypt.tex)
> 
> I have this patch in my perl bindings some time - here is the original report DCIT/perl-CryptX#25 with a short discussion

Okay, then either I did an error when copying the testvector or I don't know what else is wrong...

karel-m:
Unfortunately I can confirm, in fact there are 3 failing test rfc2268 vectors: 
* `Key length (bytes) = 7 / Effective key length (bits) = 64`
* `Key length (bytes) = 16 / Effective key length (bits) = 64`
* `Key length (bytes) = 33 / Effective key length (bits) = 129`

karel-m:
@sjackel this can be merged, right?

karel-m:
@sjaeckel One more tiny change 05f85c3 (added to develop) - is it correct?

PR #139, opened 2017-02-28
rand_prime undefined reference when using  CFLAGS="-DLTM_DESC -DLTC_EASY

sjaeckel:
linker will yield 'undefined reference' to rand_prime function (math/rand_prime.c) for any ltc function referencing it - when LTC is compiled with CFLAGS="-DLTM_DESC -DLTC_EASY.

LTC_EASY will define LTC_NO_PRNGS and rand_prime will therefore not be pulled in by cpp.


karel-m:
rebased on current develop + fix dde11de + pushed (with force)

build.sh
```diff
@@ -3,6 +3,11 @@ echo "$1 ($2, $3)..."
 
 make clean 1>/dev/null 2>/dev/null
 
+if [ -f check-source.pl ] ; then
```
sjaeckel: I don't like the fact that all builds fail now because of a formatting error in the sources

karel-m: ok, I'll remove it

karel-m: What about adding it directly to `.travis.yml` like 
```
    BUILDSCRIPT="check-source.pl"
    BUILDNAME=" "
    BUILDOPTIONS=" "
```

PR #79, opened 2015-09-10
RSA import x509

sjaeckel:
Yeah I know it's nothing like a general purpose x.509 parser, but at least it can import the public key from an RSA certificate.
Any suggestions/comments/ideas?

PR #157, opened 2017-03-01
RSA import pkcs8

karel-m:
as requested in #79


karel-m:
Indentation is not 3-spaces

karel-m:
from me this PR is ready for review / merge

karel-m:
Maybe:
```
- int rsa_import_pkcs8(unsigned char *in, unsigned long inlen, rsa_key *key);
+ int rsa_import_pkcs8(unsigned char *in, unsigned long inlen, unsigned char *passwd, unsigned long passwdlen, rsa_key *key);
```

Although the current implementation does not support encrypted (password protected) pkcs8 keys, it might be a good idea to support the password in the function call.

sjaeckel:
> ```
> - int rsa_import_pkcs8(unsigned char *in, unsigned long inlen, rsa_key *key);
> + int rsa_import_pkcs8(unsigned char *in, unsigned long inlen, unsigned char *passwd, unsigned long passwdlen, rsa_key *key);
> ```

Sounds like a good idea! I don't know the details about the passwords that can be used, but if it can also be a binary stream I'd prefer to use `..., const void* passwd, unsigned long passwdlen, ...`. In case that it's limited to a "string" I'd prefer `..., const char* passwd, rsa_key ...`.

karel-m:
password are not limited to strings

src/pk/rsa/rsa_import_pkcs8.c
```diff
@@ -0,0 +1,151 @@
...
+   LTC_SET_ASN1(key_seq_e, 1, LTC_ASN1_INTEGER, iter, 1UL);
+   LTC_SET_ASN1(alg_seq_e, 0, LTC_ASN1_OBJECT_IDENTIFIER, oid, 16UL);
+   LTC_SET_ASN1(alg_seq_e, 1, LTC_ASN1_SEQUENCE, key_seq_e, 2UL);
+   LTC_SET_ASN1(top_seq_e, 0, LTC_ASN1_SEQUENCE, alg_seq_e, 2UL);
+   LTC_SET_ASN1(top_seq_e, 1, LTC_ASN1_OCTET_STRING, buf2, buf2len);
+   err=der_decode_sequence(in, inlen, top_seq_e, 2UL);
+   if (err == CRYPT_OK) {
+      LTC_UNUSED_PARAM(passwd);
+      LTC_UNUSED_PARAM(passwdlen);
+      /* XXX: TODO encrypted pkcs8 not implemented yet */
```
sjaeckel: I just remembered that there was something on the ML https://groups.google.com/d/topic/libtom/sFdia0rRcfM/discussion

I don't know if this is what we need here, but it looks kinda similar :)

Will you have a look? Otherwise I'll do it when I have the time.

karel-m: Exactly, to handle the encrypted keys we should be able to decrypt via ` pbeWithMD5AndDES-CBC` and/or similar.

The latest spec for PBES1 / PBES2 are IMO in RFC 8018 (aka PKCS5 v2.1)
* https://tools.ietf.org/html/rfc8018#section-6.1
* https://tools.ietf.org/html/rfc8018#section-6.2

The code from mailing list looks similar to PbeWithMD5AndTripleDES

Anyway, it might be a good idea to have something like `src\misc\pkcs5\pbes1.c` + `src\misc\pkcs5\pbes2.c` so that we can later reuse it in `ecc_import_pkcs8` or event implement stuff like `rsa_import_pkcs12`.

Slightly OT: this PKCS8 related bug report related to EC keys https://github.com/DCIT/perl-CryptX/issues/28 demonstrates that PKCS8 is not easy

sjaeckel:
We can merge this one in my opinion and open a new issue for encrypted pkcs#8 files

karel-m:
I would also merge this one but there are 2 things to mention if anybody will hack on this in the future:

1/ This implementation does not support password protected keys (yes, we know it).

2/ The implementation more or less follows https://tools.ietf.org/html/rfc5208; however, it was obsoleted by https://tools.ietf.org/html/rfc5958 Which basically means that there can exist some correct/valid pkcs#8 formatted keys that my implementation cannot load - see pkcs8 related troubles in https://github.com/DCIT/perl-CryptX/issues/28

I was thinking about replacing the current implementation with our ASN.1 flexi parser.

sjaeckel:
@karel-m I added another commit, can you please review?

karel-m:
Your last commit is correct

PR #158, opened 2017-03-02
check makefiles (check_source.sh)

karel-m:


karel-m:
Seems to be working - https://travis-ci.org/libtom/libtomcrypt/jobs/206910033

check_source.sh
```diff
@@ -10,6 +10,19 @@ if [ -f check-source.pl ] ; then
...
+if [ -d .git ] ; then
+  echo "checking makefiles..."
+  ./updatemakes.sh
+  outofdate="$(git diff --name-only HEAD)"
+  if [ -n "$outofdate" ] ; then
+    echo "FAIL: makefiles not updated"
+    echo "$outofdate";
+    exit 1
+  else
+     echo "PASS"
```
sjaeckel: indentation

check_source.sh
```diff
@@ -10,6 +10,19 @@ if [ -f check-source.pl ] ; then
   perl check-source.pl || exit 1
 fi
 
+if [ -d .git ] ; then
+  echo "checking makefiles..."
+  ./updatemakes.sh
+  outofdate="$(git diff --name-only HEAD)"
```
sjaeckel: this makes the entire flow dependent if it's run from the git repo

as we're going to distribute archives as release, I think this needs another solution, but the idea is good!

makefile
```diff
@@ -42,7 +42,7 @@ LIBTEST_S=$(LIBTEST)
 
 #List of objects to compile.
 #START_INS
-OBJECTS=src/ciphers/aes/aes_enc.o src/ciphers/aes/aes.o src/ciphers/anubis.o src/ciphers/blowfish.o \
+OBJECTS=src/ciphers/aes/aes.o src/ciphers/aes/aes_enc.o src/ciphers/anubis.o src/ciphers/blowfish.o \
```
sjaeckel: srsly, the new sorting sorts alphabetically in a different order than the previous sorting? :rofl:

karel-m:
> srsly, the new sorting sorts alphabetically in a different order than the previous sorting? 

yes, because now we rely on perl's built-in sort (which should work more or less the same on all platforms), the old version was based on `sort` command which depends on locale you have set in your shell

karel-m:
is updated version fine for merge?

check_source.sh
```diff
@@ -10,6 +10,19 @@ if [ -f check-source.pl ] ; then
   perl check-source.pl || exit 1
 fi
 
+if [ -f ./updatemakes.sh ] ; then
+  echo "checking makefiles..."
+  beforeupdate=$(grep '' makefile*)
+  ./updatemakes.sh
```
sjaeckel: I like that a lot better, but I still don't like the fact that the files get now modified on disk when you run this (I especially dislike that the filename lies and says 'check' and then does 'modify' :-) )

karel-m:
now completely new approach

updatemakes.sh
```diff
@@ -1,26 +1,5 @@
...
-perl filter.pl makefile.msvc tmplist
-sed -e 's/\.o /.obj /g' -e 's/ *$//' < tmp.delme > makefile.msvc
-rm -f tmp.delme
-
-rm -f tmplist
-
-if [ $# -eq 1 ] && [ "$1" == "-c" ]; then
-  git add ${makefiles[@]} makefile.msvc && git commit -m 'Update makefiles'
-fi
+exit 0
```
sjaeckel: missing newline at eof

sjaeckel:
regarding 7046af108e90df65942c92f417594e2cefdf9464: I think we should provide aes_enc for compatibility reasons, or not?

karel-m:
> regarding 7046af1: I think we should provide aes_enc for compatibility reasons, or not?

As for MS Windows builds via `*.vcproj` I vote leaving out `aes_anc`. 

And generally I would prefer not using the hack we currently use for building `aes.o` + `aes_enc.o` from a single `aes.c`. IMO having a separate `aes_enc.c` + a handy set of includes+ifdefs would be much better and will make maintenance easier. But it is perhaps worth a separate ticket.

Do you (or any other reader of this ticket) have any idea what was the motivation for `aes_anc` and why only AES and not generally other ciphers as well?

karel-m:
I have returned the aes_enc hack to helper.pl - 6f9e427f77 (it is commented out so that it can be easily put back if needed)

PR #122, opened 2017-03-09
adler32 vs. crc32 inconsistency

sjaeckel:
in `crc32_test` we have:

``` c
{
   const void* in = "libtomcrypt";
   const unsigned char crc32[] = { 0xef, 0x76, 0x73, 0xb3 };
   ....
}
```

However the correct CRC32 from "libtomcrypt" string is IMHO `b37376ef` not `ef7673b3`.

By which I mean the actual CRC32 implementation seems to be somehow incorrect not `crc32_test`.


sjaeckel:
fine to rebase&merge

PR #166, opened 2017-03-15
more linting

fperrad:
fix a recent commit https://github.com/libtom/libtomcrypt/commit/0e9b3da3ea6f0318b9b1da85ac1f4d2b31d588a4


karel-m:
thanks

PR #172, opened 2017-03-22
Run tests for tomsfastmath

sjaeckel:
the latest tfm version has `fp_rand()` available, so we can test this now

as there is currently no static library of tfm it's also nice as this also adds a test of the `makefile.shared`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,12 +7,13 @@ addons:
     packages:
     - binutils
     - libtommath-dev
-    - libtfm-dev
 
 before_script:
   - gem install coveralls-lcov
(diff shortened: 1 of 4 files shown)


karel-m:
@sjaeckel would this one work after rebasing this PR (I mean with re-worked `makefile.shared`)?

sjaeckel:
Right, I didn't explain yet why this one breaks... https://github.com/libtom/tomsfastmath/commit/fb62c9cde89a5dbbda569d3610c2315f5fb284e2 `fp_rand()` wasn't exported in the version of tfm that is currently available in debian testing

So regarding your question: nope it wouldn't, as there's no updated tfm yet as debian is still stuck in code freeze.

But I'm working on a temporary solution...

karel-m:
👍

PR #173, opened 2017-03-22
SHA3

karel-m:
SHAKE128 and SHAKE256 need to be added - I only do not know yet how to do it (they do not fit well to our hash machinery).

The implementation is based on https://github.com/brainhub/SHA3IUF


karel-m:
**I am extremely strongly against creating dependency on `stdint.h`.**

I lost this battle in libtommath and the result is that I have to keep a specially patched `tommath.h` for my perl bindings which is a pain for me. Frankly we did not gain anything in libtommath by starting using `stdint.h`.

And please do not give me those stories about how old is C99. It was a week ago when I was trying to improve libtomcrypt's Visual C builds. And guess what, on Visual Studio 2008 Express there is no `stdint.h` so `tommath.h` hit me again.

In short: either `me` or `stdint.h` 😃

rofl0r:
using size_t does not necessarily mean including stdint.h. since you're a fan of ifdef hacks over configure tests you could hardcode a list of CPP defines which define HAVE_STDINT_H and only use it there, for the other platforms fall back to typedefing it to unsigned long, or unsigned long long on windows64 bit.

karel-m:
Let's move `size_t` discussion to  #175 so that it does not block this PR

sjaeckel:
Regarding 4abbf45731c235c973da9a1ae349bc788eb232ca wouldn't it make sense to have a separate `sha3_shake` hash descriptor so we can re-use the hash helpers?

karel-m:
unfortunately `sha3_shake_done` does not fit into `ltc_hash_descriptor` because it is missing the `outlen` (SHAKE functions are designed in a way that they give the output as long as you ask.

```C
int (*done)(hash_state *hash, unsigned char *out);
```

karel-m:
Kind of `outlen` inconsistency we have here:
```C
int hash_memory(int hash, const unsigned char *in,  unsigned long inlen, unsigned char *out, unsigned long *outlen);
int sha3_shake_memory(int num, const unsigned char *in, unsigned long inlen, unsigned char *out, unsigned long outlen);
```

The thing is that `hash_memory` gives you the output as long as the `hashsize` is, whereas `sha3_shake_memory` gives you the output as long as you have requested by `outlen` param. So these functions are a bit different in their nature, `sha3_shake_memory` is more like `hkdf` or `prng->read` where we use `outlen` not `*outlen`.

But providing that we have a separate outlen related issue #176 I propose to discuss/solve it rather there than to block this PR. When we decide to change `outlen` to `*outlen` it is trivial to change the implementation.

karel-m:
Rebased & squashed into 1 commit, travis-ci all green. I propose merging to develop.

sjaeckel:
I'd prefer to have `sha3_shake_memory()` already with `*outlen` then it's fine by me to merge

karel-m:
`sha3_shake_memory` changed

Should I also change `int sha3_shake_done(hash_state *md, unsigned char *out, unsigned long outlen)` to `*outlen`? By "normal" hashes we have no outlen - like `int sha3_done(hash_state *md, unsigned char *hash)`

sjaeckel:
> Should I also change `int sha3_shake_done(hash_state *md, unsigned char *out, unsigned long outlen)` to `*outlen`? By "normal" hashes we have no outlen - like `int sha3_done(hash_state *md, unsigned char *hash)`

I also thought about that, but as there's no reference I'd say we keep it like that.

btw. I also had the idea why can't we provide a `sha3_shake_set_outlen(hash_state *md, unsigned long outlen)`, then we could have `sha3_shake_done(hash_state *md, unsigned char *hash)` :)

but it was just an idea and as there's no way to properly integrate this in the 'hashing framework' I think it doesn't make sense to squeeze the sha3 shake API in there

PR #171, opened 2017-03-21
RFC 7539 - ChaCha20 and Poly1305

karel-m:
Basically the reference implementation adopted to our C macros and our calling conventions.

https://tools.ietf.org/html/rfc7539

PR #178, opened 2017-03-26
RC4 PRNG - broken import/export

karel-m:
IMO a bug.

I have run into this when thinking about the idea of `src/stream/rc4` (mentioned in #171)

### DIFF
--- a/src/prngs/rc4.c
+++ b/src/prngs/rc4.c
@@ -257,31 +257,6 @@ int rc4_test(void)
           return CRYPT_FAIL_TESTVECTOR;
        }
    }
-   {
-      prng_state prng1, prng2;
-      unsigned char dump[300], buf1[100], buf2[100];
-      unsigned long dumplen = sizeof(dump), i;
-      unsigned char entropy[] = { 0x01, 0x23, 0x45, 0x67, 0x89, 0xab, 0xcd, 0xef };
-
-      if ((err = rc4_start(&prng1)) != CRYPT_OK)                                 return err;
-      if ((err = rc4_add_entropy(entropy, sizeof(entropy), &prng1)) != CRYPT_OK) return err;
-      if ((err = rc4_ready(&prng1)) != CRYPT_OK)                                 return err;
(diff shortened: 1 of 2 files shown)


karel-m:
Just to make it clear - what I have committed is just a failing test not a fix.

sjaeckel:
as a prng it's still fine ;-)

IIRC that was the point in this import/export functionality

if we make it a stream cipher, there won't be a import/export, right?

karel-m:
Import/export is possible with stream cipher based PRNG, see my recent https://github.com/libtom/libtomcrypt/blob/pr/chacha20poly1305/src/prngs/chacha.c (we have `src/stream/chacha.c` = stream cipher and `src/prngs/chacha.c` = chacha based PRNG).

My plan is to update rc4 stuff to the style I have used in chacha implementation (I will commit it to this PR after #171 got it to develop).

karel-m:
So it seems that export/import is not working as I would expect by neither PRNG. Maybe I expect too much or I simply missed the idea behind export/import.

Is there anybody among readers of this PR who is using PRNG export/import?

sjaeckel:
> So it seems that export/import is not working as I would expect by neither PRNG. Maybe I expect too much or I simply missed the idea behind export/import.

IIRC I asked tom once about the same thing and he then said that this behavior is by design.

I think as it's a PRNG it doesn't really matter if the import does an exact recovery of the exported state... a.k.a. it is a successful import as long as it returns random data after the import (with a different internal state than no initialization/seeding)... :-)

karel-m:
If you put it that way I tend to propose dropping export/import functions from prngs because:
* `export` is the same as `read` (some amount of random bytes)
* `import` is equivalent to calling `add_entropy` (the bytes from previous step) + `make_ready`

IMO `export` / `import` should work as "save current state" & "return later exactly to the saved point" otherwise they do no make much sense (at least to me).

karel-m:
I am withdrawing this PR as according the doc import/export was not meant to be like "save current state" & "return later exactly to the saved point".

It was probably intended to be similar to e.g. `RAND_load_file` / `RAND_write_file` https://wiki.openssl.org/index.php/Manual:RAND_load_file(3)

PR #182, opened 2017-03-30
add GeneralizedTime DER processing

sjaeckel:
This implementation is a bit hairy as I couldn't find how exactly to implement this.

IIUC the reference should be in the [standard X.690](https://www.itu.int/rec/T-REC-X.690-201508-I/en)
But if I look at [this](http://www.oss.com/asn1/products/documentation/asn1-java/api/toed/com/oss/asn1/GeneralizedTime.html), [this](https://www.obj-sys.com/asn1tutorial/node14.html) or [this](http://javadoc.iaik.tugraz.at/iaik_jce/current/iaik/asn1/GeneralizedTime.html) I don't really know what to do :laughing:

### DIFF
--- a/libtomcrypt_VS2005.vcproj
+++ b/libtomcrypt_VS2005.vcproj
@@ -1457,6 +1457,22 @@
 						</File>
 					</Filter>
 					<Filter
+						Name="generalizedtime"
+						>
+						<File
+							RelativePath="src\pk\asn1\der\generalizedtime\der_decode_generalizedtime.c"
+							>
+						</File>
+						<File
+							RelativePath="src\pk\asn1\der\generalizedtime\der_encode_generalizedtime.c"
+							>
+						</File>
+						<File
+							RelativePath="src\pk\asn1\der\generalizedtime\der_length_generalizedtime.c"
+							>
+						</File>
+					</Filter>
+					<Filter
 						Name="ia5"
 						>
 						<File
--- a/libtomcrypt_VS2008.vcproj
+++ b/libtomcrypt_VS2008.vcproj
@@ -1698,6 +1698,22 @@
 						</File>
 					</Filter>
 					<Filter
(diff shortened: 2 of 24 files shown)


sjaeckel:
which spec did you look at?

karel-m:
Yes "a time of day, to **any** of the precisions defined in ISO 8601" sounds scary. 

Ad fractions you probably mean "Decimal fractions may be added to any of the three time elements. However, a fraction may only be added to the lowest order time element in the representation." - I think that we can for now ignore fractions of hours/minutes.

The other approach might be to turn`/* possible encodings are ..` comment into `/* currently supported encodings are ..` and exactly document what we can/cannot parse and leave the implementation as it is.

karel-m:
According to https://tools.ietf.org/html/rfc5280#section-4.1.2.5.2 the current implementation is sufficient for parsing X509 certificates (validity dates).

So as I proposed above: document exactly what we can + cannot handle and leave the implementation as it is.

sjaeckel:
uhm, I'd propose to even remove the fractional seconds support as I can't think of a neat way to implement signaling `.004` in an `unsigned`...

For now I'd have these ideas, where I'd prefer 2 if you think we should go for it...:
1. store it in reverse order `400` means `.004`
2. store it bitmasked, e.g. 6 bits zero, 10 bits accuracy, would allow us to have 63 leading zeros and fractional seconds up to 1023 ... bits 15-10 zeroes, bits 9-0 fractional seconds

karel-m:
I agree that we can completely drop fractional second support as RFC5280 says "GeneralizedTime values MUST NOT include fractional seconds."

Ad implementation I would simply store nanoseconds (if we can afford one "ulong32") like:
```C
- unsigned fs, /* fractional seconds */
+ ulong32  ns, /* nano second */
```
but I have not think too much about consequences.

PR #188, opened 2017-04-05
misuse of compare_testvector

fperrad:
`compare_testvector` has two different definitions see https://github.com/libtom/libtomcrypt/blob/develop/src/headers/tomcrypt_misc.h#L102-L109
When the macro definition is used, the comparison `compare_testvector() != 0` is wrong.
When the function definition is used, the comparison is useless.
So, this commit removes the comparison.

### DIFF
--- a/src/ciphers/rc2.c
+++ b/src/ciphers/rc2.c
@@ -370,8 +370,8 @@ int rc2_test(void)
         rc2_ecb_encrypt(tests[x].pt, tmp[0], &skey);
         rc2_ecb_decrypt(tmp[0], tmp[1], &skey);
 
-        if (compare_testvector(tmp[0], 8, tests[x].ct, 8, "RC2 CT", x) != 0 ||
-              compare_testvector(tmp[1], 8, tests[x].pt, 8, "RC2 PT", x) != 0) {
+        if (compare_testvector(tmp[0], 8, tests[x].ct, 8, "RC2 CT", x) ||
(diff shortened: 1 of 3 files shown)


sjaeckel:
very true

sjaeckel:
I just thought would it make sense to add this patch instead?

```
diff --git a/src/headers/tomcrypt_misc.h b/src/headers/tomcrypt_misc.h
index 91f87c5..76f4f6b 100644
--- a/src/headers/tomcrypt_misc.h
+++ b/src/headers/tomcrypt_misc.h
@@ -105,7 +105,7 @@ void print_hex(const char* what, const void* v, const unsigned long l);
 int compare_testvector(const void* is, const unsigned long is_len, const void* should, const unsigned long should_len, const
 #else
 #define compare_testvector(is, is_len, should, should_len, what, which) \
-   (((is_len) != (should_len)) || (XMEMCMP((is), (should), (is_len)) != 0))
+   ((((is_len) != (should_len)) || (XMEMCMP((is), (should), (is_len)) != 0)) ? 1 : 0)
 #endif
 
 /* $Source$ */
```

karel-m:
The macro `compare_testvector` should be fixed as proposed by @sjaeckel anyway.

Using `if (compare_testvector(..))` vs. `if(compare_testvector(..) != 0)` is just a matter of taste. Its nature is similar to `XMEMCMP` which we use in both ways `if (XMEMCMP(..))` as well as `if (XMEMCMP(..) != 0)`. I think we can leave out `!= 0` in these cases.

PR #190, opened 2017-04-07
chacha: more linting

fperrad:


src/stream/chacha/chacha_setup.c
```diff
@@ -16,8 +16,8 @@
 
 #ifdef LTC_CHACHA
 
-static const char sigma[16] = "expand 32-byte k";
-static const char tau[16]   = "expand 16-byte k";
+static const char * const sigma = "expand 32-byte k";
+static const char * const tau   = "expand 16-byte k";
```
karel-m: I would prefer
```
static const char sigma[] = "...";
static const char tau[]   = "...";
```
Simply just leaving both 16's

sjaeckel: I like it as proposed as it's the most natural & strict way to declare these pointers.

karel-m: OK, so merge it as it is

PR #195, opened 2017-04-23
improve hashsum

sjaeckel:
... and run hashsum for all hashes


coverage.sh
```diff
@@ -32,6 +32,15 @@ fi
 ./sizes
 ./constants
 ./multi
+for i in $(./hashsum | tail -n +3 | awk '{print $1}' | sort); do echo -n "$i: " && ./hashsum $i testprof/test.key ; done > hashsum_tv.txt
```
karel-m: Wouldn't it be better to move this part to `build.sh` - section `echo -n "testing..."` where we already do similar diffs with `tv_gen` outputs?

sjaeckel: I also thought about that, but as we have configurations like e.g. `LTC_EASY` which don't include all hashes I thought it's enough here.
But we could also move it to build.sh and add separate known-good testvectors for the different configurations.

karel-m: I see. IMO not worth blocking this PR. 

BTW shouldn't it be better tu utilize `hash_file` and/or `hash_filehandle` in `hashsum.c`?

PR #200, opened 2017-04-29
LTC_PTHREAD: some linting

fperrad:
the macro LTC_MUTEX_TYPE already contains a semicolon
see https://github.com/libtom/libtomcrypt/blob/develop/src/headers/tomcrypt_custom.h#L552

### DIFF
--- a/src/headers/tomcrypt_prng.h
+++ b/src/headers/tomcrypt_prng.h
@@ -66,7 +66,7 @@ typedef struct {
 #endif
    };
    short ready;            /* ready flag 0-1 */
-   LTC_MUTEX_TYPE(lock);   /* lock */
+   LTC_MUTEX_TYPE(lock)    /* lock */
 } prng_state;
 
 /** PRNG descriptor */


sjaeckel:
I'd say that this is the wrong fix - we should remove the semicolon from the define of `LTC_MUTEX_TYPE` (and also remove all the other semicolons related to `LTX_MUTEX`).
Any other opinions?

fperrad:
See the definition when no LTC_PTHREAD, https://github.com/libtom/libtomcrypt/blob/develop/src/headers/tomcrypt_custom.h#L562
The macros in `tomcrypt_custom.h` are correct.

sjaeckel:
The reason why I don't like that patch is that you suddenly have lines in your source code which aren't terminated by a semicolon.

1. it looks strange and like an error
2. https://www.securecoding.cert.org/confluence/display/c/PRE11-C.+Do+not+conclude+macro+definitions+with+a+semicolon

karel-m:
@fperrad was it just `LTC_MUTEX_TYPE` that throws this warning? Looking at other macros like `LTC_MUTEX_LOCK` I would expect the same warning complaining about double `;`.

fperrad:
[PRE11-C](https://www.securecoding.cert.org/confluence/display/c/PRE11-C.+Do+not+conclude+macro+definitions+with+a+semicolon) is a rule concerning the use of macro in function.
The use of a macro could not always looks like a function call, especially outside a function.

sjaeckel:
> PRE11-C is a rule concerning the use of macro in function.

uhm okay, that's not mentioned anywhere but if you say so...

> @fperrad was it just `LTC_MUTEX_TYPE` that throws this warning? Looking at other macros like `LTC_MUTEX_LOCK` I would expect the same warning complaining about double `;`.

and? TBH I would've expected the same

karel-m:
An argument for **"not having ;"** in `LTC_MUTEX_*` macro definitions might be consistency with `tomcrypt_macros.h` where we also do not use `;`

sjaeckel:
[mutex.patch.txt](https://github.com/libtom/libtomcrypt/files/966690/mutex.patch.txt)

what does your linter say after applying this patch?
for both configurations... with and without `LTC_PTHREAD`

fperrad:
Your patch is OK with `LTC_PTHREAD` defined,
but KO without `LTC_PTHREAD`
```
                                _
LTC_MUTEX_PROTO(ltc_cipher_mutex);
src/headers/tomcrypt_cipher.h  938  Error 19: Useless Declaration
                              _
LTC_MUTEX_PROTO(ltc_hash_mutex);
src/headers/tomcrypt_hash.h  465  Error 19: Useless Declaration
                      _
   LTC_MUTEX_TYPE(lock);   /* lock */
src/headers/tomcrypt_prng.h  69  Error 19: Useless Declaration
                              _
LTC_MUTEX_PROTO(ltc_prng_mutex);
src/headers/tomcrypt_prng.h  203  Error 19: Useless Declaration
                               _
LTC_MUTEX_GLOBAL(ltc_hash_mutex);
src/misc/crypt/crypt_hash_descriptor.c  22  Error 19: Useless Declaration

```

These 3 macros `LTC_MUTEX_PROTO`, `LTC_MUTEX_TYPE` and `LTC_MUTEX_GLOBAL` are used outside C function.

The refactor of `LTC_MUTEX_INIT`, `LTC_MUTEX_LOCK` and `LTC_MUTEX_UNLOCK` is correct.

karel-m:
I suggest to merge the proposed 1-line patch patch and discuss the further improvements in a separate issue as it seems that current develop is broken on MSVC (#204) and this 1-line patch can fix it.

Issue #199, opened 2017-04-29
Lint the latest release/1.18.0


karel-m:
@fperrad could you please run your lint-tool on the latest develop?

fperrad:
Here, a bunch of warnings which need some investigations:

```
--- Module:   src/pk/dh/dh.c (C)
                  _
   return CRYPT_OK;
src/pk/dh/dh.c  321  Warning 438: Last value assigned to variable 'y' (defined at line 259) not used
src/pk/dh/dh.c  259  Info 830: Location cited in prior message

--- Module:   src/pk/rsa/rsa_import_x509.c (C)
                                  _
   tmpbuf = XCALLOC(1, tmpbuf_len);
src/pk/rsa/rsa_import_x509.c  45  Info 838: Previously assigned value to variable 'tmpbuf' has not been used

--- Module:   src/pk/ecc/ltc_ecc_map.c (C)
                                                                  _
   if ((err = mp_montgomery_reduce(P->z, modulus, mp)) != CRYPT_OK)           { goto done; }
src/pk/ecc/ltc_ecc_map.c  47  Info 838: Previously assigned value to variable 'err' has not been used

--- Module:   src/pk/ecc/ecc_verify_hash.c (C)
           _
   mp    = NULL;
src/pk/ecc/ecc_verify_hash.c  63  Info 838: Previously assigned value to variable 'mp' has not been used
                                                                       _
                                  LTC_ASN1_EOL, 0UL, NULL)) != CRYPT_OK) {
src/pk/ecc/ecc_verify_hash.c  186  Info 838: Previously assigned value to variable 'err' has not been used

--- Module:   src/pk/dsa/dsa_decrypt_key.c (C)
                                                         _
   y = MIN(mp_unsigned_bin_size(key->p) + 1, MAXBLOCKSIZE);
src/pk/dsa/dsa_decrypt_key.c  95  Warning 666: Expression with side effects passed to repeated parameter 1 in macro 'MIN'

--- Module:   src/pk/dsa/dsa_verify_hash.c (C)
                                                                           _
                                  LTC_ASN1_EOL,     0UL, NULL)) != CRYPT_OK) {
src/pk/dsa/dsa_verify_hash.c  112  Info 838: Previously assigned value to variable 'err' has not been used

--- Module:   src/pk/pkcs1/pkcs_1_oaep_decode.c (C)
             _
   return err;
[...]

PR #202, opened 2017-04-30
more linting

fperrad:


karel-m:
Ad `Unusual use of a Boolean expression`  👍 

The stuff related to dh_static goes back to 2bdebb3 and here is the discussion/reasons https://github.com/libtom/libtomcrypt/pull/37 why it is as it is.

sjaeckel:
The stuff related to dh also isn't mentioned anywhere in the linter-outputs in #199 ...

fperrad:
rebased with only [Unusual use of a Boolean expression](https://github.com/libtom/libtomcrypt/pull/202/commits/a52b586ed2be0cdd9e359fbd4797d00af81c1ae4)

PR #203, opened 2017-04-30
Improve hashsum

sjaeckel:
Improved hashsum to be minimally compatible for creating hash-files that can be verified by the `{shaX,md5,...}sum` family resp. to verify files which were created with one of these tools


karel-m:
I agree with renaming `crypt` to `ltcrypt`, I would even agree with renaming `hashsum` to `lthash` (or something like that).

Using `/usr/local/` instead of `/usr/` is also a good idea.

I am only not sure whether `ltcrypt` is worth installing at all as its is nowadays IMO a bit weak. It does only a plain encryption without an integrity check, and considering the fact that the key/password is entered by the user from terminal, one would expect utilizing pbkdf2, not only a plain sha256. Looking at the arsenal we have in `src/authenc` it would be better to reimplement `ltcrypt` with e.g. GCM or ChaCha20Poly1305.

PR #204, opened 2017-05-01
fix MSVC compile errors

momo5502:
The semicolon after the LTC_MUTEX_TYPE macro causes compile errors if LTC_PTHREAD is not defined


karel-m:
this is duplicate to #200

karel-m:
now fixed in develop

Issue #209, opened 2017-05-02
AIX failures


karel-m:
While working on #208 I have spent some time with esoteric compilers from IBM and HP.

Here are some maybe interesting warnings/errors I have experienced with IBM C compiler:

```
"src/modes/xts/xts_test.c", line 28.9: 1506-068 (W) Operation between types "void*" and "int(*)(const unsigned char*,unsigned char*,unsigned long,unsigned char*,union Symmetric_key*,union Symmetric_key*)" is not allowed.
"src/modes/xts/xts_test.c", line 35.52: 1506-068 (W) Operation between types "int(*)(const unsigned char*,unsigned char*,unsigned long,unsigned char*,union Symmetric_key*,union Symmetric_key*)" and "void*" is not allowed.
"src/modes/xts/xts_test.c", line 53.9: 1506-068 (W) Operation between types "void*" and "int(*)(const unsigned char*,unsigned char*,unsigned long,unsigned char*,union Symmetric_key*,union Symmetric_key*)" is not allowed.
"src/modes/xts/xts_test.c", line 60.52: 1506-068 (W) Operation between types "int(*)(const unsigned char*,unsigned char*,unsigned long,unsigned char*,union Symmetric_key*,union Symmetric_key*)" and "void*" is not allowed.

"src/pk/dh/dh_static.h", line 73.1: 1506-277 (S) Syntax error: possible missing ';' or ','?
"src/pk/dh/dh_static.h", line 72.8: 1506-485 (S) Parameter declaration list is incompatible with declarator for inline.
"src/pk/dh/dh_static.h", line 90.15: 1506-045 (S) Undeclared identifier src.
"src/pk/dh/dh_static.h", line 99.8: 1506-045 (S) Undeclared identifier section.
"src/pk/dh/dh_static.h", line 99.34: 1506-045 (S) Undeclared identifier subsection.
```

sjaeckel:
> ```
> "src/modes/xts/xts_test.c", line 28.9: 1506-068 (W) Operation between types "void*" and "int(*)(const unsigned char*,unsigned char*,unsigned long,unsigned char*,union Symmetric_key*,union Symmetric_key*)" is not allowed.
> ...
> ```

very true

> ```
> "src/pk/dh/dh_static.h", line 73.1: 1506-277 (S) Syntax error: possible missing ';' or ','?
> ...
> ```

ugh, that compiler seems to do fancy stuff for inline functions...

karel-m:
[...]

PR #210, opened 2017-05-02
IBM xlc compiler does not like "static inline" (related to #209)

karel-m:


karel-m:
How did I manage this: `coverage decreased (-0.4%) to 90.026%` by bcce9e0 ?

sjaeckel:
as all jobs have this decrease I get the impression coveralls changed something internally...

PR #214, opened 2017-05-08
use codecov.io instead of coveralls for code coverage

sjaeckel:


sjaeckel:
not needed for now with the latest changes where we're able to re-use the lcov info to feed coveralls

PR #219, opened 2017-05-12
Introduce testprof/common.c

sjaeckel:
I introduced `testprof/common.c` which is included in the test library and can also be used by the demos.
The target was to remove duplicate code from the demos and tests.


karel-m:
Copy of my comments from another issue:

ad `common.c` and related stuff

Most of all I do not like the way we already mix 3 IMO **separate** things:
* the library
* the tests
* the utilities/demos

The proposed `common.(c|h)` makes IMO tests and demos even more (inter)linked.

In my opinion:
* If you look at `x86_prof.c` it already is a set of misc functions shared by several `*_test.c` files. So perhaps we can rename it  to `common.c` and add some extra stuff, maybe also merge `test_driver.c` into the new `common.c`.
* The library (== `./src/`) should be just a library (== all that goes to `libtomcrypt.a`). As I mentioned somewhere else I do not like having `<whatever>_test` function in "production" .a library
* For example I do not like the part `#if defined(LTC_TEST) && defined(LTC_TEST_DBG)` in `tomcrypt_misc.h` - this kind of stuff should be only in `testprof/tomcrypt_test.h` and all `<whatever>_test.c` files should include `tomcrypt_test.h`.
* Simply all test stuff should go to `libtomcrypt_prof.a` and should be declared in `tomcrypt_test.h` and ideally should be located in `./testprof/` (we might consider renaming all of them to ./tests/ + libtomcrypt_tests.a + libtomcrypt_tests.h) - all of those should be used only for `make test && ./test` (`libtomcrypt_prof.a` should not be installed, not even as a separate target `make install_test`)
* Therefore I also think that `./demos/test.c` + `./demos/timing.c` should better be placed in `./testprof/` not in `./demos/`
* I prefer the stuff in `./demos` to be only simple utilities (mostly implemented in a single .c file) which are linked only with the main library (libtomcrypt.a), not with the test library, and do not use anything from `./testprof/` dir (namely `tomcrypt_test.h`)

sjaeckel:
> * If you look at `x86_prof.c` it already is a set of misc functions shared by several `*_test.c` files. So perhaps we can rename it  to `common.c` and add some extra stuff, maybe also merge `test_driver.c` into the new `common.c`.

Sure, sounds perfectly fine!
We'd have to find a compromise for the shared code parts of the demo's and the stuff in `testprof/` (c.f. my answer to your last point)

> * The library (== `./src/`) should be just a library (== all that goes to `libtomcrypt.a`). As I mentioned somewhere else I do not like having `<whatever>_test` function in "production" .a library

As I also already mentioned there, I'm totally against removing the self-test functionality of the algorithms from the production library.

Why? Sometimes it's simply required by standardization to run these tests on device start-up (IIRC e.g. [PCI](https://www.pcisecuritystandards.org/)).

> * For example I do not like the part `#if defined(LTC_TEST) && defined(LTC_TEST_DBG)` in `tomcrypt_misc.h` - this kind of stuff should be only in `testprof/tomcrypt_test.h` and all `<whatever>_test.c` files should include `tomcrypt_test.h`.

True. As I'd prefer to have the self-tests inside the main lib I propose to move these two functions to the main lib as well. (Then we should also start to think about a list of exported symbols as done in tfm as I doubt it's useful to export `print_hex()` or we'd have a copy of `print_hex()` in the tests and declare it static in the main lib and only use it from `compare_testvector()`).

> * Simply all test stuff should go to `libtomcrypt_prof.a` and should be declared in `tomcrypt_test.h` and ideally should be located in `./testprof/` (we might consider renaming all of them to ./tests/ + libtomcrypt_tests.a + libtomcrypt_tests.h) - all of those should be used only for `make test && ./test` (`libtomcrypt_prof.a` should not be installed, not even as a separate target `make install_test`)

Also sounds perfectly fine.

> * Therefore I also think that `./demos/test.c` + `./demos/timing.c` should better be placed in `./testprof/` not in `./demos/`

True, I thought the same multiple times.

> * I prefer the stuff in `./demos` to be only simple utilities (mostly implemented in a single .c file) which are linked only with the main library (libtomcrypt.a), not with the test library, and do not use anything from `./testprof/` dir (namely `tomcrypt_test.h`)

Also very true. I still think that a common part could be useful to share code like the registration of ciphers, but I don't think it'd be useful to add something like `crypt_register_all_ciphers()` to the main library as it's simply not required for the applications that use the library.

And some additions:

* I thought about removing `libtomcrypt_prof.a` as it's only used by `demo` and `timing` and we could simply link the two binaries against the sources of the tests (then we could also think about leaving `make install_test` as it'd only additionally install the two binaries).
* Removing the self-tests is currently not a way to go as:
    1. all algorithms have the `test()` functions in their descriptors and if we remove the functionality from the library we should also remove these functions from the descriptors which would break API/ABI more than we want for now, yadda yadda
    2. if we work hard to code something together I could think of a way to have either a `libtomcrypt_testable.a` which has the same (or similar) descriptors as the current version or a second `libtomcrypt_tests.a` which provides only the self-tests and still has a useable API (e.g. `int crypt_test_cipher(ltc_cipher_descriptor*)` but TBH I don't like it as:
    a. I think it tears core-functionality apart which should be together (the algorithms and the self-tests) and if someone wants the library without self-tests, well then he should re-build it like that.
    b. providing `crypt_test_cipher()` instead of having the tests inside the descriptor means we'd solve dependency-resolving by code and not by structure and that's against the nature of computing in my opinion

sjaeckel:
> > * For example I do not like the part `#if defined(LTC_TEST) && defined(LTC_TEST_DBG)` in `tomcrypt_misc.h` - this kind of stuff should be only in `testprof/tomcrypt_test.h` and all `<whatever>_test.c` files should include `tomcrypt_test.h`.
> 
> True. As I'd prefer to have the self-tests inside the main lib I propose to move these two functions to the main lib as well. (Then we should also start to think about a list of exported symbols as done in tfm as I doubt it's useful to export `print_hex()` or we'd have a copy of `print_hex()` in the tests and declare it static in the main lib and only use it from `compare_testvector()`).

* Moved `compare_testvector()` to misc and duplicated `print_hex()`.

> > * Simply all test stuff should go to `libtomcrypt_prof.a` and should be declared in `tomcrypt_test.h` and ideally should be located in `./testprof/` (we might consider renaming all of them to ./tests/ + libtomcrypt_tests.a + libtomcrypt_tests.h) - all of those should be used only for `make test && ./test` (`libtomcrypt_prof.a` should not be installed, not even as a separate target `make install_test`)
> 
> Also sounds perfectly fine.
> 
> > * Therefore I also think that `./demos/test.c` + `./demos/timing.c` should better be placed in `./testprof/` not in `./demos/`
> 
> True, I thought the same multiple times.
> 
> > * I prefer the stuff in `./demos` to be only simple utilities (mostly implemented in a single .c file) which are linked only with the main library (libtomcrypt.a), not with the test library, and do not use anything from `./testprof/` dir (namely `tomcrypt_test.h`)
> 
> Also very true. I still think that a common part could be useful to share code like the registration of ciphers, but I don't think it'd be useful to add something like `crypt_register_all_ciphers()` to the main library as it's simply not required for the applications that use the library.
> 
> And some additions:
> 
> * I thought about removing `libtomcrypt_prof.a` as it's only used by `demo` and `timing` and we could simply link the two binaries against the sources of the tests (then we could also think about leaving `make install_test` as it'd only additionally install the two binaries).

* Renamed `./testprof/` to `./tests/`
* Moved all timing tests to `demo/timing.c`
* Removed libtomcrypt_prof library and simply link all tests together

Additionally

* Added a build-run with debugging options enabled

karel-m:
What about renaming `der_tests` > `der_test` to keep `*_test` names consistent?

sjaeckel:
Sure

karel-m:
Ad `ecc_test`:
* we have `tests/ecc_test.c` but it contains `int ecc_tests(void) { return CRYPT_NOP; }`
* then we have `src/pk/ecc/ecc_test.c` which contains `int ecc_test(void)`

I am for moving all ECC tests to `tests/ecc_test.c`

karel-m:
Ad `compare_testvector`

The following part of `ccm_test.c` should be IMO converted to `compare_testvector`
```C
      if (XMEMCMP(buf, tests[x].ct, tests[x].ptlen)) {
#if defined(LTC_TEST_DBG)
         printf("\n%d: x=%lu y=%lu\n", __LINE__, x, y);
         print_hex("ct is    ", buf, tests[x].ptlen);
         print_hex("ct should", tests[x].ct, tests[x].ptlen);
#endif
         return CRYPT_FAIL_TESTVECTOR;
      }
```

Similarly in `hmac_test.c`
```C
        if(XMEMCMP(digest, cases[i].digest, (size_t)hash_descriptor[hash].hashsize) != 0)  {
            failed++;
#ifdef LTC_TEST_DBG
          {
            printf("\nHMAC-%s test %s: Failed\n", cases[i].algo, cases[i].num);
            print_hex("is", digest, hash_descriptor[hash].hashsize);
            print_hex("should", cases[i].digest, hash_descriptor[hash].hashsize);
            return CRYPT_FAIL_TESTVECTOR;
          } 
```

And `pkcs_5_test.c`
```
        else if(XMEMCMP(DK, cases_5_2[i].DK, (size_t)cases_5_2[i].dkLen) != 0) {
            printf("\nPKCS#5_2 test #%d: Failed/3\n", i);
#ifdef LTC_TEST_DBG
            print_hex("is", DK, cases_5_2[i].dkLen);
            print_hex("should", cases_5_2[i].DK, cases_5_2[i].dkLen);
#endif
            ++failed; 
```

And `xts_test.c`
```
            if (XMEMCMP(OUT, tests[i].CTX, tests[i].PTLEN)) {
#ifdef LTC_TEST_DBG
               printf("\nTestcase #%d with original length %lu and half of it "
                      "%lu\n",
                      i, tests[i].PTLEN, len);
               printf("\nencrypt\n");
               print_hex("should", tests[i].CTX, tests[i].PTLEN);
               print_hex("is", OUT, tests[i].PTLEN);
#endif 
```

And `rsa_test.c`
```C
   if (len != sizeof(openssl_public_rsa) || memcmp(buf, openssl_public_rsa, len)) {
      fprintf(stderr, "RSA(public) SSL public X.509 export failed to match OpenSSL output\n");
      print_hex("should", openssl_public_rsa, sizeof(openssl_public_rsa));
      print_hex("is", buf, len);
      return 1;
   } 
```

And `base64_test.c` probably can be without `print_hex` as `compare_testvector` already has printing capabilities
```C
   if (compare_testvector(tmp, l2, in, l2, "relaxed base64 decoding", -1)) {
       print_hex("input ", out, l1);
       return 1;
   } 
```

In the end there is no reason to use `print_hex` from `./src/*` and it can be left out from `tomcrypt_misc.h`.

On top of that I would change the corresponding part of `tomcrypt_misc.h` to simple:
```C
int compare_testvector(const void* is, const unsigned long is_len, const void* should, const unsigned long should_len, const char* what, int which); 
```

And move all `#if ..` magic to `compare_testvector.c` like:
```C
int compare_testvector(const void* is, const unsigned long is_len, const void* should, const unsigned long should_len, const char* what, int which)
{
   int res = 0;
   if(is_len != should_len)
      res = is_len > should_len ? -1 : 1;
   else
      res = XMEMCMP(is, should, MAX(is_len, should_len));

#if defined(LTC_TEST) && defined(LTC_TEST_DBG) 
   if (res != 0) {
      fprintf(stderr, "Testvector #%i of %s failed:\n", which, what);
      _print_hex("SHOULD", should, should_len);
      _print_hex("IS    ", is, is_len);
   }
#endif

   return res;
}
```

karel-m:
Just a summary why I do not like using `tests/common.c` + `tests/common.h` from `demo/*.c`
* I consider `tests/*` to be something we need to run the test suite
* Whereas `demo/*` is something we have prepared for our users to quickly get the basic idea how to use libtomcrypt
* I would prefer to have in `demo/*.c` only C files, which can be compiled and linked with "pure" libtomcrypt (+libtommath)
* The files `demo/*.c` should be something that anybody can just copy & paste & hack and quickly have some working code (which we made a bit complicated)

I do not like it, but it is not a blocker for me. I only want you to know :)

karel-m:
Ad `crypt_register_all_ciphers` mentioned earlier in this thread
* it might be (sometimes) handy
* see https://metacpan.org/source/MIK/CryptX-0.047/CryptX.xs#L287

sjaeckel:
> Ad ecc_test:
> ...
> I am for moving all ECC tests to tests/ecc_test.c

We shouldn't do that as we said we'd like to keep compatibility at a max and that would remove an existing API function. Also it's not the case that `tests/ecc_test.c` only invokes the tests from `src/pk/ecc/ecc_test.c` but extends them.

sjaeckel:
> Ad `compare_testvector`

Removed all invocations of `print_hex()` in the library and made `compare_testvector()` always a function

> * I would prefer to have in `demo/*.c` only C files, which can be compiled and linked with "pure" libtomcrypt (+libtommath)
> * The files `demo/*.c` should be something that anybody can just copy & paste & hack and quickly have some working code (which we made a bit complicated)

These points got me, cleaned it up.

> Ad `crypt_register_all_ciphers` mentioned earlier in this thread

Added `register_all_{ciphers,hashes,prngs}()`

sjaeckel:
FYI I modified all the posix-make usages again as `$?` doesn't exactly behave as I thought...

PR #198, opened 2017-06-08
Travis-CI: a new build with -DLTC_PTHREAD

sjaeckel:
_This was originally raised by @karel-m but the issue to PR conversion changed the original author :-)_

@sjaeckel could you please add a new travis build with defined `-DLTC_PTHREAD`, so that we at least basically test the MUTEX related code in prngs?


sjaeckel:
is this one fine?

PR #227, opened 2017-06-12
DH facelift part 2 (dh_make_key_ex & co.)

karel-m:
The idea of `dh_make_key_ex` (which allows using custom DH group params) was suggested and implemented by @lkinley as a contribution to my libtomcrypt perl bindings.

It is simply about adding:
```C
int dh_make_key_ex(prng_state *prng, int wprng, char *prime_hex, char *base_hex, dh_key *key)
```

Unfortunately the implementation turned into a bunch of troubles&workarounds like changes to dh_key structure changes (incompatible) to import/export format etc.

Some questions:
* How far can we go in breaking DH API? (currently only in develop for some time, not released yet)
* Is the format used for DH key export/import intended to compatible with some specification? (this PR changed it in an incompatible way)


sjaeckel:
> How far can we go in breaking DH API? (currently only in develop for some time, not released yet)

As it wasn't in 1.17 we can do whatever is necessary

> Is the format used for DH key export/import intended to compatible with some specification? (this PR changed it in an incompatible way)

I guess that's why there's _the library version_<sup>TM</sup> included in the header... but I doubt someone's using that functionality as DH was removed in 1.06. I'm not even sure if these functions make sense anymore... I think it'd be more useful to ex-/import the keys from/to a standard format...

karel-m:
The question is what's the standard format for exporting/importing DH keys?

If we want to have our custom format I would prefer some simple ASN.1
sequence than the custom bignum dump we currently use in our DH code.

sjaeckel:
For the public key we have `SubjectPublicKeyInfo` https://tools.ietf.org/html/rfc3279#section-2.3.3

For the private key I'd propose something like
```
Sequence {
  SubjectPublicKeyInfo publicKey
  Integer privateKey
}
```
I think we should also provide a way to import DHParam's as generated by OpenSSL at one point even if it's trivial :)
```
Sequence {
  Integer prime
  Integer group
}
```
If it's too much work, remove the ex-/import ;-)

karel-m:
Here is my proposal of new DH related formats:

1/ key format used by `dh_export` / `dh_import`

Considering the fact that we want to allow custom DH group params (via `dh_make_key_ex`) we also need to be able to export `base` + `prime`.

private key
```C
   der_encode_sequence_multi(out, outlen,
                             LTC_ASN1_BIT_STRING, 1UL, flags,     /* flags[0] = 1 for type == PK_PRIVATE */
                             LTC_ASN1_INTEGER,    1UL, key->base,
                             LTC_ASN1_INTEGER,    1UL, key->prime,
                             LTC_ASN1_INTEGER,    1UL, key->x,    /* no need to export key->y as we can compute it during import */
                             LTC_ASN1_EOL,        0UL, NULL);
```

public key
```C
   der_encode_sequence_multi(out, outlen,
                             LTC_ASN1_BIT_STRING, 1UL, flags,     /* flags[0] = 0 for type == PK_PUBLIC */
                             LTC_ASN1_INTEGER,    1UL, key->base,
                             LTC_ASN1_INTEGER,    1UL, key->prime,
                             LTC_ASN1_INTEGER,    1UL, key->y,
                             LTC_ASN1_EOL,        0UL, NULL);
```

2/ `dh_sign_hash` / `dh_verify_hash` (this one is the easiest)
```C
   der_encode_sequence_multi(out, outlen,
                             LTC_ASN1_INTEGER,    1UL, a,
                             LTC_ASN1_INTEGER,    1UL, b,
                             LTC_ASN1_EOL,        0UL, NULL);
```

3/ `dh_encrypt_key` / `dh_decrypt_key`

```C
   der_encode_sequence_multi(out, outlen,
                             LTC_ASN1_OBJECT_IDENTIFIER, hash_descriptor[hash].OIDlen, hash_descriptor[hash].OID,
                             LTC_ASN1_OCTET_STRING,      pubkeysize,                   pub_expt, /* public key */
                             LTC_ASN1_OCTET_STRING,      inlen,                        skey,     /* encrypted data */
                             LTC_ASN1_EOL,               0UL,                          NULL);
```

The `pub_expt` is an exported public key and it is desirable to keep it as short as possible.

Unfortunately the proposed changes to `dh_export` / `dh_import` are making the exported data longer, much longer, as it includes `prime` (up to 1000 bytes) and `base`.

So we perhaps need something like:
* `dh_export_raw` / `dh_import_raw`
* which will export the key with the output as short as possible
* it means exporting only the public `key->y` or private `key->x` as a raw binary data

The interface might look like:
* `dh_export_raw(unsigned char *out, unsigned long *outlen, int type, dh_key *key)`
* `dh_import_raw(const unsigned char *in, unsigned long inlen, int type, char *prime_hex, char *base_hex, dh_key *key)`

After switching DH formats to ASN.1 as proposed above we can get rid of `dh_static.h` + `dh_static.c` (macros OUTPUT_BIGNUM/INPUT_BIGNUM & co. will not be needed anymore).

sjaeckel:
> Here is my proposal of new DH related formats:
> 
> 1/ key format used by `dh_export` / `dh_import`
> 
> Considering the fact that we want to allow custom DH group params (via `dh_make_key_ex`) we also need to be able to export `base` + `prime`.
> 
> private key
> ```C
>    der_encode_sequence_multi(out, outlen,
>                              LTC_ASN1_BIT_STRING, 1UL, flags,     /* flags[0] = 1 for type == PK_PRIVATE */
>                              LTC_ASN1_INTEGER,    1UL, key->base,
>                              LTC_ASN1_INTEGER,    1UL, key->prime,
>                              LTC_ASN1_INTEGER,    1UL, key->x,    /* no need to export key->y as we can compute it during import */
>                              LTC_ASN1_EOL,        0UL, NULL);
> ```

as there's no standard way to do this, fine by me.

> public key
> ```C
>    der_encode_sequence_multi(out, outlen,
>                              LTC_ASN1_BIT_STRING, 1UL, flags,     /* flags[0] = 0 for type == PK_PUBLIC */
>                              LTC_ASN1_INTEGER,    1UL, key->base,
>                              LTC_ASN1_INTEGER,    1UL, key->prime,
>                              LTC_ASN1_INTEGER,    1UL, key->y,
>                              LTC_ASN1_EOL,        0UL, NULL);
> ```

I just realized that for the `SubjectPublicKeyInfo` we'd need `q       INTEGER,           -- factor of p-1`... can we get that somehow?
If you think it's too much work it's fine as you did it.

> 2/ `dh_sign_hash` / `dh_verify_hash` (this one is the easiest)
> ```C
>    der_encode_sequence_multi(out, outlen,
>                              LTC_ASN1_INTEGER,    1UL, a,
>                              LTC_ASN1_INTEGER,    1UL, b,
>                              LTC_ASN1_EOL,        0UL, NULL);
> ```
>
> 3/ `dh_encrypt_key` / `dh_decrypt_key`
> 
> ```C
>    der_encode_sequence_multi(out, outlen,
>                              LTC_ASN1_OBJECT_IDENTIFIER, hash_descriptor[hash].OIDlen, hash_descriptor[hash].OID,
>                              LTC_ASN1_OCTET_STRING,      pubkeysize,                   pub_expt, /* public key */
>                              LTC_ASN1_OCTET_STRING,      inlen,                        skey,     /* encrypted data */
>                              LTC_ASN1_EOL,               0UL,                          NULL);
> ```

asn1-all-the-things

> The `pub_expt` is an exported public key and it is desirable to keep it as short as possible.
> 
> Unfortunately the proposed changes to `dh_export` / `dh_import` are making the exported data longer, much longer, as it includes `prime` (up to 1000 bytes) and `base`.

If you really care about the size how about using a compressed format for the modp groups? We could use OID `1.3.6.1.4.1.272.4.26.5.1.22` to indicate the group.

something like:
```
flags ::= BIT STRING[2] -- [0] : 0=public, 1=private; [1] : 0=plain, 1=compressed

modpGroup ::= SEQUENCE {
      ipsecPeerIkeGroup OBJECT IDENTIFIER ::= {
         iso(1) identified-organization(3) dod(6) internet(1) private(4) enterprise(1) 272 bibo(4) ipsec(26) 
         ipsecPeerTable(5) ipsecPeerEntry(1) ipsecPeerIkeGroup(22) },
      group INTEGER }

CompressedDHKey ::= SEQUENCE {
   group modpGroup,
   xy INTEGER -- either x or y depending on flag[0]
}
```



> 
> So we perhaps need something like:
> * `dh_export_raw` / `dh_import_raw`
> * which will export the key with the output as short as possible
> * it means exporting only the public `key->y` or private `key->x` as a raw binary data
> 
> The interface might look like:
> * `dh_export_raw(unsigned char *out, unsigned long *outlen, int type, dh_key *key)`
> * `dh_import_raw(const unsigned char *in, unsigned long inlen, int type, char *prime_hex, char *base_hex, dh_key *key)`
> 
> After switching DH formats to ASN.1 as proposed above we can get rid of `dh_static.h` + `dh_static.c` (macros OUTPUT_BIGNUM/INPUT_BIGNUM & co. will not be needed anymore).
>

karel-m:
The formats above as well as the the implementation is heavily inspired by what we have in `ecc_*`.

**As for the formats** I'd keep what I have already committed&pushed. The idea with OID does not fit quite well as we want to allow using/generating keys with custom DH parameters (without assigned OID).

But I am not that strongly against format changes (if they make sense).

**As for the API changes:**

Old functions:
```C
int dh_export(out, outlen, type, key);
int dh_import(in, inlen, key);
int dh_make_key(prng, wprng, groupsize, key);
void dh_free(key);
int dh_shared_secret(dh_key private_key, dh_key public_key, out, outlen);
int dh_encrypt_key(in, keylen, out, outlen, prng, wprng, hash, dh_key key);
int dh_decrypt_key(in, inlen, out, outlen, key);
int dh_sign_hash(in, inlen, out, outlen, prng, wprng, key);
int dh_verify_hash(sig, siglen, hash, hashlen, stat, key);
```

Renamed functions:
```C
void dh_groupsizes(low, high);
int dh_get_groupsize(key);
```

Newly added functions:
```C
int dh_groupsize_to_keysize(groupsize);
int dh_make_key_ex(prng, wprng, prime_hex, base_hex, key);
int dh_check_pubkey(key);
int dh_export_raw(out, outlen, type, key);
int dh_import_raw(in, inlen, type, prime_hex, base_hex, key);
```

The whole `src/pk/dh/*` was completely re-worked (no `dh_static.*` functions split into a separate `*.c` files). Therefore for review simply go to https://github.com/libtom/libtomcrypt/tree/pr/dh_make_key_ex/src/pk/dh and check the new files.

And on top of all of that - Travic-CI is green :)

sjaeckel:
> The formats above as well as the the implementation is heavily inspired by what we have in ecc_*.

I see.

> ...The idea with OID does not fit quite well as we want to allow using/generating keys with custom DH parameters (without assigned OID).

well that's for sure.
I guessed most of the implementations will use one of the standard groups, therefore it would make sense to implement something as proposed above.

> int dh_export_raw(out, outlen, type, key);
> int dh_import_raw(in, inlen, type, prime_hex, base_hex, key);

This API doesn't make sense in my opinion as I can't see a way to import what you exported, besides if you store your prime and base somewhere else... (c.f. the list a bit further down)

> And on top of all of that - Travis-CI is green :)

Looks like we need a travis job that checks if API changes/new API's are also reflected in the documentation :-P

Generally good work, I would've expected a bit more established-standards compliance when it comes to contact with the outside world. I.e.

1. expecting an ASN.1 sequence as created by `openssl dhparams` as input to `dh_make_key_ex()`. Then we could also use the same as parameter to `dh_import_raw()`
2. exporting the public key as `SubjectPublicKeyInfo` (if possible, c.f. https://github.com/libtom/libtomcrypt/pull/227#issuecomment-308826042)
3. what's `dh_{de,en}crypt_key()` again for? correct me if I'm wrong... I guess it does its job fine, but as it's kind of a ltc-private protocol it doesn't really make sense to have it in this library which normally only provides primitives, or?

I realized most of these things only when you started working on it (I shouldn't have taken it for granted the way it was).
I'm sure you won't like my proposal as you put a lot of effort in there, but how about limiting DH for now to its basic functionality?

```
int dh_make_key(prng_state *prng, int wprng, int groupsize, dh_key *key);
void dh_free(dh_key *key);

int dh_export(unsigned char *out, unsigned long *outlen, int type, dh_key *key);
int dh_import(const unsigned char *in, unsigned long inlen, dh_key *key);

int dh_shared_secret(dh_key        *private_key, dh_key        *public_key,
                     unsigned char *out,         unsigned long *outlen);
```
I also think that 
- we should add a "version INTEGER -- currently only 0" as first field to the exported DH keys
- we should start using local header files for internal functions (e.g. `dh_check_pubkey()`), but I'll add that to #228 

TBH thanks for starting working on this, I only realized now how non-sensical some parts of it really are (and now I also realized what tom meant when he dropped dh because it was a PITA to maintain...)

karel-m:
> expecting an ASN.1 sequence as created by openssl dhparams as input to dh_make_key_ex(). Then we could also use the same as parameter to dh_import_raw()

This one was easy, see `dh_make_key_dhparam`.

> exporting the public key as SubjectPublicKeyInfo

According to https://tools.ietf.org/html/rfc3279#section-2.3.3 we need `q` (factor of `p-1`) which we do not have and I do not know about a way how to get it (and unfortunately it is mandatory). FYI `q` is not even a part of dhparam generated by openssl. Which means that we are not able to follow rfc3279 when exporting public DH keys. Having non-standard SubjectPublicKeyInfo-based format is basically the same as having completely proprietary format.

> what's dh_{de,en}crypt_key() again for? 

Well, I agree that these are a bit unusual but we have something similar in `ecc_*` so why not to have them here.

> how about limiting DH for now to its basic functionality?

I am not using DH that much myself; however for my perl bindings I need nearly all of the functions I have implemented, including `dh_make_key_ex`, `dh_export_raw`, `dh_import_raw`, `dh_check_pubkey`. I can imagine that we "replace" `dh_import_raw` with something like `dh_import_radix` (like already existing `rsa_import_radix` + `dsa_import_radix`), but I basically need them. @lkinley is the user (and contributor) of perl module Crypt::PK::DH so he can perhaps say a bit more about his use case(s).

Here are links to "raw" import/export I have in my perl bindings
* https://metacpan.org/pod/Crypt::PK::DH#import_key_raw
* https://metacpan.org/pod/Crypt::PK::DH#export_key_raw

BTW `ecc_import_raw` and `ecc_export_raw` are also part on my ECC enhancements as i need them for:
* https://metacpan.org/pod/Crypt::PK::ECC#import_key_raw
* https://metacpan.org/pod/Crypt::PK::ECC#export_key_raw

sjaeckel:
> > expecting an ASN.1 sequence as created by openssl dhparams as input to dh_make_key_ex(). Then we could also use the same as parameter to dh_import_raw()
> 
> This one was easy, see `dh_make_key_dhparam`.

True, feel free to squash the fixups together and force-push :-D

> > exporting the public key as SubjectPublicKeyInfo
> 
> According to https://tools.ietf.org/html/rfc3279#section-2.3.3 we need `q` (factor of `p-1`) which we do not have and I do not know about a way how to get it (and unfortunately it is mandatory). FYI `q` is not even a part of dhparam generated by openssl. Which means that we are not able to follow rfc3279 when exporting public DH keys. Having non-standard SubjectPublicKeyInfo-based format is basically the same as having completely proprietary format.

IIUC it boils down to using the DSA key-generation where we get `q`...
I came to the conclusion when I saw that `openssl dhparam` offers an option `-dsaparam` which simply converts the DSA key and suddenly the DH key has 3 `INTEGER`s...

> > what's dh_{de,en}crypt_key() again for? 
> 
> Well, I agree that these are a bit unusual but we have something similar in `ecc_*` so why not to have them here.

It's the same with `ecc_verify_hash()` where we have now a `ecc_verify_hash_raw()` because the old version was way too special...
I don't think `dh_{de,en}crypt_key()` are useful and therefor I'd say we remove them if @lkinley doesn't need them.
 
> > how about limiting DH for now to its basic functionality?
> 
> I am not using DH that much myself; however for my perl bindings I need nearly all of the functions I have implemented, including `dh_make_key_ex`, `dh_export_raw`, `dh_import_raw`, `dh_check_pubkey`. I can imagine that we "replace" `dh_import_raw` with something like `dh_import_radix` (like already existing `rsa_import_radix` + `dsa_import_radix`), but I basically need them. @lkinley is the user (and contributor) of perl module Crypt::PK::DH so he can perhaps say a bit more about his use case(s).

I couldn't find a reference to e.g. `dh_check_pubkey()` but I also didn't search very hard... 
`dh_make_key_ex()` seems to be only required because the MODP groups aren't in ltc yet... yeah sure you could also pass the two strings as parameters in your `generate_key` function but _please correct me if I'm wrong_ I doubt there's any software that originally exports its groups as hex-string...
`{ex,im}port_raw()` are perfectly fine.

@lkinley can you please say which API's you really need?

lkinley:
I am using Crypt::PK::DH in the module Net::SSH::Perl for the key exchange portion of the SSH protocol.  To implement the various Kex methods, it needs to be able to specify the p and g values.

With group 1, 14, 16 and 18 key exchange, the p and g values are static.  In group exchange, the SSH server generates p and g and sends them to the client.

I need:
- generate DH from p, g
- export raw public key
- import raw public key
- generate shared secret using imported raw public key

karel-m:
You also wanted a pubkey validation for some reason or not?

sjaeckel:
> You also wanted a pubkey validation for some reason or not?

I think the validation is better placed internally (as it's already done now)

lkinley:
Yes, I believe the pub key validity is checked on import.

sjaeckel:
As of https://github.com/libtom/libtomcrypt/pull/227#issuecomment-309053395 I think we can safely remove
* `dh_{de,en}crypt_key()`
* `dh_{sign,verify}_hash()`
* most of the utility functions away from the public API, but we can do that in the context of #228 

Btw. how about moving `dh_{de,en}crypt_key()` & `dh_{sign,verify}_hash()` to a demo?

> we should add a "version INTEGER -- currently only 0" as first field to the exported DH keys

what do you think about that? it'd allow easier implementation later-on of e.g. the compressed approach

karel-m:
I really do need `dh_import_raw` or `dh_import_radix`.

Providing the fact that there is no standardized way for storing/exporting DH keys, what would you recommend libtomcrypt users who want to import a public key to derive a shared secret via DH? According to your proposal we will offer only `dh_import` which is able to import only our proprietary format.

src/pk/dh/dh_check_pubkey.c
```diff
@@ -0,0 +1,65 @@
...
+      goto error;
+   }
+   if (mp_cmp(key->y, p_minus1) != LTC_MP_LT || mp_cmp_d(key->y, 1) != LTC_MP_GT) {
+      err = CRYPT_INVALID_ARG;
+      goto error;
+   }
+
+   /* public key must have more than one bit set */
+   digit_count = mp_get_digit_count(key->y);
+   for (i = 0; i < digit_count; i++) {
```
karel-m: I agree, fixed

src/pk/dh/dh_check_pubkey.c
```diff
@@ -0,0 +1,65 @@
...
+      digit = mp_get_digit(key->y, i);
+      while (digit > 0) {
+         if (digit & 1) bits_set++;
+         digit >>= 1;
+      }
+   }
+   if (bits_set <= 1) {
```
karel-m: fixed

src/pk/dh/dh_decrypt_key.c
```diff
@@ -0,0 +1,143 @@
...
+   /* Success */
+   err = CRYPT_OK;
+
+LBL_ERR:
+#ifdef LTC_CLEAN_STACK
+   zeromem(pub_expt,      DH_BUF_SIZE);
+   zeromem(shared_secret, DH_BUF_SIZE);
+   zeromem(skey,          MAXBLOCKSIZE);
+#endif
+   XFREE(pub_expt);
```
karel-m: Sounds good. Only the name `LTC_CLEAN_STACK` is a bit misleading as we are not dealing with the stack in this case.

karel-m:
> we should add a "version INTEGER -- currently only 0" as first field to the exported DH keys

done

karel-m:
>  I think we can safely remove `dh_{de,en}crypt_key()` , `dh_{sign,verify}_hash()`

done

sjaeckel:
> I really do need `dh_import_raw` or `dh_import_radix`.

sure, if you need it go for it

src/pk/dh/dh_make_key.c
```diff
@@ -0,0 +1,192 @@
...
+ *
+ * The library is free for all purposes without any express
+ * guarantee it works.
+ */
+
+#include "tomcrypt.h"
+
+#ifdef LTC_MDH
+
+static int dh_groupsize_to_keysize_internal(int groupsize)
```
sjaeckel: I still think it doesn't make sense to mark internal functions (be it static or 'library-internal') in some way and if you think you have to do it, please don't suffix static functions with `_internal` but prefix with `_` (we had that discussion already)

src/headers/tomcrypt_pk.h
```diff
@@ -207,10 +207,13 @@ int dh_import_raw(unsigned char *in, unsigned long inlen, int type,
 int dh_shared_secret(dh_key        *private_key, dh_key        *public_key,
                      unsigned char *out,         unsigned long *outlen);
 
-int dh_check_pubkey(dh_key *key); /* INTERNAL ONLY */
-
+#ifdef LTC_SOURCE
```
sjaeckel: as also pointed out by @rofl0r in #228 please don't do it like that.

src/pk/dh/dh_make_key.c
karel-m: done

src/headers/tomcrypt_pk.h
sjaeckel: leave it like that, I'll clean it up!

sjaeckel:
looks like we can remove `addmod()` and `submod()` again from the `ltc_mp` (if you don't want to move `dh_{de,en}crypt_key()` & `dh_{sign,verify}_hash()` to a demo)

karel-m:
> ... if you don't want to move dh_{de,en}crypt_key() & dh_{sign,verify}_hash() to a demo

I am not against this idea, but maybe later, not in this PR.

src/pk/dh/dh_export.c
```diff
@@ -0,0 +1,62 @@
...
+  @param outlen [in/out] The max size and resulting size of the DH key
+  @param type   Which type of key (PK_PRIVATE or PK_PUBLIC)
+  @param key    The key you wish to export
+  @return CRYPT_OK if successful
+*/
+int dh_export(unsigned char *out, unsigned long *outlen, int type, dh_key *key)
+{
+   unsigned char flags[1];
+   int err;
+   unsigned long version = 1;
```
sjaeckel: IIUC common practice in the ASN.1 world is to start with version 0.
Any reason not to do it like that?

src/pk/dh/dh_import.c
```diff
@@ -0,0 +1,97 @@
...
+   /* find out what type of key it is */
+   err = der_decode_sequence_multi(in, inlen,
+                                   LTC_ASN1_SHORT_INTEGER, 1UL, &version,
+                                   LTC_ASN1_BIT_STRING, 1UL, &flags,
+                                   LTC_ASN1_EOL, 0UL, NULL);
+   if (err != CRYPT_OK) {
+      goto error;
+   }
+
+   if (version == 1) {
```
sjaeckel: c.f. `dh_export()`

src/pk/dh/dh_make_key.c
```diff
@@ -0,0 +1,196 @@
...
+         goto freebuf;
+      }
+      /* compute the y value - public key */
+      if ((err = mp_exptmod(key->base, key->x, key->prime, key->y)) != CRYPT_OK) {
+         goto freebuf;
+      }
+   } while (dh_check_pubkey(key) != CRYPT_OK);
+
+   /* success */
+   return CRYPT_OK;
```
sjaeckel: forgot to free `buf`

src/pk/dh/dh_make_key.c
```diff
@@ -0,0 +1,196 @@
...
+      }
+      /* load the x value - private key */
+      if ((err = mp_read_unsigned_bin(key->x, buf, keysize)) != CRYPT_OK) {
+         goto freebuf;
+      }
+      /* compute the y value - public key */
+      if ((err = mp_exptmod(key->base, key->x, key->prime, key->y)) != CRYPT_OK) {
+         goto freebuf;
+      }
+   } while (dh_check_pubkey(key) != CRYPT_OK);
```
sjaeckel: should we really retry forever?

demos/timing.c
```diff
@@ -899,29 +898,10 @@ static void time_dh(void)
...
        }
        t2 >>= 4;
        fprintf(stderr, "DH-%4lu make_key    took %15llu cycles\n", x*8, t2);
-
-       t2 = 0;
-       for (y = 0; y < 16; y++) {
-           t_start();
-           t1 = t_read();
-           z = sizeof(buf[1]);
-           if ((err = dh_encrypt_key(buf[0], 20, buf[1], &z, &yarrow_prng, find_prng("yarrow"), find_hash("sha1"),
```
sjaeckel: can't we easily replace this with `dh_shared_secret()`?

src/pk/dh/dh_export.c
karel-m: fixed

src/pk/dh/dh_import.c
karel-m: fixed

src/pk/dh/dh_make_key.c
karel-m: fixed

demos/timing.c
karel-m: depends how you define easily, but yes, it is just some work

karel-m:
I am still not sure about the interface of `dh_import_raw`

Now we have
```C
int dh_import_raw(unsigned char *in, unsigned long inlen, int type,
                  const char *prime_hex, const char *base_hex, dh_key *key)
```

Wouldn't it be better (more consistent) to change it to passing prime/base as raw octects:
```C
int dh_import_raw(unsigned char *in,    unsigned long inlen,
                  unsigned char *prime, unsigned long primelen,
                  unsigned char *base,  unsigned long baselen,
                  int type, dh_key *key)
```
?

If there is a demand for importing hexadecimals we can introduce `dh_import_radix` (similar to `rsa_import_radix` and `dsa_import_radix`)

sjaeckel:
> I am still not sure about the interface of `dh_import_raw`

Thanks for mentioning that, I had the same thought multiple times.

How about this?
```
int dh_import_radix(int radix,
                    unsigned char *in,    unsigned long inlen,
                    unsigned char *prime, unsigned long primelen,
                    unsigned char *base,  unsigned long baselen,
                    int type, dh_key *key)
{
  if (radix == 256) {
    /* import with mp_read_unsigned_bin() */
  } else {
    /* import with mp_read_radix() */
  }
}
```

src/pk/dh/dh_make_key.c
```diff
@@ -46,7 +46,7 @@ static int _dh_make_key(prng_state *prng, int wprng, void *prime, void *base, dh
 {
    unsigned char *buf;
    unsigned long keysize;
-   int err;
+   int err, max_iterations = 20;
```
sjaeckel: please add `#define DH_MAKE_KEY_MAX_ITERATIONS`

karel-m: fixed

karel-m:
You idea of enhanced `dh_import_radix` sounds interesting

sjaeckel:
> You idea of enhanced `dh_import_radix` sounds interesting

We could also do the same for dsa and rsa...

karel-m:
Currently it is:
```C
int dh_export_radix(int radix,
                    void *out, unsigned long *outlen,
                    int type, dh_key *key);
int dh_import_radix(int radix,
                    void *in,    unsigned long inlen,
                    void *prime, unsigned long primelen,
                    void *base,  unsigned long baselen,
                    int type, dh_key *key);
```

where `radix` can be `2..64` or `256`

karel-m:
And also radix-ified `dh_make_key_ex`
```C
int dh_make_key_ex(prng_state *prng, int wprng, int radix,
                   void *prime, unsigned long primelen,
                   void *base,  unsigned long baselen,
                   dh_key *key);
```

karel-m:
Should I apply the new radix style also to:
```C
int rsa_import_radix(int radix, char *N, char *e, char *d, char *p, char *q, char *dP, char *dQ, char *qP, rsa_key *key);
int dsa_import_radix(int radix, char *p, char *q, char *g, char *x, char *y, dsa_key *key);
int dsa_make_key_ex(prng_state *prng, int wprng, int group_size, int modulus_size, dsa_key *key, char* p_hex, char* q_hex, char* g_hex);
```
? (all of these will be new in 1.18)

It will nearly double the number of arguments.

sjaeckel:
> Should I apply the new radix style also to:
> ```C
> int rsa_import_radix(int radix, char *N, char *e, char *d, char *p, char *q, char *dP, char *dQ, char *qP, rsa_key *key);
> int dsa_import_radix(int radix, char *p, char *q, char *g, char *x, char *y, dsa_key *key);
> int dsa_make_key_ex(prng_state *prng, int wprng, int group_size, int modulus_size, dsa_key *key, char* p_hex, char* q_hex, char* g_hex);
> ```
> ? (all of these will be new in 1.18)
> 
> It will nearly double the number of arguments.

I thought the same...

I also thought that 1 radix is pretty unflexible, you e.g. wouldn't be able to store dh prime and base as hex string and the public/private part as binary...

how about this?
```
typedef struct {
  void* p;
  unsigned long len;
  int radix;
} ltc_pk_part;
```

karel-m:
Well, this does not look nice:
```C
int rsa_import_radix(int radix, 
                     void *N,  unsigned long Nlen,
                     void *e,  unsigned long elen,
                     void *d,  unsigned long dlen,
                     void *p,  unsigned long plen,
                     void *q,  unsigned long qlen,
                     void *dP, unsigned long dPlen,
                     void *dQ, unsigned long dQlen,
                     void *qP, unsigned long qPlen,
                     rsa_key *key);
```
But the `struct` approach would make it more complicated for the caller.

sjaeckel:
> ```C
> int rsa_import_radix(int radix, 
>                      void *N,  unsigned long Nlen,
>                      void *e,  unsigned long elen,
>                      void *d,  unsigned long dlen,
>                      void *p,  unsigned long plen,
>                      void *q,  unsigned long qlen,
>                      void *dP, unsigned long dPlen,
>                      void *dQ, unsigned long dQlen,
>                      void *qP, unsigned long qPlen,
>                      rsa_key *key);
> ```

true, that's ugly... but I don't have a better idea than the struct approach

> But the struct approach would make it more complicated for the caller.

we could help by providing something like

```
#define PK_PART_HEX(s) &((ltc_pk_part){s, 0, 16})
#define PK_PART_DEC(s) &((ltc_pk_part){s, 0, 10})
```

karel-m:
@sjaeckel I am out of ideas here, feel free to rebase this PR on develop and commit the unfinished/upolished radix related stuff here

sjaeckel:
> I am out of ideas here

me too

> feel free to rebase this PR on develop and commit the unfinished/upolished radix related stuff here

I don't want to have unfinished stuff in a potential RC

does someone of you have a good idea how to solve https://github.com/libtom/libtomcrypt/pull/227#issuecomment-309674046 ff.? @lkinley @rofl0r @sebastianas

sjaeckel:
> I don't want to have unfinished stuff in a potential RC

btw. I think we should wait with the RC until this PR is ready

karel-m:
Does it make sense to extract everything **except**:
* dh_export_radix
* dh_import_radix
* dh_make_key_ex

into a separate PR so that it can be merged to develop?

sjaeckel:
> Does it make sense to extract everything **except**:
> * dh_export_radix
> * dh_import_radix
> * dh_make_key_ex
> 
> into a separate PR so that it can be merged to develop?

sounds like a good idea to me

sjaeckel:
I'm going to rebase this on top of #238 and add the necessary functions

karel-m:
> I'm going to rebase this on top of #238 and add the necessary functions

no problem, go ahead

sjaeckel:
done! @karel-m can you quickly review please?
I'll merge then to #238 and we do a proper review there

karel-m:
At the first sight it looks good. I have to go through it more carefully but it would be better to have it all in one PR with RSA/DSA changes.

PR #230, opened 2017-06-14
Fix number of Miller-Rabin rounds

sjaeckel:
FIPS 180-4 defines a specific number of rounds of MR for DSA primes, c.f. FIPS-186-4 - Appendix C.3 - table C.1.

gmp also has an estimation which number is appropriate, c.f. doc of `mpz_probab_prime_p()` [here](https://gmplib.org/manual/Number-Theoretic-Functions.html)

This PR implements these recommendations.

### DIFF
--- a/src/headers/tomcrypt_math.h
+++ b/src/headers/tomcrypt_math.h
@@ -24,6 +24,12 @@
    typedef void rsa_key;
 #endif
 
+#ifndef LTC_MILLER_RABIN_REPS
+   /* Number of rounds of the Miller-Rabin test
+    * "Reasonable values of reps are between 15 and 50." c.f. gmp doc of mpz_probab_prime_p() */
+   #define LTC_MILLER_RABIN_REPS    35
+#endif
+
 /** math descriptor */
 typedef struct {
    /** Name of the math provider */
@@ -345,7 +351,7 @@ typedef struct {
 
    /** Primality testing
        @param a     The integer to test
-       @param b     The number of tests that shall be executed
+       @param b     The number of Miller-Rabin tests that shall be executed
        @param c     The destination of the result (FP_YES if prime)
        @return CRYPT_OK on success
    */
@@ -472,13 +478,13 @@ typedef struct {
    int (*submod)(void *a, void *b, void *c, void *d);
 
 /* ---- misc stuff ---- */
+
    /** Make a pseudo-random mpi
       @param  a     The mpi to make random
       @param  size  The desired length
       @return CRYPT_OK on success
    */
    int (*rand)(void *a, int size);
-
 } ltc_math_descriptor;
 
 extern ltc_math_descriptor ltc_mp;
--- a/src/math/gmp_desc.c
+++ b/src/math/gmp_desc.c
@@ -446,7 +446,7 @@ static int isprime(void *a, int b, int *c)
    LTC_ARGCHK(a != NULL);
    LTC_ARGCHK(c != NULL);
    if (b == 0) {
-       b = 8;
+       b = LTC_MILLER_RABIN_REPS;
    } /* if */
    *c = mpz_probab_prime_p(a, b) > 0 ? LTC_MP_YES : LTC_MP_NO;
    return CRYPT_OK;
--- a/src/math/ltm_desc.c
+++ b/src/math/ltm_desc.c
@@ -404,7 +404,7 @@ static int isprime(void *a, int b, int *c)
    LTC_ARGCHK(a != NULL);
    LTC_ARGCHK(c != NULL);
    if (b == 0) {
-       b = 8;
+       b = LTC_MILLER_RABIN_REPS;
    } /* if */
    err = mpi_to_ltc_error(mp_prime_is_prime(a, b, c));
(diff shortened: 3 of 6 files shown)


karel-m:
The first thing I want to point out: currently `mp_prime_is_prime(num,t,result)` does `t` M-R rounds with the first `t` primes (2,3,5,7..); however, it is considered better to uses `t` randomly chosen bases between `2` and `num-2` (see FIPS 184-4 C.3.1). So it is not just about the number of rounds.

Next, the pure M-R test might not be the best we can have today (I mean combining M-R with Lucas & co.). Couple of years ago I wanted to improve primality tests in libtomcrypt. Here is the related discussion https://rt.cpan.org/Public/Bug/Display.html?id=89400 It is not an easy task and it has impact not only to libtomcrypt but also to libtommath.

Some abandoned/untested code:
* https://github.com/DCIT/perl-CryptX/blob/better-primes/src/ltm/bn_mp_prime_is_prime.c
* https://github.com/DCIT/perl-CryptX/blob/better-primes/src/ltm/bn_mp_prime_miller_rabin_random.c

To sum up, I want to say that this rabbit hole is pretty deep.

sjaeckel:
> The first thing I want to point out: currently `mp_prime_is_prime(num,t,result)` does `t` M-R rounds with the first `t` primes (2,3,5,7..); however, it is considered better to uses `t` randomly chosen bases between `2` and `num-2` (see FIPS 184-4 C.3.1). So it is not just about the number of rounds.

Yeah, I saw the same thing but then I thought... well as long as nobody complains... but I also estimated how long it'll take to implement and then decided it's currently too long for me.

> Next, the pure M-R test might not be the best we can have today (I mean combining M-R with Lucas & co.). Couple of years ago I wanted to improve primality tests in libtomcrypt. Here is the related discussion https://rt.cpan.org/Public/Bug/Display.html?id=89400 It is not an easy task and it has impact not only to libtomcrypt but also to libtommath.

Especially the fact that it would have impact on ltm made me think, why not simply implement it via ltc_mp primitives in ltc? It's most likely a lot of effort, but less effort that adding it to ltm, tfm and gmp, as gmp has AFAICS also only MR for primality tests!?! I didn't look at the code, but the doc gave me the impression.

> Some abandoned/untested code:
> * https://github.com/DCIT/perl-CryptX/blob/better-primes/src/ltm/bn_mp_prime_is_prime.c
> * https://github.com/DCIT/perl-CryptX/blob/better-primes/src/ltm/bn_mp_prime_miller_rabin_random.c
> 
> To sum up, I want to say that this rabbit hole is pretty deep.

In the end I came multiple times back to the ltm fork of @czurnieden ... I think https://github.com/czurnieden/libtommath could be another good starting point

_Regarding this PR:_
I just wanted to implement the recommendations when I saw them and I think we should work on the primality testing, but for now it's out of scope. We should probably discuss the entire primality thing in another issue that this one.

I also thought about adding a function similar to ltm's `mp_prime_rabin_miller_trials()` with an API like
`int miller_rabin_rounds(void* mpi, int type)` where `type` would be one of `DSA_MR, DSA_MR_LUCAS` for now (or something like that...).

sjaeckel:
any ideas how we should go on with this?

karel-m:
The trouble is that FIPS-186-4 Appendix C.3 is talking about Miller-Rabin probabilistic primality tests using randomly chosen bases between 2 and n-2.

GMP's mpz_probab_prime_p is using Miller-Rabin with a deterministic sequence of bases, a[j] = j^2 + j + 41 using Euler's polynomial (doc says it is primality test found in Knuth's Seminumerical Algorithms book).

However LTC currently user Miller-Rabin with fixed t bases (first t primes) - I have only found this reference in HAC - Ch. 4 Public-Key Parameter (unfortunately it does not say much about how good or bad it is).

```
4.28  Remark (fixed bases in Miller-Rabin) If a1 and a2 are strong liars for n, their product
a1a2 is very likely, but not certain, to also be a strong liar for n. A strategy that is some-
times employed is to fix the bases a in the Miller-Rabin algorithm to be the first few primes
(composite bases are ignored because of the preceding statement), instead of choosing them
at random.
```

We are mixing rounds from FIPS (probabilistic tests using randomly chosen bases) and our rounds (deterministic, first t primes).

I am not able to say that increasing MR rounds (for our deterministic test with first t primes) from 8 to 35 is ok. Looking at libtommath's `ltm_prime_tab` 35 wouldn't work for `MP_8BIT`.

sjaeckel:
> We are mixing rounds from FIPS (probabilistic tests using randomly chosen bases) and our rounds (deterministic, first t primes).
> 
> I am not able to say that increasing MR rounds (for our deterministic test with first t primes) from 8 to 35 is ok.

I'm definitely sure that it's 'ok', but I'm not sure if it's enough.




> Looking at libtommath's ltm_prime_tab 35 wouldn't work for MP_8BIT

We shouldn't care about MP_8BIT, it's only there "for educational reasons".

karel-m:
> I'm definitely sure that it's 'ok',

Well, so do merge. It is definitely not making things worse.

Issue #228, opened 2017-06-14
Pre-release cosmetics


karel-m:
- [x] 1/ We should remove the old URLs from all `src/*` files ... changing the first comment to ..., c.f. #233 

 - [x] 2/ We should perhaps also replace the old tags `/* $Source$ */ ...` with ..., c.f. #233 

 - [x] 3/ We might want to add a check for 1/ + 2/ to `helper.pl`, c.f. 2cd69fb54109bc48b674d9edf1e326cd57821ab3

 - [x] 4/ https://github.com/libtom/libtomcrypt/issues/228#issuecomment-308953394 - local header files for internal functions, c.f. #241 

 - [x] ~5/ https://github.com/libtom/libtomcrypt/issues/228#issuecomment-309139392 - libtomcrypt.symbols~, c.f. https://github.com/libtom/libtomcrypt/issues/228#issuecomment-310444316

 - [x] 6/ https://github.com/libtom/libtomcrypt/issues/228#issuecomment-309425999 - enforce naming convention `static type _funcname(..)`, c.f. #234 

 - [x] 7/ https://github.com/libtom/libtomcrypt/issues/228#issuecomment-309873479 - ecc_sign+verify_hash_raw API, c.f. #235 

- [x] 8/ https://github.com/libtom/libtomcrypt/issues/228#issuecomment-310313978 - rsa_import_radix, dsa_import_radix, dsa_make_key_ex, c.f. #238 

- [x] 9/ https://github.com/libtom/libtomcrypt/issues/228#issuecomment-311676384 - Fix `make doxygen` warnings #264

karel-m:
Ad 3/ now you can run `./helper.pl -o` (it is not part of `./helper.pl --check-all`)

sjaeckel:
> 1/ We should remove the old URLs from all `src/*` files - which means changing the first comment to:
> ...
> 2/ We should perhaps also replace the old tags `/* $Source$ */ ...` with:

Sure!

Can we use this issue as collection for similar issues? and open separate PR's with changes

karel-m:
> Can we use this issue as collection for similar issues? and open separate PR's with changes

Yes, good idea.

sjaeckel:
4/ We should use local header files for internal functions

karel-m:
> We should use local header files for internal functions

Wouldn't it be enough to simply use:
```C
int some_public_func_a(int a, int b);
int some_public_func_b(int a, int b);
#ifdef LTC_SOURCE
[...]

PR #235, opened 2017-06-21
ECC ecc_sign+verify_hash_raw > ecc_sign+verify_hash_rfc7518

karel-m:
related to #228 point 7/

### DIFF
--- a/src/pk/ecc/ecc_sign_hash.c
+++ b/src/pk/ecc/ecc_sign_hash.c
@@ -22,7 +22,7 @@ static int _ecc_sign_hash(const unsigned char *in,  unsigned long inlen,
 {
    ecc_key       pubkey;
    void          *r, *s, *e, *p;
-   int           err, max_iterations = 20;
+   int           err, max_iterations = PK_MAX_RETRIES;
    unsigned long pbits, pbytes, i, shift_right;
    unsigned char ch, buf[MAXBLOCKSIZE];
 
--- a/src/pk/ecc/ecc_sign_hash.c
+++ b/src/pk/ecc/ecc_sign_hash.c
@@ -96,7 +96,7 @@ static int _ecc_sign_hash(const unsigned char *in,  unsigned long inlen,
    } while (--max_iterations > 0);
 
    if (max_iterations == 0) {
-     goto errnokey;
+      goto errnokey;
    }
 
    if (sigformat == 1) {
--- a/src/pk/ecc/ecc_sign_hash.c
+++ b/src/pk/ecc/ecc_sign_hash.c
@@ -7,43 +7,29 @@
  * guarantee it works.
  */
 
-/* Implements ECC over Z/pZ for curve y^2 = x^3 - 3x + b
- *
- * All curves taken from NIST recommendation paper of July 1999
- * Available at http://csrc.nist.gov/cryptval/dss.htm
- */
 #include "tomcrypt.h"
+
+#ifdef LTC_MECC
 
 /**
   @file ecc_sign_hash.c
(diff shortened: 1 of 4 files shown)


src/pk/ecc/ecc_sign_hash.c
```diff
@@ -7,43 +7,29 @@
...
-  @param s        The "s" integer of the signature (caller must initialize with mp_init() first)
-  @param prng     An active PRNG state
-  @param wprng    The index of the PRNG desired
-  @param key      A private ECC key
-  @return CRYPT_OK if successful
-*/
-int ecc_sign_hash_raw(const unsigned char *in,  unsigned long inlen,
-                            void *r, void *s,
-                            prng_state *prng, int wprng, ecc_key *key)
+static int ecc_sign_hash_ex(const unsigned char *in,  unsigned long inlen,
```
sjaeckel: ~~`_ecc_sign_hash_ex`~~
even better `_ecc_sign_hash`

karel-m: fixed

src/pk/ecc/ecc_verify_hash.c
```diff
@@ -116,9 +126,6 @@ int ecc_verify_hash_raw(      void   *r, void   *s,
       if ((err = ltc_mp.ecc_ptmul(u1, mG, mG, m, 0)) != CRYPT_OK)                                       { goto error; }
       if ((err = ltc_mp.ecc_ptmul(u2, mQ, mQ, m, 0)) != CRYPT_OK)                                       { goto error; }
 
-      /* find the montgomery mp */
```
karel-m: It was mistake, I was removing some code related to curves with custom A and removed also this one.

src/pk/ecc/ecc_sign_hash.c
```diff
@@ -59,13 +45,30 @@ int ecc_sign_hash_raw(const unsigned char *in,  unsigned long inlen,
...
+      for (i=0, ch=0; i<pbytes; i++) {
+        buf[i] = ch;
+        ch = (in[i] << (8-shift_right));
+        buf[i] = buf[i] ^ (in[i] >> shift_right);
+      }
+      if ((err = mp_read_unsigned_bin(e, (unsigned char *)buf, pbytes)) != CRYPT_OK)  { goto errnokey; }
+   }
 
    /* make up a key and export the public copy */
    for (;;) {
```
karel-m: I have added `max_iterations`, I am just not sure if it is worth another new define.

src/pk/ecc/ecc_sign_hash.c
```diff
@@ -7,43 +7,29 @@
...
-                            prng_state *prng, int wprng, ecc_key *key)
+static int _ecc_sign_hash(const unsigned char *in,  unsigned long inlen,
+                                unsigned char *out, unsigned long *outlen,
+                                prng_state *prng, int wprng, ecc_key *key, int sigformat)
 {
    ecc_key       pubkey;
-   void          *e, *p;
-   int           err;
+   void          *r, *s, *e, *p;
+   int           err, max_iterations = 20;
```
sjaeckel: probably we could introduce `PK_MAX_RETRIES` or something like that which can be used in all cases where we have to do retry-loops (c.f. `DH_MAKE_KEY_MAX_ITERATIONS`)? Otherwise we should use a separate one here as well.

src/pk/ecc/ecc_sign_hash.c
sjaeckel: c.f. https://github.com/libtom/libtomcrypt/pull/235#discussion_r123222654

PR #236, opened 2017-06-21
ECC-step2: curves y^2 = x^3 + ax + b

karel-m:
While hacking on `ecc_sign+verify_hash_rfc7518` earlier today I realized that it would be possible to extract from my branch `miko-ecc-enhancements` the crucial part implementing generalized elliptic curves `y^2 = x^3 + ax + b` (current `develop` branch supports only `y^2 = x^3 - 3x + b`).

The API changes comprise of the following new functions:
```C
int  ecc_export_raw(unsigned char *out, unsigned long *outlen, int type, ecc_key *key);
int  ecc_import_raw(const unsigned char *in, unsigned long inlen, ecc_key *key, ltc_ecc_set_type *dp);

ltc_ecc_set_type* ecc_dp_find_by_oid(unsigned long *oid, unsigned long oidsize);
ltc_ecc_set_type* ecc_dp_find_by_name(char *curve_name);
ltc_ecc_set_type* ecc_dp_find_by_params(char *hex_prime, char *hex_A, char *hex_B, char *hex_order, char *hex_Gx, char *hex_Gy, unsigned long cofactor);
```

These are missing due to #187
* `ecc_export_full` - export EC key in OpenSSL compatible format
* `ecc_import_full` - import EC key in OpenSSL compatible format
* `ecc_import_pkcs8` - import EC key in PKCS#8 format


karel-m:
A note related to this PR:

> We need a ltm-1.0 package backported to trusty to fix these build failures.
I had to remove the ltm from sid in #217 as I couldn't update clang otherwise...

sjaeckel:
Another note related to this PR:

We will need `sqrtmod_prime()` in tfm and gmp before this can be merged.

src/headers/tomcrypt_pk.h
```diff
@@ -292,13 +302,19 @@ int  ecc_test(void);
...
+ltc_ecc_set_type* ecc_dp_find_by_params(char *hex_prime, char *hex_A, char *hex_B, char *hex_order, char *hex_Gx, char *hex_Gy, unsigned long cofactor);
+
 int  ecc_make_key(prng_state *prng, int wprng, int keysize, ecc_key *key);
 int  ecc_make_key_ex(prng_state *prng, int wprng, ecc_key *key, const ltc_ecc_set_type *dp);
 void ecc_free(ecc_key *key);
 
 int  ecc_export(unsigned char *out, unsigned long *outlen, int type, ecc_key *key);
 int  ecc_import(const unsigned char *in, unsigned long inlen, ecc_key *key);
 int  ecc_import_ex(const unsigned char *in, unsigned long inlen, ecc_key *key, const ltc_ecc_set_type *dp);
+int  ecc_export_raw(unsigned char *out, unsigned long *outlen, int type, ecc_key *key);
```
sjaeckel: especially since this has the same signature as `ecc_export()`, how about re-using the same pattern that introduced `PK_STD`?
```
#define PK_RAW 0x2000
#define PK_COMPRESSED 0x4000
```

src/headers/tomcrypt_pk.h
```diff
@@ -11,7 +11,8 @@
 
 enum {
    PK_PUBLIC=0,
-   PK_PRIVATE=1
+   PK_PRIVATE=1,
+   PK_PUBLIC_COMPRESSED=2 /* used only when exporting public ECC key */
```
sjaeckel: c.f. my comment to `ecc_export_raw()`

src/headers/tomcrypt_pk.h
```diff
@@ -332,23 +348,32 @@ int  ecc_verify_hash(const unsigned char *sig,  unsigned long siglen,
                      const unsigned char *hash, unsigned long hashlen,
                      int *stat, ecc_key *key);
 
+int  ecc_verify_key(ecc_key *key);
+
+#ifdef LTC_SOURCE
```
sjaeckel: :+1:

src/pk/ecc/ecc_dp_find_by_name.c
```diff
@@ -0,0 +1,30 @@
...
+
+#ifdef LTC_MECC
+
+ltc_ecc_set_type* ecc_dp_find_by_name(char *curve_name)
+{
+   int i;
+
+   for (i = 0; ltc_ecc_sets[i].size != 0; i++) {
+      if (ltc_ecc_sets[i].name != NULL && XSTRCMP(ltc_ecc_sets[i].name, curve_name) == 0) {
+         return (ltc_ecc_set_type*)&ltc_ecc_sets[i];
```
sjaeckel: why is this cast required?

src/pk/ecc/ecc_dp_find_by_oid.c
```diff
@@ -0,0 +1,31 @@
...
+#ifdef LTC_MECC
+
+ltc_ecc_set_type* ecc_dp_find_by_oid(unsigned long *oid, unsigned long oidsize)
+{
+   int i;
+
+   for (i = 0; ltc_ecc_sets[i].size != 0; i++) {
+      if ((oidsize == ltc_ecc_sets[i].oid.OIDlen) &&
+          (XMEM_NEQ(oid, ltc_ecc_sets[i].oid.OID, sizeof(unsigned long) * ltc_ecc_sets[i].oid.OIDlen) == 0)) {
+         return (ltc_ecc_set_type*)&ltc_ecc_sets[i];
```
sjaeckel: same

src/pk/ecc/ecc_dp_find_by_params.c
```diff
@@ -0,0 +1,60 @@
...
+
+   for (i = 0; ltc_ecc_sets[i].size != 0; i++) {
+      if (_hexstrcmp(ltc_ecc_sets[i].prime, hex_prime) == 0 &&
+          _hexstrcmp(ltc_ecc_sets[i].A,     hex_A)     == 0 &&
+          _hexstrcmp(ltc_ecc_sets[i].B,     hex_B)     == 0 &&
+          _hexstrcmp(ltc_ecc_sets[i].order, hex_order) == 0 &&
+          _hexstrcmp(ltc_ecc_sets[i].Gx,    hex_Gx)    == 0 &&
+          _hexstrcmp(ltc_ecc_sets[i].Gy,    hex_Gy)    == 0 &&
+          ltc_ecc_sets[i].cofactor == cofactor) {
+         return (ltc_ecc_set_type*)&ltc_ecc_sets[i];
```
sjaeckel: same

src/pk/ecc/ecc_make_key.c
```diff
@@ -97,12 +96,21 @@ int ecc_make_key_ex(prng_state *prng, int wprng, ecc_key *key, const ltc_ecc_set
...
+    * the generated private key k should be the range [1, order-1]
+    *  a/ N = bitlen(order)
+    *  b/ generate N random bits and convert them into big integer k
+    *  c/ if k not in [1, order-1] go to b/
+    *  e/ Q = k*G
+    */
+   orderbits = mp_count_bits(order);
+   do {
+     if ((err = rand_bn_bits(key->k, orderbits, prng, wprng)) != CRYPT_OK)                      { goto errkey; }
+   } while (mp_iszero(key->k) || mp_cmp(key->k, order) != LTC_MP_LT);
```
sjaeckel: looks like we'll loop here forever as well and that the standard says to do so...

src/pk/ecc/ecc_dp_find_by_name.c
karel-m: If you mean `ecc_dp_find_by_name` it is for convenience like:
```C
dp = ecc_dp_find_by_name("SECP160R2");
err = ecc_make_key_ex(prng, wprng, key, dp);
```

src/pk/ecc/ecc_dp_find_by_oid.c
karel-m: this can be moved to ECC part2, where it is required as some DER formats do not contain all curve parameters but only OID of the curve

src/pk/ecc/ecc_dp_find_by_params.c
karel-m: I'll will also move this to ECC part 2 (this one might end up removed completely)

src/pk/ecc/ecc_make_key.c
karel-m: OK, I'll apply PK_MAX_RETRIES here

src/headers/tomcrypt_pk.h
karel-m: With EC things are a bit tricky as we have couple more scenarios:
* there is just one way of exporting RAW private key 
* RAW public key can be exported in full/long form
* of compressed/short form
we have our old DER format (basically compatible just with LTC) where the curve is identified just by its size
* later when dealing with OpenSSL compatible DER format we can export the key + curve OID
* or the key + all curve params

I am not yet sure what would be the best approach here. But yes we can handle all of them in one single `ecc_(import|export)_key`.

karel-m: On the other hand in ECC part 2 we call `ecc_(import|export)_raw` during importing/exporting OpenSSL keys as `ecc_(import|export)_raw` is kind of lower level action.

src/pk/ecc/ecc_dp_find_by_name.c
sjaeckel: what I meant was "why is _this_ `return `__`(ltc_ecc_set_type*)`__`&ltc_ecc_sets[i];` cast required?"

and I just got the answer... because the API signature is lying and you're returning a const object as non-const

```
src/pk/ecc/ecc_dp_find_by_name.c: In function ‘ecc_dp_find_by_name’:
src/pk/ecc/ecc_dp_find_by_name.c:20:17: warning: return discards ‘const’ qualifier from pointer target type [-Wdiscarded-qualifiers]
          return &ltc_ecc_sets[i];
```
three possibilities to solve this in order of preference:
1. make the return value a `const ltc_ecc_set_type*`
1. make a copy of the object you want to return before returning it as non-const (which the caller then has to free)
1. make `ltc_ecc_sets` not const

karel-m: This stuff gets even more complicated in ECC part 2 where we are importing keys in OpenSSL compatible DER format:
* one scenario is that the imported key has curve specified by OID - in this case we use `ecc_dp_find_by_oid(..)` which finds corresponding `ltc_ecc_sets[i]` item (const)
* the other scenario is that the imported key has explicitly defined curve parametes - here we can try whether these curve params do exist in our `ltc_ecc_sets` - this is the idea behind `ecc_dp_find_by_params(..)` - otherwise we have to allocate and properly fill `ltc_ecc_set_type` structure and free it during the key destruction in `ecc_free` (here it is tricky as `ecc_free` has no idea whether the `ltc_ecc_set_type` structure was dynamically allocated or not)

sjaeckel: and suddenly possibility 2 seems to become the preferred one, simply always allocate a copy so we can always free

karel-m: Yes, and it should be perhaps better named `ecc_dp_new_by_name`

karel-m: I have created some draft related to "possibility 2" 
* https://github.com/libtom/libtomcrypt/blob/pr/ecc-non-asn1-part/src/pk/ecc/ecc_dp_utils1.c

But it turned into a question: Why do we make copies of hexadecimal strings? It would make IMO more sense to convert those hex string into bignums during making a "copy":
* https://github.com/libtom/libtomcrypt/blob/pr/ecc-non-asn1-part/src/pk/ecc/ecc_dp_utils2.c

sjaeckel:
> A note (#65 (comment)) related to this PR:
> 
>> We need a ltm-1.0 package backported to trusty to fix these build failures.
>> I had to remove the ltm from sid in #217 as I couldn't update clang otherwise...

done

karel-m:
Some thoughts about applying similar API ideas as we did in RSA, DSA, DH.

This is what we have in 1.18-rc1 (and probably want to keep it for backwards compatibility):
```C
void ecc_sizes(int *low, int *high);
int ecc_get_size(ecc_key *key);
int ecc_make_key(prng_state *prng, int wprng, int keysize, ecc_key *key);
int ecc_make_key_ex(prng_state *prng, int wprng, ecc_key *key, const ltc_ecc_set_type *dp);
void ecc_free(ecc_key *key);
int ecc_export(unsigned char *out, unsigned long *outlen, int type, ecc_key *key);
int ecc_import(const unsigned char *in, unsigned long inlen, ecc_key *key);
int ecc_import_ex(const unsigned char *in, unsigned long inlen, ecc_key *key, const ltc_ecc_set_type *dp);
int ecc_ansi_x963_export(ecc_key *key, unsigned char *out, unsigned long *outlen);
int ecc_ansi_x963_import(const unsigned char *in, unsigned long inlen, ecc_key *key);
int ecc_ansi_x963_import_ex(const unsigned char *in, unsigned long inlen, ecc_key *key, ltc_ecc_set_type *dp);
int ecc_shared_secret(ecc_key *private_key, ecc_key *public_key, unsigned char *out, unsigned long *outlen);
int ecc_encrypt_key(const unsigned char *in, unsigned long inlen, unsigned char *out, unsigned long *outlen, prng_state *prng, int wprng, int hash, ecc_key *key);
int ecc_decrypt_key(const unsigned char *in, unsigned long inlen, unsigned char *out, unsigned long *outlen, ecc_key *key);
int ecc_sign_hash_rfc7518(const unsigned char *in, unsigned long inlen, unsigned char *out, unsigned long *outlen, prng_state *prng, int wprng, ecc_key *key);
int ecc_sign_hash(const unsigned char *in, unsigned long inlen, unsigned char *out, unsigned long *outlen, prng_state *prng, int wprng, ecc_key *key);
int ecc_verify_hash_rfc7518(const unsigned char *sig, unsigned long siglen, const unsigned char *hash, unsigned long hashlen, int *stat, ecc_key *key);
int ecc_verify_hash(const unsigned char *sig, unsigned long siglen, const unsigned char *hash, unsigned long hashlen, int *stat, ecc_key *key);
```

I propose the following new additions in the "next":
```C
int ecc_set_dp(const ltc_ecc_set_type *set, ecc_key *key);
int ecc_set_dp_name(const char *curve_name, ecc_key *key);
int ecc_set_dp_oid(const unsigned long *oid, unsigned long oidsize, ecc_key *key);
int ecc_set_dp_size(int size, ecc_key *key);
int ecc_set_dp_copy(const ecc_key *srckey, ecc_key *key);
int ecc_generate_key(prng_state *prng, int wprng, ecc_key *key);
int ecc_set_key(const unsigned char *in, unsigned long inlen, int type, ecc_key *key); /* type can be left out */
int ecc_export_key(unsigned char *out, unsigned long *outlen, int type, ecc_key *key);
```

karel-m:
This PR was re-factored as proposed here: https://github.com/libtom/libtomcrypt/pull/236#issuecomment-314246411

Currently it is ready only for API review (I did my best to keep the API as much backwards compatible as possible).

Definitely needs at least one more pass to check/add more LTC_ARGCHK's + check whether we correctly free what has to be freed etc.

karel-m:
Just for record - the idea:
```C
int ecc_get_set(const char *curve_name, const ltc_ecc_set_type** dp);
/* or */
const ltc_ecc_set_type* ecc_get_set(const char *curve_name);
```

karel-m:
UPDATE: the latest version of this PR adds the following to the public API
```C
int  ecc_get_set_by_name(const char* name, const ltc_ecc_set_type** dp);
int  ecc_set_dp(const ltc_ecc_set_type *set, ecc_key *key);
int  ecc_generate_key(prng_state *prng, int wprng, ecc_key *key);
int  ecc_set_key(const unsigned char *in, unsigned long inlen, int type, ecc_key *key);
int  ecc_export_key(unsigned char *out, unsigned long *outlen, int type, ecc_key *key);
```

karel-m:
sorry, already reopened

karel-m:
@sjaeckel could you please have a look at API additions proposed in https://github.com/libtom/libtomcrypt/pull/236#issuecomment-331476974 ? Just the API, no need to go through the implementation at the moment.

src/headers/tomcrypt_pk.h
```diff
@@ -284,18 +298,35 @@ typedef struct {
...
+   /** The fields B param */
+   void *B;
+   /** The order of the curve */
+   void *order;
+   /** The base point G on the curve */
+   ecc_point base;
+   /** The co-factor */
+   unsigned long cofactor;
+   /** The OID structure */
+   oid_st oid;
```
sjaeckel: this will prevent compilation of a user-space app without defining `LTC_SOURCE`

src/headers/tomcrypt_pk.h
```diff
@@ -270,6 +277,13 @@ typedef struct {
...
    /** The y co-ordinate of the base point on the curve (hex) */
    const char *Gy;
+
+   /** The co-factor */
+   unsigned long cofactor;
+
+   /** The OID */
+   unsigned long oid[16];
+   unsigned long oidlen;
 } ltc_ecc_set_type;
```
sjaeckel: is `ltc_ecc_set_type` the right name for this struct? shouldn't it better be called `ltc_ecc_curve`

karel-m: Yes, it would. However, I have not introduced this struct (only added A, cofactor, OID) it already exists and is part of the API. So for backwards compatibility we have to stay with `ltc_ecc_set_type`.

sjaeckel: we're going to change the API's so why shouldn't we fix the names in the same step?

src/headers/tomcrypt_pk.h
karel-m: The newly introduced `ltc_ecc_dp` was intended to be "invisible" from public API.  User-space apps should neither see nor manipulate this struct directly but only via `ecc_set_dp(..)`.

But I have to check whether this intention really happened.

sjaeckel: this won't be possible as `ecc_key` contains an instance of `ltc_ecc_dp`

src/headers/tomcrypt_pk.h
karel-m: I wanted to keep the changes as small as possible. In fact we are "only" adding to the public API:
* 3 new items into `ltc_ecc_set_type` struct
* 5 completely new functions

src/headers/tomcrypt_pk.h
karel-m: You are right. So either moving `oid_st` to public API or switching to `unsigned long oid[16]; unsigned long oidlen;`.

src/headers/tomcrypt_pk.h
sjaeckel: I'd say with 1.18 "keeping changes small" is finished. We're not going to release a 1.18.1 with these ECC changes, even if the changes don't look that huge we're going to have to bump the ABI again... in the light of the late ABI bump for 1.18 we could (should!?) have included this in 1.18 :/ sorry for that

karel-m: No need to be sorry. Ad ABI versioning I think we can keep it independent on lib versioning and simply bump it when necessary.

src/headers/tomcrypt_pk.h
```diff
@@ -254,9 +254,6 @@ int dh_check_pubkey(dh_key *key);
 
 /** Structure defines a GF(p) curve */
 typedef struct {
-   /** The size of the curve in octets */
-   int size;
```
J08nY: @karel-m What is the reason for removing this?

sjaeckel: To me it looks like it's not needed as a user will normally not identify the curve by its size but by its name.
Also the size got ambiguous now that there are multiple standard-curves with the same size.

But that's just a wild guess :)
@karel-m ?

buggywhip:
Sent fm iTouch

> On Nov 30, 2017, at 12:48 AM, Steffen Jaeckel <notifications@github.com> wrote:
> 
> To me it looks like it's not needed as a user will normally not identify the curve by its size but by its name.
> 
Definitely agree!

src/headers/tomcrypt_pk.h
J08nY: Hmm, I think that keeping the size or rather, changing the size to refer to the curve size in bits would be better. Many crypto libraries allow one to just specify a requested curve size in bits and requested field type and then provide a default named curve if one is present. (For example Java Cryptographic Architecture requires that)

Removing the size or storing the byte size can also pose problems for curves of bit size `% 8 != 0`.

sjaeckel: There's still `ecc_make_key()` which takes the size of the desired key and then chooses one for you :)

src/headers/tomcrypt_pk.h
```diff
@@ -309,6 +338,12 @@ int  ecc_test(void);
 void ecc_sizes(int *low, int *high);
 int  ecc_get_size(ecc_key *key);
 
+int  ecc_get_set_by_name(const char* name, const ltc_ecc_set_type** dp);
+int  ecc_set_dp(const ltc_ecc_set_type *set, ecc_key *key);
```
sjaeckel: that one's redundant in the public API with `ecc_make_key_ex()`

src/math/tfm_desc.c
```diff
@@ -741,6 +776,7 @@ const ltc_math_descriptor tfm_desc = {
    &mul,
    &muli,
    &sqr,
+   &sqrtmod_prime,
```
sjaeckel: why can't we just set this to NULL and check before the invocation?
Just an idea...
AFAIU there's no functional difference of compressed and uncompressed points, right?
so we could disable compression at runtime in case tfm is used... (in fact that would be the same behavior at runtime than what is currently done in `ecc_encrypt_key()` at compile-time)

src/headers/tomcrypt_pk.h
sjaeckel: ah, or is there a use-case where you want to do the following?

```
ecc_get_set_by_name()
ecc_set_dp()
ecc_set_key(..., PK_PUBLIC, ...)
```

src/headers/tomcrypt_pk.h
J08nY: `ecc_make_key()` currently works by looking at the `size` attribute of curves.

This PR removes that attribute and does the lookup at [ecc_set_dp_size](https://github.com/libtom/libtomcrypt/pull/236/files#diff-72b3a87b26a1dfde185685298ed58b33R51) by hardcoding the values.

src/headers/tomcrypt_pk.h
J08nY: Well, for example when manually importing the key from other/sources formats. This should provide the convenience of having function calls instead of manipulating the ecc_key struct directly.

src/pk/ecc/ecc_set_dp_internal.c
```diff
@@ -0,0 +1,105 @@
...
+ *
+ * The library is free for all purposes without any express
+ * guarantee it works.
+ */
+
+#include "tomcrypt.h"
+
+#ifdef LTC_MECC
+
+int ecc_set_dp_oid(unsigned long *oid, unsigned long oidsize, ecc_key *key)
```
sjaeckel: `ecc_set_XX_by_oid`

src/pk/ecc/ecc_set_dp_internal.c
```diff
@@ -0,0 +1,105 @@
...
+      if ((oidsize == ltc_ecc_sets[i].oidlen) &&
+          (XMEM_NEQ(oid, ltc_ecc_sets[i].oid, sizeof(unsigned long) * ltc_ecc_sets[i].oidlen) == 0)) {
+         break;
+      }
+   }
+   if (ltc_ecc_sets[i].name == NULL) return CRYPT_ERROR; /* not found */
+   return ecc_set_dp(&ltc_ecc_sets[i], key);
+}
+
+int ecc_set_dp_copy(ecc_key *srckey, ecc_key *key)
```
sjaeckel: `ecc_copy_XX`

src/pk/ecc/ecc_set_dp_internal.c
```diff
@@ -0,0 +1,105 @@
...
+   for (i = 0; i < key->dp.oidlen; i++) key->dp.oid[i] = srckey->dp.oid[i];
+   /* success */
+   return CRYPT_OK;
+
+error:
+   ecc_free(key);
+   return err;
+}
+
+int ecc_set_dp_bn(void *a, void *b, void *prime, void *order, void *gx, void *gy, unsigned long cofactor, ecc_key *key)
```
sjaeckel: `ecc_set_XX_from_mpis`

src/pk/ecc/ecc_test.c
```diff
@@ -21,68 +16,12 @@
...
-   err = CRYPT_OK;
-done:
-   ltc_ecc_del_point(GG);
-   ltc_ecc_del_point(G);
-   mp_clear_multi(order, modulus, NULL);
-   return err;
+   /* the main ECC tests are in tests/ecc_test.c
+    * this function is kept just for API compatibility
+    */
+   return CRYPT_OK;
```
sjaeckel: at least return `CRYPT_NOP` :)

src/pk/ecc/ltc_ecc_is_point_at_infinity.c
```diff
@@ -0,0 +1,52 @@
...
+
+   /* compute y^2 */
+   if ((err = mp_mulmod(P->y, P->y, modulus, y2)) != CRYPT_OK)   goto cleanup;
+
+   /* compute x^3 */
+   if ((err = mp_mulmod(P->x, P->x, modulus, x3)) != CRYPT_OK)   goto cleanup;
+   if ((err = mp_mulmod(P->x, x3, modulus, x3))   != CRYPT_OK)   goto cleanup;
+
+   /* test y^2 == x^3 */
+   if ((mp_cmp(x3, y2) == LTC_MP_EQ) && !mp_iszero(y2)) retval = 1;
```
sjaeckel: IIUC this is a critical test, so I don't think the logic that goes "if everything went fine we should trigger the error" applies here, but we should go for "if something fails we should trigger the error".

so we should set `retval = 1` before the `mp_init_multi` and set it to zero in inverted logic here.

most likely even better would be to return `err` and add another `int*`-argument for the current `retval`.

src/pk/ecc/ltc_ecc_projective_add_point.c
```diff
@@ -45,14 +41,36 @@ int ltc_ecc_projective_add_point(ecc_point *P, ecc_point *Q, ecc_point *R, void
...
+   if ((mp_cmp(P->x, Q->x) == LTC_MP_EQ) && (mp_cmp(P->z, Q->z) == LTC_MP_EQ)) {
+      if (mp_cmp(P->y, Q->y) == LTC_MP_EQ) {
+         /* here P = Q >> Result = 2 * P (use doubling) */
+         mp_clear_multi(t1, t2, x, y, z, NULL);
+         return ltc_ecc_projective_dbl_point(P, R, ma, modulus, mp);
+      }
+      if ((err = mp_sub(modulus, Q->y, t1)) != CRYPT_OK)                       { goto done; }
+      if (mp_cmp(P->y, t1) == LTC_MP_EQ) {
+         /* here Q = -P >>> Result = the point at infinity */
+         if ((err = ltc_mp.set_int(R->x, 1)) != CRYPT_OK)                      { goto done; }
```
sjaeckel: I've seen this code now for at least the 3rd time, can you please add a `ltc_ecc_set_to_point_at_infinity()`

src/pk/ecc/ecc_set_dp_internal.c
sjaeckel: `src` can also be `const`

src/headers/tomcrypt_pk.h
sjaeckel: thou shalt never manipulate the `ecc_key` struct directly!

src/headers/tomcrypt_pk.h
sjaeckel: > This PR removes that attribute and does the lookup at ecc_set_dp_size by hardcoding the values.

yeah, it's truly not elegant at all... :)

I think it'd make sense to add a `ecc_get_dp_by_bits` (or sth like that) which finally uses bits instead of octets, then iterate there over the `ltc_ecc_sets` array, and compare the `mp_count_bits(key->dp.prime)`. `ecc_make_key()` would then call that one with `keysize*8`

karel-m: The `int size` is simply redundant (it is the byte size of `prime`). It used to be a unique identifier of elliptic curve but it is not true any more. For some sizes we several curves so it does not make sense to ask for a curve just by giving a size (no matter whether byte size or bit size). You should use curve name or curve OID. 

The `ecc_set_dp_size` was implemented just for compatibility reasons. It is not nice but it is a backward compatible mapping between old and new API.

src/headers/tomcrypt_pk.h
karel-m: The sequence: `ecc_get_set_by_name` + `ecc_set_dp` + `ecc_set_key` is IMO pretty handy. Consider `ecc_set_key` as a kind of "import key" function that is capable of loading raw private/public keys (e.g. bitcoin ECDSA keys uses this raw format, the curve is fixed/well-known "secp256k1").

karel-m:
> this feels like a too big PR to review properly

Yes it is big but the nature of this change is simply quite complex.

I suggest firstly focus on reviewing newly added functions (like proper naming, calling conventions ...) + changes to the publicly "visible" structs.

After that I can write the updated documentation (which can reveal couple of API gotchas).

And in the end we can polish implementation details.

sjaeckel:
> Yes it is big but the nature of this change is simply quite complex.
> 
> I suggest firstly focus on reviewing newly added functions (like proper naming, calling conventions ...) + changes to the publicly "visible" structs.
> 
> After that I can write the updated documentation (which can reveal couple of API gotchas).
> 
> And in the end we can polish implementation details.

sounds like a plan, we should probably now first finish #309 , I'll add documentation soon, but I think from a coding PoV it's fine. Then we go on with this one.

src/headers/tomcrypt_pk.h
karel-m: done (renamed to ltc_ecc_curve)

src/math/tfm_desc.c
karel-m: done

src/pk/ecc/ecc_set_dp.c
```diff
@@ -0,0 +1,97 @@
...
+   for (i = 0; i < key->dp.oidlen; i++) key->dp.oid[i] = set->oid[i];
+   /* success */
+   return CRYPT_OK;
+
+error:
+   ecc_free(key);
+   return err;
+}
+
+int ecc_set_dp_size(int size, ecc_key *key)
```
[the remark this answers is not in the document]
karel-m: done (ecc_set_dp_by_size)

src/pk/ecc/ecc_get_set.c
```diff
@@ -0,0 +1,40 @@
...
+ *
+ * The library is free for all purposes without any express
+ * guarantee it works.
+ */
+
+#include "tomcrypt.h"
+
+#ifdef LTC_MECC
+
+int ecc_get_set_by_name(const char* name, const ltc_ecc_set_type** dp)
```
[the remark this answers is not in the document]
karel-m: renamed to `ecc_get_curve_by_name` here some background when I use `dp` and when `curve`
* a `curve` is always a pointer to `ltc_ecc_curve` struct containing curve definition as **hex strings**
* domain parameters `dp` is part of the key, I mean `ecc_key` which looks like:
```
typedef struct {
    int type;         /* PK_PRIVATE or PK_PUBLIC */
    ltc_ecc_dp dp;    /* domain parameters */
    ecc_point pubkey; /* public key */
    void *k;          /* private key - MPI number */
}
```
* so `dp` is a structure containing also curve parameters but not as hexadecimal strings, but loaded as **mpi numbers**

So `ecc_get_curve_by_..` means: find `ltc_ecc_curve` struct which fits given criteria. On the other hand `ecc_set_dp_..` means setup `dp` part of `ecc_key`

src/pk/ecc/ecc_set_dp_internal.c
karel-m: done

src/pk/ecc/ecc_set_dp_internal.c
karel-m: done

src/pk/ecc/ecc_test.c
karel-m: done

src/pk/ecc/ltc_ecc_is_point_at_infinity.c
karel-m: done ( return err and add another int*-argument)

src/pk/ecc/ecc_set_dp_internal.c
```diff
@@ -0,0 +1,150 @@
...
+{
+   int err;
+   unsigned i;
+   void *tmp;
+   const ltc_ecc_curve *curve;
+
+   key->dp.oidlen = 0;
+   if ((err = mp_init(&tmp)) != CRYPT_OK) return;
+   for (curve = ltc_ecc_curves; curve->name != NULL; curve++) {
+      if ((err = mp_read_radix(tmp, curve->prime, 16)) != CRYPT_OK) continue;
```
sjaeckel: _copied from https://github.com/libtom/libtomcrypt/pull/236#discussion_r176715691_

how about converting the `key` elements once to strings and `strcmp()` then the elements of the sets?
alternatively add a function like
```
static int _ecc_cmp_hex_bn(const char* h, void* b, void* tmp)
{
   if (mp_read_radix(tmp, h, 16) != CRYPT_OK) return CRYPT_ERROR;
   if (mp_cmp(tmp, b) != LTC_MP_EQ) return CRYPT_ERROR;
   return CRYPT_OK;
}
```

src/pk/ecc/ltc_ecc_projective_add_point.c
sjaeckel: as there's also a case above that does similar stuff how about adding 
```
#define _LTC_ECC_POINT_ZERO 0, 0, 1
#define _LTC_ECC_POINT_INF 1, 1, 0
int ltc_ecc_set_coordinates(ltc_mp_digit x, ltc_mp_digit y, ltc_mp_digit z, ecc_point *P);
```

src/pk/ecc/ecc_set_dp_internal.c
karel-m: Well, I did basically the opposite thing, when loading the key I do convert all curve params (hex strings) into mpis. It saved couple of hex2mpi operations (grep for `mp_read_radix` in the old ecc implementation).

The function `_ecc_oid_lookup` is not heavily used, therefore I think an extra cost is acceptable here.

src/pk/ecc/ltc_ecc_projective_add_point.c
karel-m: The `ltc_ecc_set_to_point_at_infinity` looks reasonable, I do not see many uses for more general `ltc_ecc_set_coordinates`. But it may be worth considering `ltc_ecc_copy_point`.

src/pk/ecc/ecc_get_set.c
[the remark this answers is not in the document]
karel-m: TODO: I agree, a support for alternative names would be nice

karel-m:
Ad alternative names - what about to change the curve struct like this:
```C
typedef struct {
   const char *names[6]; /* HERE IS THE CHANGE!!! */
   const char *prime;
   const char *A;
   const char *B;
   const char *order;
   const char *Gx;
   const char *Gy;
   unsigned long cofactor;
   unsigned long oid[16];
   unsigned long oidlen;
} ltc_ecc_curve;
```
and later use this:
```C
{
  /* curve name */ { "SECP192R1", "NISTP192", "PRIME192V1", NULL, NULL, NULL },
  /* prime      */ "FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFEFFFFFFFFFFFFFFFF",
  /* A          */ "FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFEFFFFFFFFFFFFFFFC",
  /* B          */ "64210519E59C80E70FA7E9AB72243049FEB8DEECC146B9B1",
  /* order      */ "FFFFFFFFFFFFFFFFFFFFFFFF99DEF836146BC9B1B4D22831",
  /* Gx         */ "188DA80EB03090F67CBF20EB43A18800F4FF0AFD82FF1012",
  /* Gy         */ "07192B95FFC8DA78631011ED6B24CDD573F977A11E794811",
  /* cofactor   */ 1,
  /* OID        */ { 1,2,840,10045,3,1,1 }, 7
}
```

karel-m:
BTW this PR has broken LTC_MECC_FP support 😞

src/pk/ecc/ecc_free.c
```diff
@@ -28,7 +23,14 @@
...
    LTC_ARGCHKVD(key != NULL);
-   mp_clear_multi(key->pubkey.x, key->pubkey.y, key->pubkey.z, key->k, NULL);
+   /* clean dp */
+   mp_cleanup_multi(&key->dp.prime, &key->dp.order,
+                    &key->dp.A, &key->dp.B,
+                    &key->dp.base.x, &key->dp.base.y, &key->dp.base.z,
+                    NULL);
+
+   /* clean key */
+   mp_cleanup_multi(&key->pubkey.x, &key->pubkey.y, &key->pubkey.z, &key->k, NULL);
```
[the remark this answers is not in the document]
karel-m: done (cleanup in one call)

src/pk/ecc/ecc_make_key.c
```diff
@@ -31,89 +26,76 @@
...
+   }
+
+   /* success */
    err = CRYPT_OK;
    goto cleanup;
-errkey:
-   mp_clear_multi(key->pubkey.x, key->pubkey.y, key->pubkey.z, key->k, NULL);
+
+error:
+   ecc_free(key);
```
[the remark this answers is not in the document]
karel-m: You are right but this pattern already exist e.g. in `dsa_set_key`

sjaeckel:
> BTW this PR has broken LTC_MECC_FP support 

unfixable?

src/headers/tomcrypt_pk.h
```diff
@@ -250,17 +254,17 @@ int dh_check_pubkey(dh_key *key);
...
 
-/** Structure defines a NIST GF(p) curve */
+/** Structure defines a GF(p) curve */
 typedef struct {
-   /** The size of the curve in octets */
-   int size;
-
    /** name of curve */
-   const char *name;
+   const char *names[6];
```
sjaeckel: I'm not 100% sure if I really like this

src/pk/ecc/ecc_get_set.c
```diff
@@ -0,0 +1,63 @@
...
+#include "tomcrypt.h"
+
+#ifdef LTC_MECC
+
+/* case-insensitive match + ignore '-', '_', ' ' */
+static int _name_match(const char *left, const char *right)
+{
+   char lc_r, lc_l;
+
+   while ((*left != 0) && (*right != 0)) {
```
sjaeckel: please use `'\0'` to check for the terminating NUL byte

karel-m:
> unfixable?

we are missing `order` in `ltc_ecc_fp_mul2add` params as "standard" mul2add does not need it

sjaeckel:
> we are missing order in ltc_ecc_fp_mul2add params as "standard" mul2add does not need it

how about adding it as an argument to `ltc_ecc_fp_add_point()` and call that one (from one of our API's) before any of `fp_mul2add()` or `fp_mulmod()` are called?

_BUT_ I've no clue how widely `LTC_MECC_FP` is even used and if we should bother investing a lot of time...

src/pk/ecc/ecc_set_dp_internal.c
sjaeckel: Okay, perfectly fine. But please still add `_ecc_cmp_hex_bn()`

src/pk/ecc/ecc_set_dp.c
```diff
@@ -37,9 +37,32 @@ int ecc_set_dp(const ltc_ecc_curve *curve, ecc_key *key)
    /* cofactor & size */
    key->dp.cofactor = curve->cofactor;
    key->dp.size = mp_unsigned_bin_size(key->dp.prime);
-   /* OID */
-   key->dp.oidlen = curve->oidlen;
-   for (i = 0; i < key->dp.oidlen; i++) key->dp.oid[i] = curve->oid[i];
+   /* OID string >> unsigned long oid[16] + oidlen */
+   if (curve->OID != NULL) {
+      for (i = 0, j = 0; i < strlen(curve->OID); i++) {
+         if (curve->OID[i] == '.') {
```
sjaeckel: how about adding `str_get_oid(const char*, oid_st*)` similar to `pk_get_oid()`?

src/pk/ecc/ecc_get_curve_by_name.c
```diff
@@ -38,15 +217,27 @@ static int _name_match(const char *left, const char *right)
 int ecc_get_curve_by_name(const char *name, const ltc_ecc_curve **cu)
```
sjaeckel: then we also need a `ecc_curve_get_name()` which simply returns the first element of the names

src/pk/ecc/ecc_get_curve_by_name.c
```diff
@@ -11,6 +11,185 @@
 
 #ifdef LTC_MECC
 
+const struct {
+   const char *OID;
+   const char *names[6];
+} _curve_names[] = {
+#if defined(LTC_ECC_SECP112R1) || defined(LTC_ECC112)
+   {
+      "1.3.132.0.6", { "SECP112R1", "ECC-112", NULL }
```
sjaeckel: Can you please use the official spelling? (as proposed in https://github.com/libtom/libtomcrypt/pull/236#discussion_r176721275)

src/pk/ecc/ecc_set_dp.c
sjaeckel: ...which converts the input, not a LUT as the latter... :) and `str_to_oid(const char*, unsigned char*, unsigned long*)` would also be perfectly fine!

karel-m: done, see pk_oid_str_to_num + pk_oid_num_to_str

src/pk/ecc/ecc_get_curve_by_name.c
karel-m: I am not sure, I do not a use case for `ecc_curve_get_name`

src/pk/ecc/ecc_get_curve_by_name.c
karel-m: Do you mean:
```
secp256r1
NIST P-256
prime256v1
ECC-256
P-256
```
?

I also want to point out that apart from official names there is something like "industry standard" how these curves are named in other libraries - e.g. openssl:
```
$ openssl ecparam -list_curves | grep 'prime field'
  secp112r1 : SECG/WTLS curve over a 112 bit prime field
  secp112r2 : SECG curve over a 112 bit prime field
  secp128r1 : SECG curve over a 128 bit prime field
  secp128r2 : SECG curve over a 128 bit prime field
  secp160k1 : SECG curve over a 160 bit prime field
  secp160r1 : SECG curve over a 160 bit prime field
  secp160r2 : SECG/WTLS curve over a 160 bit prime field
  secp192k1 : SECG curve over a 192 bit prime field
  secp224k1 : SECG curve over a 224 bit prime field
  secp224r1 : NIST/SECG curve over a 224 bit prime field
  secp256k1 : SECG curve over a 256 bit prime field
  secp384r1 : NIST/SECG curve over a 384 bit prime field
  secp521r1 : NIST/SECG curve over a 521 bit prime field
  prime192v1: NIST/X9.62/SECG curve over a 192 bit prime field
  prime192v2: X9.62 curve over a 192 bit prime field
  prime192v3: X9.62 curve over a 192 bit prime field
  prime239v1: X9.62 curve over a 239 bit prime field
  prime239v2: X9.62 curve over a 239 bit prime field
  prime239v3: X9.62 curve over a 239 bit prime field
  prime256v1: X9.62/SECG curve over a 256 bit prime field
  wap-wsg-idm-ecid-wtls6: SECG/WTLS curve over a 112 bit prime field
  wap-wsg-idm-ecid-wtls7: SECG/WTLS curve over a 160 bit prime field
  wap-wsg-idm-ecid-wtls8: WTLS curve over a 112 bit prime field
  wap-wsg-idm-ecid-wtls9: WTLS curve over a 160 bit prime field
  wap-wsg-idm-ecid-wtls12: WTLS curvs over a 224 bit prime field
  brainpoolP160r1: RFC 5639 curve over a 160 bit prime field
  brainpoolP160t1: RFC 5639 curve over a 160 bit prime field
  brainpoolP192r1: RFC 5639 curve over a 192 bit prime field
  brainpoolP192t1: RFC 5639 curve over a 192 bit prime field
  brainpoolP224r1: RFC 5639 curve over a 224 bit prime field
  brainpoolP224t1: RFC 5639 curve over a 224 bit prime field
  brainpoolP256r1: RFC 5639 curve over a 256 bit prime field
  brainpoolP256t1: RFC 5639 curve over a 256 bit prime field
  brainpoolP320r1: RFC 5639 curve over a 320 bit prime field
  brainpoolP320t1: RFC 5639 curve over a 320 bit prime field
  brainpoolP384r1: RFC 5639 curve over a 384 bit prime field
  brainpoolP384t1: RFC 5639 curve over a 384 bit prime field
  brainpoolP512r1: RFC 5639 curve over a 512 bit prime field
  brainpoolP512t1: RFC 5639 curve over a 512 bit prime field
```
We should be IMO able to understand these "common" names.

src/misc/pk_oid_str.c
```diff
@@ -0,0 +1,73 @@
...
+      else {
+         return CRYPT_ERROR;
+      }
+   }
+   if (j == 0) return CRYPT_ERROR;
+   *oidlen = j + 1;
+   return CRYPT_OK;
+}
+
+int pk_oid_num_to_str(unsigned long *oid, unsigned long oidlen, char *OID, unsigned long *outlen)
```
sjaeckel: :+1: 

\*cough\* [der_test.c:313](https://github.com/libtom/libtomcrypt/blob/develop/tests/der_test.c#L313) \*cough\*

:)

src/pk/ecc/ecc_get_curve_by_name.c
sjaeckel: sounds good

sjaeckel: btw regarding the names

```
secp256r1 (secg name)
NIST P-256 (safecurves uses this name, but fips-186-4 calls it "Curve P-256" so should we add that as well?)
prime256v1 (is that the X9.62 name?)
ECC-256 (the old libtom name)
P-256 (where does this come from?)
```

this comment is only relevant if we add `ecc_curve_get_name()` ... I don't like the inconsistency in the openssl naming...

```
...
  secp521r1 : NIST/SECG curve over a 521 bit prime field
  prime192v1: NIST/X9.62/SECG curve over a 192 bit prime fie
...
  wap-wsg-idm-ecid-wtls6: SECG/WTLS curve over a 112 bit prime field
...
```

I'd propose to use the names if available in this order of the standards bodies: NIST-SECG-ANSI-BRAINPOOL-OTHERS

buggywhip:
We have  ecc_get_curve_by_name() but ought we also have a ecc_get_curve_by_oid()?

buggywhip:
Ought we have a mechanical process to confirm that a name only appears once in a list of all names that appear in all curves?

buggywhip:
I like the idea of supporting synonymous names for the same curve but am bothered by including the potentially ambiguous names like P-256 and ECC-256, past history notwithstanding. I'd much prefer developers change over to the standard-based names like secp256r1 or prime256v1.  Failing that, the doc should be crystal clear that P-xxx and ECC-XXX are only supported for reasons of backward compatibility.

src/pk/ecc/ecc_get_curve_by_name.c
karel-m: Ad P-256 & co. - these names are also used in JSON Web Signature (JWS) related RFCs 7515, 7518. It is perhaps based on curve names from https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.186-4.pdf

karel-m:
ad `int ecc_get_curve_by_oid(const char* oid, const ltc_ecc_curve** cu);`

with current implementation you can call:

```C
err = ecc_get_curve_by_name("1.3.132.0.31", &curve);
```
I know that the name is a bit misleading but we can explain it in documentation.

src/misc/pk_oid_str.c
sjaeckel: lulz, no the one before was perfectly fine, I just thought you can use it there (please keep your version!)

karel-m: OK, I'll put it back than :)

sjaeckel:
> I know that the name is a bit misleading but we can explain it in documentation.

how about renaming it to `ecc_get_curve(const char *id, const ltc_ecc_curve **cu)`?

src/misc/pk_oid_str.c
```diff
@@ -0,0 +1,79 @@
...
+            if (++k >= tmpsz) return CRYPT_ERROR;
+            j /= 10;
+         }
+      }
+      if (i > 0) {
+        tmp[k] = '.';
+        if (++k >= tmpsz) return CRYPT_ERROR;
+      }
+   }
+   if (*outlen < k + 1) return CRYPT_ERROR;
```
sjaeckel: how about returning `CRYPT_BUFFER_OVERFLOW` etc.?
I know it's still a "private" function, but then it'd already be "public-ready" :)

src/pk/ecc/ecc_set_dp_internal.c
```diff
@@ -0,0 +1,132 @@
...
+
+#include "tomcrypt.h"
+
+#ifdef LTC_MECC
+
+static int _ecc_cmp_hex_bn(const char *left_hex, void *right_bn)
+{
+   void *bn;
+   int match = 0;
+   if (mp_init(&bn) != CRYPT_OK)                    return 0;
```
sjaeckel: can you please add this as parameter so we don't have to do init&clear for each compare operation?

karel-m: done

src/misc/pk_oid_str.c
karel-m: done

src/pk/ecc/ecc_get_curve_by_name.c
karel-m: Do we really need something like `ecc_curve_get_name()`? What would be the use case?

buggywhip:
>>`err = ecc_get_curve_by_name("1.3.132.0.31", &curve);`
>>
>>I know that the name is a bit misleading but we can explain it in documentation.
>
>how about renaming it to `ecc_get_curve(const char *id, const ltc_ecc_curve **cu)`?

For me the doc should be sufficient.    ...and I kinda like the shorter name.  But I'd like to suggest we extend the arg `id` to `identifier`?  ...which can more readily encompass and associate both name and OID.
    `ecc_get_curve(const char *identifier, const ltc_ecc_curve **cu)`

karel-m:
In the end I have decided for: 
```C
int ecc_get_curve(const char* name_or_oid, const ltc_ecc_curve** cu);
```

buggywhip:
> In the end I have decided for:
> 
> int ecc_get_curve(const char* name_or_oid, const ltc_ecc_curve** cu);

I just saw that.  Bingo!

src/pk/ecc/ecc_get_curve_by_name.c
sjaeckel: > What would be the use case?

A CLI tool similar to `openssl ecparam -list_curves | grep 'prime field'`?
We can also leave it as it is for now and add the API as soon as that tool is written ;)

karel-m:
@fperrad could you please check (const params + usual stuff) with your linter `pr/ecc-non-asn1-part` branch?

fperrad:
@karel-m here, the harvest:

```
--- Module:   src/pk/ecc/ecc_verify_hash.c (C)
                                                  _
   if ((err = mp_add_d(a, 3, a_plus3)) != CRYPT_OK) {
src/pk/ecc/ecc_verify_hash.c  46  Info 838: Previously assigned value to variable 'err' has not been used

--- Module:   src/pk/ecc/ltc_ecc_verify_key.c (C)
                        _
   prime = key->dp.prime;
src/pk/ecc/ltc_ecc_verify_key.c  30  Info 838: Previously assigned value to variable 'prime' has not been used
                        _
   order = key->dp.order;
src/pk/ecc/ltc_ecc_verify_key.c  31  Info 838: Previously assigned value to variable 'order' has not been used
                    _
   a     = key->dp.A;
src/pk/ecc/ltc_ecc_verify_key.c  32  Info 838: Previously assigned value to variable 'a' has not been used
```

```
src/pk/ecc/ecc_sign_hash.c  120  Info 818: Pointer parameter 'key' (line 21) could be declared as pointing to const
src/pk/ecc/ltc_ecc_points.c  56  Info 818: Pointer parameter 'p' (line 49) could be declared as pointing to const
src/pk/ecc/ltc_ecc_points.c  65  Info 818: Pointer parameter 'dst' (line 58) could be declared as pointing to const
src/pk/ecc/ecc_get_key.c  54  Info 818: Pointer parameter 'key' (line 22) could be declared as pointing to const
src/pk/ecc/ecc_ansi_x963_export.c  62  Info 818: Pointer parameter 'key' (line 25) could be declared as pointing to const
src/pk/ecc/ecc_export.c  65  Info 818: Pointer parameter 'key' (line 27) could be declared as pointing to const
src/pk/ecc/ecc_verify_hash.c  160  Info 818: Pointer parameter 'key' (line 21) could be declared as pointing to const
src/pk/ecc/ecc_get_size.c  30  Info 818: Pointer parameter 'key' (line 24) could be declared as pointing to const
src/pk/ecc/ecc_shared_secret.c  70  Info 818: Pointer parameter 'private_key' (line 27) could be declared as pointing to const
src/pk/ecc/ecc_shared_secret.c  70  Info 818: Pointer parameter 'public_key' (line 27) could be declared as pointing to const
src/pk/ecc/ltc_ecc_verify_key.c  67  Info 818: Pointer parameter 'key' (line 22) could be declared as pointing to const
```

karel-m:
@fperrad I have fixed all of them except:
```
src/pk/ecc/ltc_ecc_points.c  56  Info 818: Pointer parameter 'p' (line 49) could be declared as pointing to const
src/pk/ecc/ltc_ecc_points.c  65  Info 818: Pointer parameter 'dst' (line 58) could be declared as pointing to const
```
These looks like false positives. Please re-run the check again.

fperrad:
I agree look like false positive.

There are 2 new ones: 
```
src/pk/ecc/ecc_decrypt_key.c  136  Info 818: Pointer parameter 'key' (line 30) could be declared as pointing to const
src/pk/ecc/ecc_encrypt_key.c  129  Info 818: Pointer parameter 'key' (line 34) could be declared as pointing to const
```

fperrad:
I missed this one:
```
Info 765: external '_curve_names' (line 17, file src/pk/ecc/ecc_get_curve.c) could be made static
```

src/headers/tomcrypt_custom.h
```diff
@@ -494,6 +494,41 @@
...
+   #define LTC_ECC_SECP160R2
+   #define LTC_ECC_SECP192K1
+   #define LTC_ECC_SECP192R1
+   #define LTC_ECC_SECP224K1
+   #define LTC_ECC_SECP224R1
+   #define LTC_ECC_SECP256K1
+   #define LTC_ECC_SECP256R1
+   #define LTC_ECC_SECP384R1
+   #define LTC_ECC_SECP521R1
+   /* OLD deprecated (but still working) defines */
```
sjaeckel: [0001-don-t-use-deprecated-LTC_ECCxxx-defines-anymore.patch.txt](https://github.com/libtom/libtomcrypt/files/1907292/0001-don-t-use-deprecated-LTC_ECCxxx-defines-anymore.patch.txt)

what do you think about applying this patch?

sjaeckel: applied & I was a bit surprised when the `./test ecc`-run returned a `nop` instead of a `passed` :)

sjaeckel:
Go for the docs! :+1:

sjaeckel:
And feel free to squash again :)

sjaeckel:
> FYI: I have tried to build dropbear with the API changes included in this PR. Unfortunately we are going to break dropbear build quite severely. I'll try to investigate whether we can somehow lower the impact.
> 
> Or maybe we should simply submit a patch for dropbear alongside the release of this new API.

I'll have a look and probably @mkj can add his two cents as well :)

mkj:
I'm fine with updating Dropbear to work with the new API - I've only had a quick look over it so far though. `ecc_set_key()` can replace Dropbear's `buf_get_ecc_raw_pubkey()`, and places Dropbear uses `key->dp` will use `ltc_ecc_curve` instead.

My main concern is how the ABI breakage happens if people have built Dropbear using system libtomcrypt on Debian - I guess the shared library version number will bump?

karel-m:
@mkj my firtst draft (not in a pull-request quality yet) is here https://github.com/karel-m/dropbear/commit/4530ff68975932680d674a33ea477fa7afc79ade

Ad ABI breakage we definitely will bump ABI & SO versions (as it turned out it is necessary even with smaller API changes than this one)

karel-m:
Great I'll squash it to a single commit and merge to develop.

I'll open a separate PR for doc changes

PR #239, opened 2017-06-22
ASN.1 changes extracted from ECC stuff

karel-m:
As mentioned in https://github.com/libtom/libtomcrypt/pull/187#discussion_r123453123

The motivation is to make rebasing of ECC branches a bit smoother.

### DIFF
--- a/src/headers/tomcrypt_pk.h
+++ b/src/headers/tomcrypt_pk.h
@@ -559,6 +559,8 @@ int der_decode_sequence_ex(const unsigned char *in, unsigned long  inlen,
 
 int der_length_sequence(ltc_asn1_list *list, unsigned long inlen,
                         unsigned long *outlen);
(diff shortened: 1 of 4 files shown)


src/headers/tomcrypt_pk.h
```diff
@@ -559,6 +559,8 @@ int der_decode_sequence_ex(const unsigned char *in, unsigned long  inlen,
 
 int der_length_sequence(ltc_asn1_list *list, unsigned long inlen,
                         unsigned long *outlen);
+int der_length_sequence_ex(ltc_asn1_list *list, unsigned long inlen,
```
sjaeckel: do we really have to expose this as API function?

karel-m:
IMO der_length_sequence_ex can be private, for internal use only. Please,
change it, I am only on email till Saturday.

Issue #232, opened 2017-06-18
libtomcrypt.pc contains DESTDIR in prefix


andreyv:
When building the library with `makefile.shared`, the generated `libtomcrypt.pc` contains `DESTDIR` as prefix.

`DESTDIR` normally denotes a staging root directory where software is installed and then archived. E.g., if the installation prefix was `/usr` and `DESTDIR` was `/tmp/package-root`, the full installation path would be `/tmp/package-root/usr`, and then a package would be created from `/tmp/package-root`, containing `usr/*` entries inside. With the current makefiles `DESTDIR` serves the role of the installation prefix instead and there is no separate way to actually specify a staging directory.

References:
https://www.gnu.org/prep/standards/html_node/DESTDIR.html
https://www.freebsd.org/doc/en/books/porters-handbook/porting-prefix.html

sjaeckel:
IIRC we had that discussion already once @rofl0r ... :-/
Fine like that?

andreyv:
Seems good.

rofl0r:
i added some comments in the commit.

rofl0r:
what you just committed doesn't look quite right...

sjaeckel:
fine like that?

rofl0r:
looks good, thanks! there's only a minor issue with double slashes; i commented in the commit.

sjaeckel:
btw there was still a real bug in `makefile_include.mk` ...

rofl0r:
great! looks like the issue is now a thing of the past <3

PR #241, opened 2017-06-23
Fourth general clean-up

sjaeckel:
* fix issues detected by coverity
* fix potential issue in `rsa_exptmod()`
* "hide" new internal functions
* final fix of #232 
* minor clean-ups

### DIFF
--- a/src/hashes/blake2b.c
+++ b/src/hashes/blake2b.c
@@ -333,14 +333,14 @@ int blake2b_process(hash_state *md, const unsigned char *in, unsigned long inlen
    LTC_ARGCHK(md != NULL);
    LTC_ARGCHK(in != NULL);
 
-   if (md->blake2b.curlen > sizeof(md->blake2b.buf)) {
+   if (md->blake2b.curlen >= sizeof(md->blake2b.buf)) {
       return CRYPT_INVALID_ARG;
    }
 
    if (inlen > 0) {
       unsigned long left = md->blake2b.curlen;
       unsigned long fill = BLAKE2B_BLOCKBYTES - left;
-      if (inlen > fill) {
+      if (inlen >= fill) {
          md->blake2b.curlen = 0;
          XMEMCPY(md->blake2b.buf + left, in, fill); /* Fill buffer */
          blake2b_increment_counter(md, BLAKE2B_BLOCKBYTES);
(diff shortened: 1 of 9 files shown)


karel-m:
It is unrelated, but could you please fix makefile.unix in this part:
```
	-@rm -f test tv_gen hashsum crypt small timing
	-@rm -f test.exe tv_gen.exe hashsum.exe crypt.exe small.exe timing.exe
```

`s/crypt/ltcrypt/`

sjaeckel:
The blake2 modifications were done after the original code triggered a "Memory - illegal accesses  (OVERRUN)" warning in coverity. It's true that the `XMEMCPY()` in the `if (inlen > fill)` case could've been called with an invalid pointer, but `fill` would have had the value `0`. But still that was a violation IIUC and coverity also doesn't complain anymore.

src/hashes/blake2b.c
```diff
@@ -342,7 +342,7 @@ int blake2b_process(hash_state *md, const unsigned char *in, unsigned long inlen
       unsigned long fill = BLAKE2B_BLOCKBYTES - left;
       if (inlen > fill) {
          md->blake2b.curlen = 0;
-         XMEMCPY(md->blake2b.buf + left, in, fill); /* Fill buffer */
+         XMEMCPY(md->blake2b.buf + (left % sizeof(md->blake2b.buf)), in, fill); /* Fill buffer */
```
karel-m: Well, we are obviously trying here to work around the situation when `left == BLAKE2B_BLOCKBYTES`. IMHO it would be better to avoid that instead of "fixing" it by %-trick.

BTW we should perhaps use `BLAKE2B_BLOCKBYTES` instead of  `sizeof(md->blake2b.buf)`

karel-m: I think that `left == BLAKE2B_BLOCKBYTES` situation can happen when you sent 1.chunk 256B + 2.chunk of any size. But if `left == BLAKE2B_BLOCKBYTES` then `fill == 0` in which case we should perhaps avoid completely calling `XMEMCPY`

sjaeckel: I guess that we could somehow re-factor this entire part, but I didn't see an added value... therefore I went for the cheap trick to make the pointer always valid ;-)
If you think we shouldn't do it like that I'll investigate a bit further.

karel-m: No, perhaps just change it to `(left % BLAKE2B_BLOCKBYTES)`

sjaeckel: I decided to use `sizeof(md->blake2b.buf)` so it's clear that we're truncating `left` by the size of the buffer, without having to lookup how large `md->blake2b.buf` really is

PR #246, opened 2017-07-09
GCM allow skipping gcm_add_aad and gcm_process

karel-m:
This patch is a by-product of fixing issues related to my perl bindings:
* https://github.com/DCIT/perl-CryptX/issues/36
* https://github.com/DCIT/perl-CryptX/issues/37

Current implementation strictly requires to call `gcm_add_aad` and `gcm_process` even if there is no AAD and no plain/cipher text to process.

With this patch you can skip calling `gcm_add_aad` if you do not want to add any additional authenticated data (and similarly  `gcm_process`).

No need to add this to 1.18

### DIFF
--- a/src/encauth/gcm/gcm_done.c
+++ b/src/encauth/gcm/gcm_done.c
@@ -40,6 +40,15 @@ int gcm_done(gcm_state *gcm,
       return err;
    }
 
+   if (gcm->mode == LTC_GCM_MODE_IV) {
+      /* let's process the IV */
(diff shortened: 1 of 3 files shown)


sjaeckel:
we could also put that one in 1.18
if you want to, then please rebase on top of the release/1.18.0 branch, set the release/1.18.0 as target branch and merge
if you don't care if it's in, simply merge and it'll be in the next release

Issue #248, opened 2017-07-10
rc1 dsa_set_key API question


karel-m:
While switching my perl bindings to the new RC1 API for DH, DSA, RSA I was thinking a bit about `dsa_set_key` API.

In rc1 we have:
```C
int dsa_set_key(const unsigned char *pub, unsigned long publen,
                const unsigned char *priv, unsigned long privlen,
                dsa_key *key);
```
but I think it would be better to have:
```C
int dsa_set_key(const unsigned char *in, unsigned long inlen,
                int type, dsa_key *key);
```

* If you have a public key then you have a public key
* If you have a private key then you have a private key + you can simply calculate the corresponding public key (which we already do in `dsa_set_key`)

sjaeckel:
> If you have a public key then you have a public key
> If you have a private key then you have a private key + you can simply calculate the corresponding public key (which we already do in dsa_set_key)

uhm, we can still change that if it makes sense... and you meant `dh_set_key()`!?

TBH I also thought about it, but then forgot to ask you...

karel-m:
Yes `dh_set_key` but actually it can be applied to `dsa_set_key` as well (in both cases it is just a simple `mp_exptmod`).

sjaeckel:
feel free to open a PR, we'll put that still in 1.18

PR #249, opened 2017-07-10
dh_set_key + dsa_set_key

karel-m:
see #248


sjaeckel:
Now the API is at least consistent on the DH part where we have `dh_export_key()` and `dh_set_key()` with the same signature. DSA still misses that part, but I also don't know if that's really relevant.

karel-m:
> DSA still misses that part, but I also don't know if that's really relevant.

You mean missing `dsa_export_key()`? Well, it would be consistent, but I personally do not need it at the moment. The `dsa_set_key()` I use in my perl bindings when importing DSA keys in SSH format (first I extract the raw values which I pass later to dsa_set_key).

PR #250, opened 2017-07-12
README building instructions

karel-m:


karel-m:
A question: Why do we need both `-DUSE_LTM` and `-DLTM_DESC`?

The `USE_LTM` seems to be used just in some tests and demos.

sjaeckel:
> A question: Why do we need both -DUSE_LTM and -DLTM_DESC?

true, strictly speaking `-DUSE_{LTM,TFM,GMP}` is only required when building `test`, 'tv_gen` and `timing`

karel-m:
Wouldn't it simplify things a bit if we replace `USE_{LTM,TFM,GMP}` with `{LTM,TFM,GMP}_DESC` in `test`, `tv_gen` and `timing`?

Or if we add to `tomcrypt_test.h` something like:
```C
#if defined(LTM_DESC)
#define USE_LTM
#elif defined(TFM_DESC)
#define USE_TFM
#elif defined(GMP_DESC)
#define USE_GMP
#else
#error no math lib
#endif
```

sjaeckel:
> Wouldn't it simplify things a bit if we replace USE_{LTM,TFM,GMP} with {LTM,TFM,GMP}_DESC in test, tv_gen and timing?

and then you can't build the library against ltm and tfm anymore and run the tests against tfm... I (currently) don't see a better way than the one implemented.
"You want to build the tests? Then choose a math library to run the tests with."

sjaeckel:
@rofl0r fine like that?

rofl0r:
LGTM! the only thing i wondered when reading the instructions was what those mentioned binaries would be...

sjaeckel:
> the only thing i wondered when reading the instructions was what those mentioned binaries would be...

sorry, what do you mean?

rofl0r:
no, i mean, i wondered what those binaries are and do, since the only practical thing i can think of is something like the "openssl" binary, ssh-keygen, etc.

sjaeckel:
> those binaries are and do, since the only practical thing i can think of is something like the "openssl" binary, ssh-keygen, etc.

exactly, I especially like `hashsum` I think it's pretty unique in its supported hashes and it's install-worthy in my opinion
the others are... well ... demos :-)

sjaeckel:
I guess we can merge this!?

karel-m:
I guess we can merge this!?


IMO yes

PR #255, opened 2017-08-01
Fix crypt.tex with newer TeXLive (thanks Norbert Preining)

stapelberg:
See https://lists.debian.org/debian-tex-maint/2017/08/msg00000.html for details


sjaeckel:
Already fixed in 749873c306e61a96286c5c41762912e8909185e4 (waiting on the release branch to be merged back to develop)

stapelberg:
I see only the second hunk in the commit you referenced. Could you also apply the first one please?

PR #256, opened 2017-08-02
OCB3 is not according to RFC7253

sjaeckel:
While documenting OCB3 I realized that the implementation is not according to [draft-irtf-cfrg-ocb-03](https://tools.ietf.org/html/draft-irtf-cfrg-ocb-03) (which later on became rfc7253) as stated in dfb4ec3b95c1518f093491c6a1847627deffc282 but to [draft-irtf-cfrg-ocb-02](https://tools.ietf.org/html/draft-irtf-cfrg-ocb-02).

Changing the implementation would also require to change the API, therefore I'm unsure if we should patch it:
1. _'at all'_ - because not much people seem to use it as nobody complained yet
1. _'in the next version'_ - so we get 1.18.0 out (TBH I'd be fine to even remove it from the release)
1. _'now'_ - because I don't like to publish draft versions of algorithms


karel-m:
Do we support empty aad (aad==NULL, aadlen==0) and empty plaintext
(pt==NULL, ptlen==0)?

To be honest, I still have not read the RFC so I am not sure if empty
plaintext is legal here.

I'll be able to do the review no sooner than during the coming weekend.

karel-m:
We should perhaps add all test vectors from RFC 7253 - Appendix A, some of
them test empty pt/aad that I've mentioned above.

sjaeckel:
> Do we support empty aad (aad==NULL, aadlen==0)

done in b2448c593a93524c26925d384def7edc6715c623

> and empty plaintext (pt==NULL, ptlen==0)?

done in 868c5a82c31c816dc39a1ea7abed7f979d7e0bd6

> We should perhaps add all test vectors from RFC 7253 - Appendix A, some of them test empty pt/aad that I've mentioned above.

ours do that as well, I added the test of the RFC with the 96bit tag
I don't think it's worth the effort adding all the tests from the RFC...

karel-m:
I have one more `ocb3_done` related issue which I've found out when switching my perl bindings to the new OCB3 API.

I'll send it as a separate PR today evening (just want you to know if you plan to release the next RC).

PR #262, opened 2017-08-07
ocb3_done taglen fix

karel-m:
With this patch `ocb3_done(&ocb, tag, &len)` accepts `len` greater that `tag_len` value used in `ocb3_init`.


karel-m:
You mean a better (strict) `taglen` check in `ocb3_init`?

sjaeckel:
> You mean a better (strict) taglen check in ocb3_init

yep, before this patch it was implicitly limited by the fact that there's no block cipher with more than 128bit blocksize...

src/encauth/ocb3/ocb3_init.c
```diff
@@ -118,6 +118,7 @@ int ocb3_init(ocb3_state *ocb, int cipher,
    if (taglen > (unsigned long)cipher_descriptor[cipher].block_length) {
       taglen = cipher_descriptor[cipher].block_length;
```
sjaeckel: also I'm not sure anymore if that's correct

sjaeckel: ah it's even against the spec

> The blockcipher must have a 128-bit blocksize

PR #263, opened 2017-08-07
missing items in crypt sizes

karel-m:


sjaeckel:
I thought it'd probably make sense to check this in `helper.pl` but somehow it seems to be not really trivial...

buggywhip:
[I wasn't able to upload zip files here so I sent them to you via email.]

"Here are 2 zip files, one for sizes, the other for constants.  Each has an old/current .c, a new .c, and a diff file.  The new files compile without warnings and a run of the demo pgms ./sizes, ./constants, and demo_dynamic.py each show what I would expect.  I still want to run some enhanced demo pgms over the next day or so."

karel-m:
once @buggywhip is fine with these changes I propose to merge it to release/1.18.0 and create RC4

sjaeckel:
:+1:

buggywhip:
> On 9Aug, 2017, at 12:13 AM, karel-m <notifications@github.com> wrote:
> 
> once @buggywhip <https://github.com/buggywhip> is fine with these changes I propose to merge it to release/1.18.0 and create RC4
> 

I have just a couple of things to check in perhaps the next hour or two.  In the meantime...


I'd like to add the following constant to demos/crypt_constants.c near line 30.
    _C_STRINGIFY(CRYPT_OK),


Question:  What do you think about adding a function added to LTC that when called with an error code, it would return a string, a phrase indicating the nature of the error?  ...something along the lines of the comments to the errors found in the enum in tomcrypt.h.  (CRYPT_UNUSED1 and CRYPT_UNUSED2 might return "Reserved", and values not in the table might get some sort of "error code not found" string.)  Errors returned to Python are just a [meaningless?] number.  Your thots?


Here is a complete replacement for demos.demo_dynamic.py




That's my list.  Like I said, just a couple more things to check.  

RC4, "real soon now"?  :-)

sjaeckel:
There's `error_to_string ()`

sjaeckel:
> What's missing is a function to return a string saying what the code means. Such a function might be useful to both C and Python app developers.

as already said, there's https://github.com/libtom/libtomcrypt/blob/develop/src/misc/error_to_string.c

buggywhip:
> On 9Aug, 2017, at 3:02 AM, Steffen Jaeckel <notifications@github.com> wrote:
> 
> What's missing is a function to return a string saying what the code means. Such a function might be useful to both C and Python app developers.
> 
> as already said, there's https://github.com/libtom/libtomcrypt/blob/develop/src/misc/error_to_string.c <https://github.com/libtom/libtomcrypt/blob/develop/src/misc/error_to_string.c>
Mea culpa, I completely missed that.  

BTW, it looks good.  Tx.

buggywhip:
Nope!  will send separately.

PR #272, opened 2017-08-18
Update demos demo dynamic.py

buggywhip:
demos/demo_dynamic.py3 now runs under both python 2 and python3 so there is no longer a reason for two versions.  demo_dynamic.py3 was deleted and demo_dynamic.py3 was renamed demo_dynamic.py.


sjaeckel:
Cool :+1:

Issue #273, opened 2017-08-24
GCM - 0 size IV is not valid


karel-m:
I am going through: https://github.com/google/wycheproof/tree/master/testvectors

We pass all tests from `aes_gcm_test.json` except cases where `0 size IV is not valid`.

We should perhaps fail to encrypt/decrypt when the length of IV is zero.

PR #276, opened 2017-08-24
GCM zero size IV

karel-m:
see #273


src/encauth/gcm/gcm_done.c
```diff
@@ -41,6 +41,8 @@ int gcm_done(gcm_state *gcm,
    }
 
    if (gcm->mode == LTC_GCM_MODE_IV) {
+      /* IV length must be > 0 */
```
sjaeckel: this check shouldn't be required as `gcm_add_aad()` already checks for it

src/encauth/gcm/gcm_process.c
```diff
@@ -53,6 +53,8 @@ int gcm_process(gcm_state *gcm,
    }
 
    if (gcm->mode == LTC_GCM_MODE_IV) {
+      /* IV length must be > 0 */
```
sjaeckel: same

sjaeckel:
good find! :+1:

karel-m:
I see what you mean, but all those extra (duplicate) `if (gcm->mode == LTC_GCM_MODE_IV) { .. }` where added in #246 - GCM allow skipping gcm_add_aad and gcm_process.

After #246 it is legal to completely skip `gcm_add_aad` and/or `gcm_process` therefore zero IV size has to be checked IMO in all 3 places (gcm_add_aad, gcm_process, gcm_done).

sjaeckel:
okay, but these two checks are then either on the wrong place or they can be removed.

These checks are both in the state `gcm->mode == LTC_GCM_MODE_IV` immediately before a call to `gcm_add_aad()`. `gcm_add_aad()` has in the state `gcm->mode == LTC_GCM_MODE_IV` as first line the same check, therefore these two checks can be removed. Or am I wrong?

karel-m:
Oh, you are right.

Issue #274, opened 2017-08-24
DSA vs. wycheproof


karel-m:
Again https://github.com/google/wycheproof/tree/master/testvectors

From `dsa_test.json` we fail the following tests (grep the json for `tcId` value):
```#   Failed test 'type: DSAVer/SHA224, tcId: 100, comment: repeat element in sequence, result: invalid'
#   Failed test 'type: DSAVer/SHA224, tcId: 12, comment: Legacy:ASN encoding of s misses leading 0, result: acceptable'
#   Failed test 'type: DSAVer/SHA224, tcId: 126, comment: Modified r or s, e.g. by adding or subtracting the order of the group, result: invalid'
#   Failed test 'type: DSAVer/SHA224, tcId: 128, comment: Modified r or s, e.g. by adding or subtracting the order of the group, result: invalid'
#   Failed test 'type: DSAVer/SHA224, tcId: 14, comment: BER:long form encoding of length, result: acceptable'
#   Failed test 'type: DSAVer/SHA224, tcId: 17, comment: BER:length contains leading 0, result: acceptable'
#   Failed test 'type: DSAVer/SHA224, tcId: 19, comment: BER:indefinite length, result: acceptable'
#   Failed test 'type: DSAVer/SHA224, tcId: 20, comment: BER:prepending 0's to integer, result: acceptable'
#   Failed test 'type: DSAVer/SHA224, tcId: 53, comment: appending 0's to sequence, result: invalid'
#   Failed test 'type: DSAVer/SHA224, tcId: 55, comment: appending unused 0's, result: invalid'
#   Failed test 'type: DSAVer/SHA224, tcId: 57, comment: appending null value, result: invalid'
#   Failed test 'type: DSAVer/SHA224, tcId: 97, comment: append empty sequence, result: invalid'
```

On top of that the following test vector seems to loop for ever:
```
{
  "comment" : "Signatures with special case values for r and s.",
  "message" : "48656c6c6f",
  "result" : "invalid",
  "sig" : "3022020101021dff450969597a870820211805983688387a10cd4dcc451a7f3f432a96a3",
  "tcId" : 139
}
```

I'll investigate more.

karel-m:
In total: 12 tests is failing, 2 hangs (out of 206)

karel-m:
2 hanging tests fixed by https://github.com/libtom/libtomcrypt/pull/278/commits/0fada4ddf4e794c60bebb45bddeb657c73f1ea6a

[...]

PR #278, opened 2017-08-24
DSA wycheproof tests

karel-m:
see #274


karel-m:
As for the our DSA vs. wycheproof this PR is complete from my side.

PR #280, opened 2017-08-29
DSA cdf tests

karel-m:
Another crypto test bench https://github.com/kudelskisecurity/cdf

Good news - our RSA implementation seems to pass. But we have couple of DSA related troubles.

The first commit fixes failures when ltc is on "validating" side. There are still some failures when ltc is on "signing" side, I'll try to analyze those later.

### DIFF
--- a/src/headers/tomcrypt_pk.h
+++ b/src/headers/tomcrypt_pk.h
@@ -480,7 +480,7 @@ int dsa_import(const unsigned char *in, unsigned long inlen, dsa_key *key);
 int dsa_export(unsigned char *out, unsigned long *outlen, int type, dsa_key *key);
 int dsa_verify_key(dsa_key *key, int *stat);
 #ifdef LTC_SOURCE
-int dsa_int_validate_key(dsa_key *key, int *stat, int mode);
+int dsa_int_validate_key(dsa_key *key, int *stat);
 #endif
 int dsa_shared_secret(void          *private_key, void *base,
                       dsa_key       *public_key,
--- a/src/headers/tomcrypt_pk.h
+++ b/src/headers/tomcrypt_pk.h
@@ -479,7 +479,9 @@ int dsa_decrypt_key(const unsigned char *in,  unsigned long  inlen,
 int dsa_import(const unsigned char *in, unsigned long inlen, dsa_key *key);
 int dsa_export(unsigned char *out, unsigned long *outlen, int type, dsa_key *key);
 int dsa_verify_key(dsa_key *key, int *stat);
-
+#ifdef LTC_SOURCE
+int dsa_verify_key_ex(dsa_key *key, int *stat, int mode);
+#endif
 int dsa_shared_secret(void          *private_key, void *base,
                       dsa_key       *public_key,
                       unsigned char *out,         unsigned long *outlen);
--- a/src/headers/tomcrypt_pk.h
+++ b/src/headers/tomcrypt_pk.h
@@ -480,7 +480,7 @@ int dsa_import(const unsigned char *in, unsigned long inlen, dsa_key *key);
 int dsa_export(unsigned char *out, unsigned long *outlen, int type, dsa_key *key);
 int dsa_verify_key(dsa_key *key, int *stat);
 #ifdef LTC_SOURCE
-int dsa_verify_key_ex(dsa_key *key, int *stat, int mode);
+int dsa_int_validate_key(dsa_key *key, int *stat, int mode);
 #endif
 int dsa_shared_secret(void          *private_key, void *base,
                       dsa_key       *public_key,
--- a/src/headers/tomcrypt_pk.h
+++ b/src/headers/tomcrypt_pk.h
@@ -480,6 +480,7 @@ int dsa_import(const unsigned char *in, unsigned long inlen, dsa_key *key);
 int dsa_export(unsigned char *out, unsigned long *outlen, int type, dsa_key *key);
 int dsa_verify_key(dsa_key *key, int *stat);
 #ifdef LTC_SOURCE
+/* internal helper functions */
 int dsa_int_validate_xy(dsa_key *key, int *stat);
 int dsa_int_validate_pqg(dsa_key *key, int *stat);
 int dsa_int_validate_primes(dsa_key *key, int *stat);
--- a/src/headers/tomcrypt_pk.h
+++ b/src/headers/tomcrypt_pk.h
@@ -480,7 +480,9 @@ int dsa_import(const unsigned char *in, unsigned long inlen, dsa_key *key);
 int dsa_export(unsigned char *out, unsigned long *outlen, int type, dsa_key *key);
 int dsa_verify_key(dsa_key *key, int *stat);
 #ifdef LTC_SOURCE
-int dsa_int_validate_key(dsa_key *key, int *stat);
+int dsa_int_validate_xy(dsa_key *key, int *stat);
+int dsa_int_validate_pqg(dsa_key *key, int *stat);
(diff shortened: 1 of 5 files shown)


karel-m:
Well, it turned out that the case when ltc is on "signing" side is OK. So this PR is complete.

karel-m:
@sjaeckel ping

sjaeckel:
We could e.g. have an internal (say `LTC_SOURCE` protected) `dsa_check_key()` which is then called from within `dsa_verify_key()`, `dsa_set_pqg_dsaparam()` and `dsa_set_pqg()` (and probably others that I missed?).

karel-m:
something like this?

sjaeckel:
Do we really need such extensive tests? (the 2 exptmod's?)
Besides that I realized that the name `verify_key()` is nonsense, therefore I've pushed my proposal to the pr/fix-dsa-cdf-2 branch. What do you think?

1. Can we make `dsa_int_validate_key()` a bit less computational expensive by moving the 2 exptmods to `dsa_validate_key()`?
2. Wouldn't it make sense to use `LTC_MILLER_RABIN_REPS` for the prime checks?
3. With your last commit you removed stuff that you added in the commit before (in `dsa_set_pqg()`)
a. Was that accidental or itentional?
b. The 'newly added' was three `mp_cmp_d()` whereas `dsa_verify_key_ex()` has only two + a different logic, can you please check back which is correct!?

karel-m:
we should perhaps keep the name `dsa_verify_key` as it is part of 1.17 API

sjaeckel:
> we should perhaps keep the name `dsa_verify_key` as it is part of 1.17 API

damn, sorry - sure

I'll just update the branch

karel-m:
> With your last commit you removed stuff that you added in the commit before (in `dsa_set_pqg()`)

The thing is that at the end of `dsa_set_pqg()` we do not have a complete key, therefore we cannot do the key check/validation.

sjaeckel:
> The thing is that at the end of `dsa_set_pqg()` we do not have a complete key, therefore we cannot do the key check/validation.

but it looks to me like part of it was required for the cdf tests to pass, right?

sjaeckel:
> I'll just update the branch

pushed

karel-m:
What we need is to avoid loading/setting/creating a DSA key with `p` or `g` or `q` (not sure what was the main trouble maker) with value `0`. 

It would be better to check it in `dsa_set_pqg` but it is IMO fine to do it a bit later during `dsa_set_key` (where we have the key complete).

karel-m:
I have slightly updated your pr/fix-dsa-cdf-2 - which works and can be merged

sjaeckel:
> What we need is to avoid loading/setting/creating a DSA key with p or g or q (not sure what was the main trouble maker) with value 0.

Okay, but the different parts of the implementation I saw had checks for `{p,g,q} > 1` `|p| > 1` and I don't know what else. Reading quickly through FIPS 186-4 I couldn't determine what is the correct approach so can you please check again what we really have to do?

I think it'd make sense to have 2 private functions `dsa_int_validate_group_params()` and `dsa_int_validate_key()` and 1 API function `dsa_verify_key()`. This allows us to correctly verify each step of the key import/generation.

Flow could be something like this:

```
int dsa_verify_key() {
 check_prime(q);
 check_prime(p);
 return dsa_int_validate_key();
}

int dsa_int_validate_group_params() {
 ...
}

int dsa_int_validate_key() {
 if (dsa_int_validate_group_params()) return error;
 ...
}
```

What do you think?

karel-m:
I'll check FIPS 186-4 but meanwhile another approach for `dsa_int_validate_*`

karel-m:
@sjaeckel could you please review this PR again?

src/pk/dsa/dsa_verify_key.c
```diff
@@ -16,81 +16,183 @@
...
+   @param key   The key to validate
+   @param stat  [out]  Result of test, 1==valid, 0==invalid
+   @return CRYPT_OK if successful
+*/
+int dsa_int_validate_pqg(dsa_key *key, int *stat)
+{
+   void *tmp1, *tmp2;
+   int  err;
+
+   *stat = 0;
```
sjaeckel: ew, de-referencing stat before checking for NULL

src/headers/tomcrypt_pk.h
```diff
@@ -479,7 +479,11 @@ int dsa_decrypt_key(const unsigned char *in,  unsigned long  inlen,
 int dsa_import(const unsigned char *in, unsigned long inlen, dsa_key *key);
 int dsa_export(unsigned char *out, unsigned long *outlen, int type, dsa_key *key);
 int dsa_verify_key(dsa_key *key, int *stat);
-
+#ifdef LTC_SOURCE
```
sjaeckel: can you please add the obligatory `/* internal helper functions */` comment

src/pk/dsa/dsa_verify_key.c
```diff
@@ -16,81 +16,183 @@
...
+   if ((err = mp_exptmod(key->g, key->q, key->p, tmp1)) != CRYPT_OK) { goto error; }
+   if (mp_cmp_d(tmp1, 1) != LTC_MP_EQ) {
+      err = CRYPT_OK;
+      goto error;
+   }
+
+   err   = CRYPT_OK;
+   *stat = 1;
+error:
+   mp_clear_multi(tmp1, tmp2, NULL);
```
sjaeckel: can you please (always) reverse the order when clearing MPI's

src/pk/dsa/dsa_import.c
```diff
@@ -116,10 +116,21 @@ int dsa_import(const unsigned char *in, unsigned long inlen, dsa_key *key)
...
+   key->qord = mp_unsigned_bin_size(key->q);
 
-  if (key->qord >= LTC_MDSA_MAX_GROUP || key->qord <= 15 ||
-      (unsigned long)key->qord >= mp_unsigned_bin_size(key->p) || (mp_unsigned_bin_size(key->p) - key->qord) >= LTC_MDSA_DELTA) {
+   /* quick p, q, g validation, without primality testing */
+   if ((err = dsa_int_validate_pqg(key, &stat)) != CRYPT_OK) {
+      goto LBL_ERR;
+   }
+   if (stat == 0) {
+      err = CRYPT_INVALID_PACKET;
```
sjaeckel: Looks like we inconsistently return either `CRYPT_INVALID_PACKET` or `CRYPT_INVALID_ARG`, we should always use the same error code. I'd say we should go for `CRYPT_INVALID_PACKET` (an alternative could be to introduce a new error code, but not sure if that's really necessary).

src/pk/dsa/dsa_verify_key.c
```diff
@@ -16,81 +16,183 @@
...
+   int err, res;
+
    *stat = 0;
+   LTC_ARGCHK(key  != NULL);
+   LTC_ARGCHK(stat != NULL);
 
-   /* first make sure key->q and key->p are prime */
-   if ((err = mp_prime_is_prime(key->q, 8, &res)) != CRYPT_OK) {
+   /* key->q prime? */
+   if ((err = mp_prime_is_prime(key->q, LTC_MILLER_RABIN_REPS, &res)) != CRYPT_OK) {
```
sjaeckel: :+1:

karel-m:
Oh sorry, I should perhaps wait for Travis-CI

sjaeckel:
Too late, the build's going to fail ;)

sjaeckel:
but nvm, as the merge will be built as well...

PR #289, opened 2017-09-14
scan-build --status-bugs

karel-m:
Travis-CI is expected to fail


karel-m:
Good, failed as expected https://travis-ci.org/libtom/libtomcrypt/jobs/275612218

PR #290, opened 2017-09-15
Add -Wwrite-strings (char* vs. const char*)

karel-m:
This PR fixes warnings which appeared after turning on `-Wwrite-strings`.

But frankly, I am not sure whether we want or need to apply this.

### DIFF
--- a/makefile_include.mk
+++ b/makefile_include.mk
@@ -70,6 +70,7 @@ LTC_CFLAGS += -Wextra
 LTC_CFLAGS += -Wsystem-headers -Wbad-function-cast -Wcast-align
 LTC_CFLAGS += -Wstrict-prototypes -Wpointer-arith
 LTC_CFLAGS += -Wdeclaration-after-statement
+LTC_CFLAGS += -Wwrite-strings
 endif
 
 LTC_CFLAGS += -Wno-type-limits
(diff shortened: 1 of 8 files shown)


sjaeckel:
> But frankly, I am not sure whether we want or need to apply this.

somehow you're right, we should probably wait until the release is done to apply this... not sure though... also in regards to ABI breakage...

sjaeckel:
@lvc could you please already tell us how this would influence our score? :-)

karel-m:
As for the API breakage I have doubts about changes in `ltc_ecc_set_type` the rest IMO should not cause big troubles.

karel-m:
Suggestion: let's merge to `release/1.18.0` the changes in `tests/*` + changes in `<name>_test()` functions. These are not going to break anything. After that we can discuss what will be left.

@sjaeckel what do you think?

sjaeckel:
K, then split the commits up please.

karel-m:
Done #292 

Once we have OK from Travis-CI I'll merge #292 to `release/1.18.0`.

karel-m:
Merged & rebased.

Now looking at the rest I think it is pretty safe to also merge:
* changes to `ltc_dh_set_type` (DH is completely new in 1.18)
* changes to `crypt_argchk`

PR #294, opened 2017-09-19
TFM + radix to bin fails for long inputs

karel-m:
`mpi_test` passes with LTM but fails with TFM


sjaeckel:
Thx

sjaeckel:
...

576 * 8 = 4608 = 4096+(8*DIGIT_BIT)

TFM default max size has been hit (again...)

should we probably increase the default size...?

karel-m:
IMO we should at least return `CRYPT_ERROR` not `CRYPT_OK`.

Perhaps TFM's `fp_read_radix` needs some overflow detection.

Anyway, not a libtomcrypt bug.

sjaeckel:
We should probably re-write `radix_to_bin()` to be independent of the MPI provider...

PR #297, opened 2017-09-22
Improve MPI provider selection

sjaeckel:
With this patch you can build the library with `-DLTM_DESC -DTFM_DESC -DGMP_DESC` and then select the MPI provider at runtime of the _tools_ `test` and `timing`

I was lazy regarding the parameter parsing and therefore you have to pass the MPI provider as 2nd argument... Running all tests or timings requires you to execute `./test _ gmp` or something like that where arg1 matches for all cases... :)

Regarding the static `init_mpi()` function: I already thought about replacing the three functions by this single one /cc @buggywhip

### DIFF
--- a/demos/timing.c
+++ b/demos/timing.c
@@ -1399,45 +1399,6 @@ static void time_encmacs(void)
    time_encmacs_(32);
 }
 
-static void init_mpi(const char* mpi)
-{
-   switch (mpi[0]) {
-#ifdef LTM_DESC
-      case 'l':
-         init_LTM();
-         break;
-#endif
-#ifdef TFM_DESC
-      case 't':
-         init_TFM();
-         break;
-#endif
-#ifdef GMP_DESC
-      case 'g':
-         init_GMP();
-         break;
-#endif
-#ifdef EXT_MATH_LIB
-      case 'e':
-         {
-            extern ltc_math_descriptor EXT_MATH_LIB;
-            ltc_mp = EXT_MATH_LIB;
-         }
-
-#define NAME_VALUE(s) #s"="NAME(s)
-#define NAME(s) #s
-         printf("EXT_MATH_LIB = %s\n", NAME_VALUE(EXT_MATH_LIB));
-#undef NAME_VALUE
-#undef NAME
-
-         break;
-#endif
-      default:
-         printf("Unknown/Invalid MPI provider: %s\n", mpi);
-         break;
-   }
-}
-
 #define LTC_TEST_FN(f)  { f, #f }
 int main(int argc, char **argv)
 {
@@ -1488,7 +1449,7 @@ register_all_prngs();
       mpi_provider = argv[2];
    }
 
-   init_mpi(mpi_provider);
+   crypt_mp_init(mpi_provider);
 
 if ((err = rng_make_prng(128, find_prng("yarrow"), &yarrow_prng, NULL)) != CRYPT_OK) {
    fprintf(stderr, "rng_make_prng failed: %s\n", error_to_string(err));
--- a/demos/timing.c
+++ b/demos/timing.c
@@ -520,19 +520,14 @@ static void time_hash(void)
 }
 
 /*#warning you need an mp_rand!!!*/
-#if !defined(USE_LTM) && !defined(USE_TFM) && !defined(USE_GMP) && !defined(EXT_MATH_LIB)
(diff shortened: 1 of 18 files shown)


sjaeckel:
btw. I think we should push this to `next`, I just wanted to get rid of the patches locally :)

karel-m:
@sjaeckel consider cherry-picking small changes in `tomcrypt_misc.h` to RC5

https://github.com/libtom/libtomcrypt/pull/297/files#diff-4e531399a5202dc48f92ed35d3860159

karel-m:
I am for merging this one to develop

buggywhip:
> On 10Oct, 2017, at 3:04 PM, karel-m <notifications@github.com> wrote:
> 
> I am for merging this one to develop
> 

I agree.  (I have been doing this without issue for a while now.)


.

sjaeckel:
what do you think about including `init_mpi()` in the library? probably then call it `crypt_mpi_init()` but I'd leave the other `init_X()` in there for now

buggywhip:
> On 10Oct, 2017, at 11:07 PM, Steffen Jaeckel <notifications@github.com> wrote:
> 
> what do you think about including init_mpi() in the library? probably then call it crypt_mpi_init() but I'd leave the other init_X() in there for now
> 
I like the idea.  It would allow for more than just LTM, TFM, and GMP.  :-))

...but then we'd either need a way 1) statically, at least a partial compile and relink, which is not that far off fm what we have now, or 2) dynamically populate and load a DESC and then make the call to its init.  ...in C.  ...but then this is computer science [fiction?], right?  ;-)

Play with it some to see what your real problems would be.  


.

karel-m:
> what do you think about including init_mpi() in the library

I do not think `init_mpi("ltm");` is a significant improvement comparing to `init_LTM();`

sjaeckel:
I'd say we should just leave the concept as it is now, I like it and it allows already for more than just the supplied MPI providers, you just have to provider your own version of the `ltc_math_descriptor`.

> I do not think `init_mpi("ltm");` is a significant improvement comparing to `init_LTM();`

yeah, it isn't a significant improvement but it would then be as it should've been the way to go in the first place.

buggywhip:
> On 11Oct, 2017, at 1:22 AM, Steffen Jaeckel <notifications@github.com> wrote:
> 
> you just have to provide your own version of the ltc_math_descriptor.
> 

...and and a linkable init function.


.

sjaeckel:
> ...and and a linkable init function.

what do you mean?

buggywhip:
...and and a linkable init function.
    what do you mean?

Won't you need the equivalent of init_LTM() as found in misc/crypt/crypt_init.c to copy the DESC to ltc_mp?  And will it not need to be linked in?  

The alternative is to try for a binding at run time ("dynamic") which IIRC has special challenges of its own.  For example, the arg to init_mpi() would need more than the simple equivalent of "LTM".  That arg would need to provide sufficient info to find the dynamic/shared MP library so it could be loaded.  Yes/no?

sjaeckel:
IIUC you can go the way as OP-TEE did https://github.com/OP-TEE/optee_os/blob/master/core/lib/libtomcrypt/src/mpa_desc.c and simply define your own variable of type `ltc_math_descriptor` called `ltc_mp` and the linker will use that one

buggywhip:
That said....

While the dynamic approach is appealing, I believe it is too complex for what I think will be the typical use case.  IMO the developer will know which MP she wants to use and will likely have the sources to populate a DESC and create a init_XXX().  Much simpler?

buggywhip:
then you don't even have to call anything to init

Hmmm...  a new trick?  ;-)

sjaeckel:
> Hmmm... a new trick? ;-)

nah, that's the job of the linker: it resolves symbols for you... :)

(more or less) easy example: 
1. take src/math/ltm_desc.c and c&p lines 15-506 into test/test.c
2. now change `const ltc_math_descriptor ltm_desc` to `ltc_math_descriptor ltc_mp` in test/test.c
3. add `#undef LTC_SOURCE` before the first include of test/test.c
4. rebuild test and see :)

I won't even try to explain how&why this works as @rofl0r can certainly explain the details/important parts better

karel-m:
What about some doc update?
* obviously something about `crypt_mp_init`
* perhaps also a new chapter about deprecated functions

PR #299, opened 2017-09-26
libtool + cygwin

karel-m:
While hacking on SONAME issue (discussed off-github) I found out that successful building of a shared library on cygwin requires to pass `-no-undefined` param to libtool.

If there is a RC5 than this should be merged. But it can also wait for the `next`.


karel-m:
@sjaeckel please merge this yourself depending on your plan of the next steps in the release process

PR #301, opened 2017-09-28
Fixes large heap over-allocation in rsa_import

rswindell:
Bug-fix: MAX_RSA_SIZE is the maximum RSA key size in bits (as commented in tomcrypt_custom.h), so the proper conversion to bytes (as the argument value to XCALLOC) would be to divide by 8 (bits per byte), not multiply by 8. This excessive allocation of heap memory (32 Kbytes instead of 512 bytes) is readily apparent in memory-constrained environments.

### DIFF
--- a/src/pk/rsa/rsa_import.c
+++ b/src/pk/rsa/rsa_import.c
@@ -40,7 +40,7 @@ int rsa_import(const unsigned char *in, unsigned long inlen, rsa_key *key)
    }
 
    /* see if the OpenSSL DER format RSA public key will work */
-   tmpbuf_len = MAX_RSA_SIZE * 8;
+   tmpbuf_len = MAX_RSA_SIZE / 8;
    tmpbuf = XCALLOC(1, tmpbuf_len);
    if (tmpbuf == NULL) {
        err = CRYPT_MEM;


mattkelly:
Hah, nice find!

sjaeckel:
especially since `der_decode_subject_public_key_info()` will also allocate 32k ;)

sjaeckel:
`LTC_DER_MAX_PUBKEY_SIZE` isn't enough in `der_decode_subject_public_key_info()`, but at least we shouldn't need `LTC_DER_MAX_PUBKEY_SIZE * 8`... there's probably a way to find out what we really need to allocate...

sjaeckel:
> there's probably a way to find out what we really need to allocate

I guess the easiest would be to use `inlen`

karel-m:
Could you please add a test case to tests/rsa_test.c which will make a key
of size MAX_RSA_SIZE/8 and then export-priv, import-priv, export-pub,
import-pub.

sjaeckel:
> export-priv, import-priv, export-pub, import-pub

+ export-pub&PK_STD, import-pub (

karel-m:
As for the actual fix, Steffen's inlen idea is IMO better e.g.

tmpbuf = XMALLOC(inlen);

sjaeckel:
Btw I tried it out and needs some more tweaking! `MALLOC(inlen)` isn't enough, that fails

sjaeckel:
I've pushed the fix to the [fix/pr/301](https://github.com/libtom/libtomcrypt/tree/fix/pr/301) branch, tests still missing.

sjaeckel:
Closed in favor of #302

PR #311, opened 2017-10-12
Improve travis build

sjaeckel:
Introduce a new meta_build where we build several things in one spin-up of a travis instance

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ addons:
 
 install:
     - sudo apt-get update -qq
-    - sudo apt-get install libtommath-dev
+    - sudo apt-get install libtommath-dev libgmp-dev
 
 before_script:
   - gem install coveralls-lcov
@@ -38,17 +38,9 @@ script:
   - bash "${BUILDSCRIPT}" "${BUILDNAME}" "${BUILDOPTIONS}" "makefile.shared V=1" "-DUSE_TFM -DTFM_DESC" "-ltfm"
 env:
   - |
-    BUILDSCRIPT="check_source.sh"
-    BUILDNAME="CHECK_SOURCES"
-    BUILDOPTIONS=" "
-  - |
-    BUILDSCRIPT="scan_build.sh"
-    BUILDNAME="SCAN_BUILD"
-    BUILDOPTIONS=" "
-  - |
-    BUILDSCRIPT="coverage.sh"
-    BUILDNAME="COVERAGE"
-    BUILDOPTIONS=" "
+    BUILDSCRIPT="meta_builds.sh"
+    BUILDNAME="META_BUILS"
+    BUILDOPTIONS="-DGMP_DESC"
   - |
     BUILDSCRIPT="run.sh"
     BUILDNAME="STOCK"
@@ -113,14 +105,6 @@ env:
     BUILDSCRIPT="run.sh"
     BUILDNAME="CLEANSTACK+NOTABLES+SMALL+NO_ASM+NO_TIMING_RESISTANCE+PTHREAD"
     BUILDOPTIONS="-DLTC_CLEAN_STACK -DLTC_NO_TABLES -DLTC_SMALL_CODE -DLTC_NO_ECC_TIMING_RESISTANT -DLTC_NO_RSA_BLINDING -DLTC_PTHREAD"
-  - |
-    BUILDSCRIPT="testbuild.sh"
-    BUILDNAME="NOTEST"
-    BUILDOPTIONS="-DLTC_NO_TEST"
-  - |
-    BUILDSCRIPT="testbuild.sh"
-    BUILDNAME="NOFILE"
-    BUILDOPTIONS="-DLTC_NO_FILE"
 
 after_failure:
   - cat test_std.txt
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ addons:
 
 install:
     - sudo apt-get update -qq
-    - sudo apt-get install libtommath-dev libgmp-dev
+    - sudo apt-get install libtommath-dev libgmp-dev valgrind
 
 before_script:
   - gem install coveralls-lcov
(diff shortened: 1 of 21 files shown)


.travis.yml
```diff
@@ -13,7 +13,7 @@ addons:
 
 install:
     - sudo apt-get update -qq
-    - sudo apt-get install libtommath-dev
+    - sudo apt-get install libtommath-dev libgmp-dev valgrind
```
karel-m: This is what was needed in ecc branch (latest ltm + tfm)
https://github.com/libtom/libtomcrypt/blob/pr/ecc-asn1-part/.travis.yml#L22

sjaeckel: yeah, I'd like to stay on standard-built packages as long as possible as the packages that are installed in the ecc branch are built by me ;)

meta_builds.sh
```diff
@@ -0,0 +1,62 @@
...
+   make clean &>/dev/null
+
+   make CFLAGS="$2 $CFLAGS $4" EXTRALIBS="$5" test LTC_DEBUG=1 1>gcc_1.txt 2>gcc_2.txt
+
+   valgrind --error-exitcode=666 --leak-check=full --show-leak-kinds=all --errors-for-leak-kinds=all ./test 1>test_std.txt 2> test_err.txt
+
+   make clean &>/dev/null
+
+   make CFLAGS="-fsanitize=address -fno-omit-frame-pointer -static-libasan $2 $CFLAGS $4" EXTRALIBS="-lasan $5" test LTC_DEBUG=1 1>gcc_1.txt 2>gcc_2.txt
+   ASAN_OPTIONS=verbosity=1 ./test t ltm 1>test_std.txt 2> test_err.txt
```
karel-m: shouldn't ASAN go into `run_clang` ?

sjaeckel: I thought so too ... then I wasn't able to get the ASAN build done with clang which can be installed in xenial ... but it worked with gcc, so I thought let's give it a try...

src/misc/crypt/crypt_inits.c
```diff
@@ -37,6 +37,54 @@ void init_GMP(void)
 }
 #endif
 
+int crypt_mp_init(const char* mpi)
```
karel-m: I am still not sure if we need this. I am not strongly against, just in doubts.

src/headers/tomcrypt_cfg.h
```diff
@@ -277,6 +277,15 @@ typedef unsigned long ltc_mp_digit;
...
 #endif
 
+#ifdef __GNUC__
+   #define LTC_DEPRECATED __attribute__((deprecated))
+#elif defined(_MSC_VER)
+   #define LTC_DEPRECATED __declspec(deprecated)
+#endif
+
+#ifndef LTC_DEPRECATED
+   #error "You need to define LTC_DEPRECATED for this compiler"
```
karel-m: I do not like this. It is very unfriendly to non-gcc/non-msvc compilers.

src/misc/crypt/crypt_inits.c
sjaeckel: No we absolutely don't need it, but I thought it's better than the other init functions that exist, so we just remove them ASAP (that's also why I marked them already as deprecated)

scan_build.sh
```diff
@@ -1,5 +1,11 @@
 #!/bin/bash
```
karel-m: Is it worth a separate script (`scan_build.sh`)? What about incorporating the stuff into `meta_builds.sh`?

src/headers/tomcrypt_cfg.h
sjaeckel: I hoped you're going to build this on "the other compilers" and then either fill it in or define it to nothing if the "deprecation" feature doesn't exist for "the other compilers"

scan_build.sh
sjaeckel: nope, I just re-used it as it mostly worked OOTB - I'm going to incorporate it.

src/misc/crypt/crypt_inits.c
sjaeckel: damn, now I know why this change is in this PR, I changed the base branch :(

src/headers/tomcrypt_cfg.h
karel-m: Consider:
```C
#ifdef __GNUC__
   #define LTC_DEPRECATED __attribute__((deprecated))
#elif defined(_MSC_VER)
   #define LTC_DEPRECATED __declspec(deprecated)
#else
   #define LTC_DEPRECATED
#endif

It might also cause troubles on older gcc as I do not know since what version this attribute is supported.
```

sjaeckel: looks like [gcc-3.1](https://gcc.gnu.org/onlinedocs/gcc-3.1/gcc/Type-Attributes.html) introduced it as it wasn't in [gcc-3.0.4](https://gcc.gnu.org/onlinedocs/gcc-3.0.4/gcc/Type-Attributes.html)

karel-m: It seems to be supported on gcc 3.1+

src/misc/crypt/crypt_inits.c
sjaeckel: let's focus on the other PR first, I've still to get this one to build

src/headers/tomcrypt_cfg.h
karel-m: `#if (__GNUC__ > 3) || (__GNUC__ == 3 && __GNUC_MINOR__ >= 0)`

karel-m:
closed? what about `meta_builds.sh`?

sjaeckel:
github closed this when I deleted the base branch and somehow I can't re-open... @github

karel-m:
try to rebase on develop locally and `push --force`

sjaeckel:
did that already (did it now once again) but the reopen button is greyed out

PR #312, opened 2017-10-16
IDEA cipher

karel-m:
Adding the old IDEA cipher

### Checklist


src/ciphers/idea.c
```diff
@@ -0,0 +1,252 @@
...
+   }
+   d_key[i*6+0] = _mul_inv(e_key[(LTC_IDEA_ROUNDS-i)*6+0]);
+   d_key[i*6+1] = _add_inv(e_key[(LTC_IDEA_ROUNDS-i)*6+1]);
+   d_key[i*6+2] = _add_inv(e_key[(LTC_IDEA_ROUNDS-i)*6+2]);
+   d_key[i*6+3] = _mul_inv(e_key[(LTC_IDEA_ROUNDS-i)*6+3]);
+
+   return CRYPT_OK;
+}
+
+static int _process_block(const unsigned char *in, unsigned char *out, ushort16 *m_key)
```
sjaeckel: should we implement `LTC_CLEAN_STACK` variants immediately?

karel-m: see ee484a0bc0

karel-m:
The other thing is that nowadays it seems to be a better practice to avoid the code where jumps/ifs are decided on the `key` value (or its part). Here it IMO might mean to find out "if-less" version of `_MUL` macro; however, I am afraid we do not care (yet) too much about this sort of side-channel attacks at a few more places.

sjaeckel:
> nowadays it seems to be a better practice to avoid the code where jumps/ifs are decided on the key value (or its part)

sure, that's to prevent timing analysis ...and if you ever want to dig into that topic we're going to have a hard time I guess... whereas the `burn_stack()` approach prevents some other function stumbling over some memory that contains (part of) your keys. So I think it's perfectly fine to continue supporting it and we should probably even think about enabling it per default

karel-m:
> ...and if you ever want to dig into that topic

well, no 😃

sjaeckel:
> that way you can use burn_stack everywhere without littering all the code with ugly ifdefs, and have the ifdef contained in a single place only.

sounds like a good idea, probably #7 is a good task where we can look into that!

PR #315, opened 2017-10-18
Improve/travis build

sjaeckel:
Follow-up of #311


karel-m:
What about moving travis-ci related scripts to `./ci/*.sh` subdir or rename them like `ci_*.sh`?

PR #318, opened 2017-10-21
Add salsa20

buggywhip:
### Checklist

* [x] tests are added or updated  (will be added soon)


karel-m:
Looks good, I would only add a simple test case for `salsa20_keystream` to `salsa20_test()` (with other stream ciphers we do not test "keastream" functions as they are called during prngs tests).

According to my testing it passes all test vectors from:
* http://www.ecrypt.eu.org/stream/svn/viewcvs.cgi/ecrypt/trunk/submissions/salsa20/reduced/12-rounds/verified.test-vectors?rev=161&view=markup
* http://www.ecrypt.eu.org/stream/svn/viewcvs.cgi/ecrypt/trunk/submissions/salsa20/reduced/8-rounds/verified.test-vectors?rev=161&view=markup
* http://www.ecrypt.eu.org/stream/svn/viewcvs.cgi/ecrypt/trunk/submissions/salsa20/full/verified.test-vectors?rev=161&view=markup

rofl0r:
not that my opinion matters, though in most projects people would try to refactor their commits so each one encapsulates a logic change, while at the same trying to produce meaningful commit messages. so in the end, you could read the commit history like a good book, well explaining why things are as they are. a good example for such practice is https://git.musl-libc.org/cgit/musl/log/

karel-m:
Good point, try `git rebase develop -i` and squash all commits into one (or two)

PR #322, opened 2017-11-02
add sosemanuk

buggywhip:
### Checklist

### DIFF
--- a/src/headers/tomcrypt_cipher.h
+++ b/src/headers/tomcrypt_cipher.h
@@ -1031,6 +1031,34 @@ int salsa20_test(void);
 
 #endif /* LTC_SALSA20 */
 
+
+
+#ifdef LTC_SOSEMANUK
+
+typedef struct {
+    ulong32 kc[100];    /* key_context */
+    ulong32 s00, s01, s02, s03, s04, s05, s06, s07, s08, s09;
+    ulong32 r1, r2;
+    /*
+     * Buffering: the stream cipher produces output data by
+     * blocks of 640 bits. buf[] contains such a block, and
+     * "ptr" is the index of the next output byte.
+     */
+    unsigned char buf[80];
+    unsigned ptr;
+} sosemanuk_state;
+
+int sosemanuk_setup(sosemanuk_state *ss, unsigned char *key, unsigned long keylen);
+int sosemanuk_setiv(sosemanuk_state *ss, unsigned char *iv, unsigned long ivlen);
+int sosemanuk_crypt(sosemanuk_state *ss, const unsigned char *in, unsigned long datalen, unsigned char *out);
+int sosemanuk_keystream(sosemanuk_state *ss, unsigned char *out, unsigned long outlen);
+int sosemanuk_done(sosemanuk_state *ss);
+int sosemanuk_test(void);
+
+#endif /* LTC_SOSEMANUK */
+
+
+
 #ifdef LTC_RC4_STREAM
 
 typedef struct {
--- a/src/stream/sosemanuk/sosemanuk.c
+++ b/src/stream/sosemanuk/sosemanuk.c
@@ -1,0 +1,819 @@
+/* LibTomCrypt, modular cryptographic library -- Tom St Denis
+ *
+ * LibTomCrypt is a library that provides various cryptographic
+ * algorithms in a highly modular and flexible manner.
+ *
+ * The library is free for all purposes without any express
+ * guarantee it works.
+ */
+
+/*
+ * This LTC implementation was adapted from:
+ *    http://www.ecrypt.eu.org/stream/e2-sosemanuk.html
+ */
+
+/*
+ * SOSEMANUK reference implementation.
+ *
+ * This code is supposed to run on any conforming C implementation (C90
+ * or later).
+ *
+ * (c) 2005 X-CRYPT project. This software is provided 'as-is', without
+ * any express or implied warranty. In no event will the authors be held
+ * liable for any damages arising from the use of this software.
+ *
+ * Permission is granted to anyone to use this software for any purpose,
+ * including commercial applications, and to alter it and redistribute it
+ * freely, subject to no restriction.
+ *
+ * Technical remarks and questions can be addressed to
+ * <thomas.pornin@cryptolog.com>
+ */
+
+#include "tomcrypt.h"
+
+#ifdef LTC_SOSEMANUK
+
+/* ======================================================================== */
+
+/*
+ * We want (and sometimes need) to perform explicit truncations to 32 bits.
+ */
+#define T32(x)   ((x) & (ulong32)0xFFFFFFFF)
+
+/*
+ * Some of our functions will be tagged as "inline" to help the compiler
+ * optimize things. We use "inline" only if the compiler is advanced
+ * enough to understand it; C99 compilers, and pre-C99 versions of gcc,
+ * understand enough "inline" for our purposes.
+ */
+
+/* ======================================================================== */
+
+/*
+ * Serpent S-boxes, implemented in bitslice mode. These circuits have
+ * been published by Dag Arne Osvik ("Speeding up Serpent", published in
+ * the 3rd AES Candidate Conference) and work on five 32-bit registers:
+ * the four inputs, and a fifth scratch register. There are meant to be
+ * quite fast on Pentium-class processors. These are not the fastest
+ * published, but they are "fast enough" and they are unencumbered as
+ * far as intellectual property is concerned (note: these are rewritten
+ * from the article itself, and hence are not covered by the GPL on
+ * Dag's code, which was not used here).
+ *
+ * The output bits are permuted. Here is the correspondance:
+ *   S0:  1420
+ *   S1:  2031
+ *   S2:  2314
+ *   S3:  1234
+ *   S4:  1403
+ *   S5:  1302
+ *   S6:  0142
+ *   S7:  4310
+ * (for instance, the output of S0 is in "r1, r4, r2, r0").
+ */
+
+#define S0(r0, r1, r2, r3, r4)   do { \
+        r3 ^= r0;  r4  = r1; \
+        r1 &= r3;  r4 ^= r2; \
+        r1 ^= r0;  r0 |= r3; \
+        r0 ^= r4;  r4 ^= r3; \
+        r3 ^= r2;  r2 |= r1; \
+        r2 ^= r4;  r4 = ~r4; \
+        r4 |= r1;  r1 ^= r3; \
+        r1 ^= r4;  r3 |= r0; \
+        r1 ^= r3;  r4 ^= r3; \
+    } while (0)
+
+#define S1(r0, r1, r2, r3, r4)   do { \
+        r0 = ~r0;  r2 = ~r2; \
+        r4  = r0;  r0 &= r1; \
+        r2 ^= r0;  r0 |= r3; \
+        r3 ^= r2;  r1 ^= r0; \
+        r0 ^= r4;  r4 |= r1; \
+        r1 ^= r3;  r2 |= r0; \
+        r2 &= r4;  r0 ^= r1; \
+        r1 &= r2; \
+        r1 ^= r0;  r0 &= r2; \
+        r0 ^= r4; \
+    } while (0)
+
+#define S2(r0, r1, r2, r3, r4)   do { \
+        r4  = r0;  r0 &= r2; \
+        r0 ^= r3;  r2 ^= r1; \
+        r2 ^= r0;  r3 |= r4; \
+        r3 ^= r1;  r4 ^= r2; \
+        r1  = r3;  r3 |= r4; \
+        r3 ^= r0;  r0 &= r1; \
+        r4 ^= r0;  r1 ^= r3; \
+        r1 ^= r4;  r4 = ~r4; \
+    } while (0)
+
+#define S3(r0, r1, r2, r3, r4)   do { \
+        r4  = r0;  r0 |= r3; \
+        r3 ^= r1;  r1 &= r4; \
+        r4 ^= r2;  r2 ^= r3; \
+        r3 &= r0;  r4 |= r1; \
+        r3 ^= r4;  r0 ^= r1; \
+        r4 &= r0;  r1 ^= r3; \
+        r4 ^= r2;  r1 |= r0; \
+        r1 ^= r2;  r0 ^= r3; \
+        r2  = r1;  r1 |= r3; \
+        r1 ^= r0; \
+    } while (0)
+
+#define S4(r0, r1, r2, r3, r4)   do { \
+        r1 ^= r3;  r3 = ~r3; \
+        r2 ^= r3;  r3 ^= r0; \
+        r4  = r1;  r1 &= r3; \
+        r1 ^= r2;  r4 ^= r3; \
+        r0 ^= r4;  r2 &= r4; \
+        r2 ^= r0;  r0 &= r1; \
+        r3 ^= r0;  r4 |= r1; \
+        r4 ^= r0;  r0 |= r3; \
+        r0 ^= r2;  r2 &= r3; \
+        r0 = ~r0;  r4 ^= r2; \
+    } while (0)
+
+#define S5(r0, r1, r2, r3, r4)   do { \
+        r0 ^= r1;  r1 ^= r3; \
+        r3 = ~r3;  r4  = r1; \
+        r1 &= r0;  r2 ^= r3; \
+        r1 ^= r2;  r2 |= r4; \
+        r4 ^= r3;  r3 &= r1; \
+        r3 ^= r0;  r4 ^= r1; \
+        r4 ^= r2;  r2 ^= r0; \
+        r0 &= r3;  r2 = ~r2; \
+        r0 ^= r4;  r4 |= r3; \
+        r2 ^= r4; \
+    } while (0)
+
+#define S6(r0, r1, r2, r3, r4)   do { \
+        r2 = ~r2;  r4  = r3; \
+        r3 &= r0;  r0 ^= r4; \
+        r3 ^= r2;  r2 |= r4; \
+        r1 ^= r3;  r2 ^= r0; \
+        r0 |= r1;  r2 ^= r1; \
+        r4 ^= r0;  r0 |= r3; \
+        r0 ^= r2;  r4 ^= r3; \
+        r4 ^= r0;  r3 = ~r3; \
+        r2 &= r4; \
+        r2 ^= r3; \
+    } while (0)
+
+#define S7(r0, r1, r2, r3, r4)   do { \
+        r4  = r1;  r1 |= r2; \
+        r1 ^= r3;  r4 ^= r2; \
+        r2 ^= r1;  r3 |= r4; \
+        r3 &= r0;  r4 ^= r2; \
+        r3 ^= r1;  r1 |= r4; \
+        r1 ^= r0;  r0 |= r4; \
+        r0 ^= r2;  r1 ^= r4; \
+        r2 ^= r1;  r1 &= r0; \
+        r1 ^= r4;  r2 = ~r2; \
+        r2 |= r0; \
+        r4 ^= r2; \
+    } while (0)
+
+/*
+ * The Serpent linear transform.
+ */
+#define SERPENT_LT(x0, x1, x2, x3)  do { \
+        x0 = ROLc(x0, 13); \
+        x2 = ROLc(x2, 3); \
+        x1 = x1 ^ x0 ^ x2; \
+        x3 = x3 ^ x2 ^ T32(x0 << 3); \
+        x1 = ROLc(x1, 1); \
+        x3 = ROLc(x3, 7); \
+        x0 = x0 ^ x1 ^ x3; \
+        x2 = x2 ^ x3 ^ T32(x1 << 7); \
+        x0 = ROLc(x0, 5); \
+        x2 = ROLc(x2, 22); \
+    } while (0)
+
+/* ======================================================================== */
+
+/*
+ * Key schedule: initialize the key context structure with the provided
+ * secret key. The secret key is an array of 1 to 32 bytes.
+ * @param ss       The Sosemanuk state
+ * @param key      Key
+ * @param keylen   Length of key
+ * @return CRYPT_OK on success
+ */
+int sosemanuk_setup(sosemanuk_state *ss, unsigned char *key, unsigned long keylen)
+{
+    /*
+     * This key schedule is actually a truncated Serpent key schedule.
+     * The key-derived words (w_i) are computed within the eight
+     * local variables w0 to w7, which are reused again and again.
+     */
+
+#define SKS(S, o0, o1, o2, o3, d0, d1, d2, d3)   do { \
+        ulong32 r0, r1, r2, r3, r4; \
+        r0 = w ## o0; \
+        r1 = w ## o1; \
+        r2 = w ## o2; \
+        r3 = w ## o3; \
+        S(r0, r1, r2, r3, r4); \
+        ss->kc[i ++] = r ## d0; \
+        ss->kc[i ++] = r ## d1; \
+        ss->kc[i ++] = r ## d2; \
+        ss->kc[i ++] = r ## d3; \
+    } while (0)
+
+#define SKS0    SKS(S0, 4, 5, 6, 7, 1, 4, 2, 0)
+#define SKS1    SKS(S1, 0, 1, 2, 3, 2, 0, 3, 1)
+#define SKS2    SKS(S2, 4, 5, 6, 7, 2, 3, 1, 4)
+#define SKS3    SKS(S3, 0, 1, 2, 3, 1, 2, 3, 4)
+#define SKS4    SKS(S4, 4, 5, 6, 7, 1, 4, 0, 3)
+#define SKS5    SKS(S5, 0, 1, 2, 3, 1, 3, 0, 2)
+#define SKS6    SKS(S6, 4, 5, 6, 7, 0, 1, 4, 2)
+#define SKS7    SKS(S7, 0, 1, 2, 3, 4, 3, 1, 0)
+
+#define WUP(wi, wi5, wi3, wi1, cc)   do { \
+        ulong32 tt = (wi) ^ (wi5) ^ (wi3) \
+            ^ (wi1) ^ (0x9E3779B9 ^ (ulong32)(cc)); \
+        (wi) = ROLc(tt, 11); \
+    } while (0)
+
+#define WUP0(cc)   do { \
+        WUP(w0, w3, w5, w7, cc); \
+        WUP(w1, w4, w6, w0, cc + 1); \
+        WUP(w2, w5, w7, w1, cc + 2); \
+        WUP(w3, w6, w0, w2, cc + 3); \
+    } while (0)
+
+#define WUP1(cc)   do { \
+        WUP(w4, w7, w1, w3, cc); \
+        WUP(w5, w0, w2, w4, cc + 1); \
+        WUP(w6, w1, w3, w5, cc + 2); \
+        WUP(w7, w2, w4, w6, cc + 3); \
+    } while (0)
+
+    unsigned char wbuf[32];
+    ulong32 w0, w1, w2, w3, w4, w5, w6, w7;
+    int i = 0;
+
+   LTC_ARGCHK(ss  != NULL);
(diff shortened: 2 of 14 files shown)


src/stream/sosemanuk/sosemanuk.c
```diff
@@ -0,0 +1,989 @@
...
+
+/*
+ * Some of our functions will be tagged as "inline" to help the compiler
+ * optimize things. We use "inline" only if the compiler is advanced
+ * enough to understand it; C99 compilers, and pre-C99 versions of gcc,
+ * understand enough "inline" for our purposes.
+ */
+#if (!defined BROKEN_C99 && defined __STDC__ && __STDC_VERSION__ >= 199901L) \
+    || defined __GNUC__
+#define INLINE inline
```
karel-m: We already have `LTC_INLINE` in `tomcrypt_cfg.h` - any improvements should go there. IMO we should introduce `LTC_STATIC_INLINE` as some compilers are fine with `inline` but bot with `static inline`.

src/stream/sosemanuk/sosemanuk.c
```diff
@@ -0,0 +1,989 @@
...
+
+#define WUP1(cc)   do { \
+        WUP(w4, w7, w1, w3, cc); \
+        WUP(w5, w0, w2, w4, cc + 1); \
+        WUP(w6, w1, w3, w5, cc + 2); \
+        WUP(w7, w2, w4, w6, cc + 3); \
+    } while (0)
+
+    unsigned char wbuf[32];
+    register ulong32 w0, w1, w2, w3, w4, w5, w6, w7;
```
karel-m: I am for not using `register`; it is IMO compiler's job to optimize these things

src/stream/sosemanuk/sosemanuk.c
```diff
@@ -0,0 +1,989 @@
...
+     * the linear transformation for that last round.
+     */
+#define FSF(zc, S, i0, i1, i2, i3, i4, o0, o1, o2, o3)  do { \
+        KA(zc, r ## i0, r ## i1, r ## i2, r ## i3); \
+        S(r ## i0, r ## i1, r ## i2, r ## i3, r ## i4); \
+        SERPENT_LT(r ## o0, r ## o1, r ## o2, r ## o3); \
+        KA(zc + 4, r ## o0, r ## o1, r ## o2, r ## o3); \
+    } while (0)
+
+    register ulong32 r0, r1, r2, r3, r4;
```
karel-m: I am for not using `register` see above

src/stream/sosemanuk/sosemanuk.c
```diff
@@ -0,0 +1,989 @@
...
+    || defined __GNUC__
+#define INLINE inline
+#else
+#define INLINE
+#endif
+
+/*
+ * 32-bit data decoding, little endian.
+ */
+static INLINE ulong32 decode32le(unsigned char *data)
```
karel-m: I am for completely removing `decode32le` and replacing with `LOAD32L` (if somebody wants to optimize it then it should be done globally in `LOAD32L` for the whole ltc)

src/stream/sosemanuk/sosemanuk.c
```diff
@@ -0,0 +1,989 @@
...
+#endif
+
+    w0 = decode32le(wbuf);
+    w1 = decode32le(wbuf + 4);
+    w2 = decode32le(wbuf + 8);
+    w3 = decode32le(wbuf + 12);
+    w4 = decode32le(wbuf + 16);
+    w5 = decode32le(wbuf + 20);
+    w6 = decode32le(wbuf + 24);
+    w7 = decode32le(wbuf + 28);
```
karel-m: replace all `decode32le` >> `LOAD32L`

src/stream/sosemanuk/sosemanuk.c
```diff
@@ -0,0 +1,989 @@
...
+    }
+#endif
+
+    /*
+     * Decode IV into four 32-bit words (little-endian).
+     */
+    r0 = decode32le(ivtmp);
+    r1 = decode32le(ivtmp + 4);
+    r2 = decode32le(ivtmp + 8);
+    r3 = decode32le(ivtmp + 12);
```
karel-m: replace all `decode32le` >> `LOAD32L`

src/stream/sosemanuk/sosemanuk.c
```diff
@@ -0,0 +1,989 @@
...
+        | ((ulong32)data[1] << 8)
+        | ((ulong32)data[2] << 16)
+        | ((ulong32)data[3] << 24);
+#endif
+}
+
+/*
+ * 32-bit data encoding, little-endian.
+ */
+static INLINE void encode32le(unsigned char *dst, ulong32 val)
```
karel-m: I am for removing `encode32le` and replacing with `STORE32L`

src/stream/sosemanuk/sosemanuk.c
```diff
@@ -0,0 +1,989 @@
...
+#define OUTWORD_BASE   (rc->buf)
+
+#define SRD(S, x0, x1, x2, x3, ooff)   do { \
+        PSPIN(u0, u1, u2, u3); \
+        S(u0, u1, u2, u3, u4); \
+        PSPOUT(u ## x0, u ## x1, u ## x2, u ## x3); \
+        encode32le(OUTWORD_BASE + ooff, u ## x0 ^ v0); \
+        encode32le(OUTWORD_BASE + ooff + 4, u ## x1 ^ v1); \
+        encode32le(OUTWORD_BASE + ooff + 8, u ## x2 ^ v2); \
+        encode32le(OUTWORD_BASE + ooff + 12, u ## x3 ^ v3); \
```
karel-m: replace all `encode32le` >> `STORE32L`

src/stream/sosemanuk/sosemanuk.c
```diff
@@ -0,0 +1,989 @@
...
+    dst[1] = (val >>  8) & 0xFF;
+    dst[2] = (val >> 16) & 0xFF;
+    dst[3] = (val >> 24) & 0xFF;
+#endif
+}
+
+/*
+ * Left-rotation by n bits (0 < n < 32).
+ */
+#define ROTL(x, n)    (T32(((x) << (n)) | T32((x) >> (32 - (n)))))
```
karel-m: I suggest removing `ROTL` and using already existing `ROLc` instead.

src/stream/sosemanuk/sosemanuk.c
```diff
@@ -0,0 +1,989 @@
...
+        r2 |= r0; \
+        r4 ^= r2; \
+    } while (0)
+
+/*
+ * The Serpent linear transform.
+ */
+#define SERPENT_LT(x0, x1, x2, x3)  do { \
+        x0 = ROTL(x0, 13); \
+        x2 = ROTL(x2, 3); \
```
karel-m: `ROTL` >> `ROLc`

src/stream/sosemanuk/sosemanuk.c
```diff
@@ -0,0 +1,989 @@
...
+        x1 = x1 ^ x0 ^ x2; \
+        x3 = x3 ^ x2 ^ T32(x0 << 3); \
+        x1 = ROTL(x1, 1); \
+        x3 = ROTL(x3, 7); \
```
karel-m: `ROTL` >> `ROLc`

src/stream/sosemanuk/sosemanuk.c
```diff
@@ -0,0 +1,989 @@
...
+        x0 = x0 ^ x1 ^ x3; \
+        x2 = x2 ^ x3 ^ T32(x1 << 7); \
+        x0 = ROTL(x0, 5); \
+        x2 = ROTL(x2, 22); \
```
karel-m: `ROTL` >> `ROLc`

src/stream/sosemanuk/sosemanuk.c
```diff
@@ -0,0 +1,989 @@
...
+#define SKS3    SKS(S3, 0, 1, 2, 3, 1, 2, 3, 4)
+#define SKS4    SKS(S4, 4, 5, 6, 7, 1, 4, 0, 3)
+#define SKS5    SKS(S5, 0, 1, 2, 3, 1, 3, 0, 2)
+#define SKS6    SKS(S6, 4, 5, 6, 7, 0, 1, 4, 2)
+#define SKS7    SKS(S7, 0, 1, 2, 3, 4, 3, 1, 0)
+
+#define WUP(wi, wi5, wi3, wi1, cc)   do { \
+        ulong32 tt = (wi) ^ (wi5) ^ (wi3) \
+            ^ (wi1) ^ (0x9E3779B9 ^ (ulong32)(cc)); \
+        (wi) = ROTL(tt, 11); \
```
karel-m: `ROTL` >> `ROLc`

src/stream/sosemanuk/sosemanuk.c
```diff
@@ -0,0 +1,989 @@
...
+    /*
+     * FSM() updates the finite state machine.
+     */
+#define FSM(x0, x1, x2, x3, x4, x5, x6, x7, x8, x9)   do { \
+        ulong32 tt, or1; \
+        tt = XMUX(r1, s ## x1, s ## x8); \
+        or1 = r1; \
+        r1 = T32(r2 + tt); \
+        tt = T32(or1 * 0x54655307); \
+        r2 = ROTL(tt, 7); \
```
karel-m: `ROTL` >> `ROLc`

src/stream/sosemanuk/sosemanuk.c
```diff
@@ -0,0 +1,989 @@
...
+#undef SOSEMANUK_VECTOR
+#undef SOSEMANUK_SPEED
+
+/* ======================================================================== */
+
+/*
+ * We want (and sometimes need) to perform explicit truncations to 32 bits.
+ */
+#define ONE32    ((ulong32)0xFFFFFFFF)
+#define T32(x)   ((x) & ONE32)
```
karel-m: Looks like we can simply define just:
```C
#define T32(x)   ((x) & (ulong32)0xFFFFFFFF)
```

src/stream/sosemanuk/sosemanuk.c
```diff
@@ -0,0 +1,989 @@
...
+ *
+ * Technical remarks and questions can be addressed to
+ * <thomas.pornin@cryptolog.com>
+ */
+
+#include "tomcrypt.h"
+
+#ifdef LTC_SOSEMANUK
+
+#undef SOSEMANUK_VECTOR
```
karel-m: `SOSEMANUK_VECTOR` seems like a debug stuff which we have in ltc only in `<something>_test` functions

src/stream/sosemanuk/sosemanuk.c
```diff
@@ -0,0 +1,989 @@
...
+#undef SOSEMANUK_SPEED
```
karel-m: `SOSEMANUK_SPEED` is not used so it can be IMO removed

src/stream/sosemanuk/sosemanuk.c
```diff
@@ -0,0 +1,989 @@
...
+{
+   LTC_ARGCHK(kc != NULL);
+   LTC_ARGCHK(rc != NULL);
+   XMEMSET(kc, 0, sizeof(sosemanuk_key_context));
+   XMEMSET(rc, 0, sizeof(sosemanuk_run_context));
+   return CRYPT_OK;
+}
+
+
+#undef INLINE
```
karel-m: if we switch to `LTC_INLINE` this line is not needed

src/headers/tomcrypt_cipher.h
```diff
@@ -1031,6 +1031,37 @@ int salsa20_test(void);
...
+    ulong32 s00, s01, s02, s03, s04, s05, s06, s07, s08, s09;
+    ulong32 r1, r2;
+    /*
+     * Buffering: the stream cipher produces output data by
+     * blocks of 640 bits. buf[] contains such a block, and
+     * "ptr" is the index of the next output byte.
+     */
+    unsigned char buf[80];
+    unsigned ptr;
+} sosemanuk_run_context;
```
karel-m: Why not simply:
```C
typedef struct {
    ulong32 sk[100]; /* key context */
    ulong32 s00, s01, s02, s03, s04, s05, s06, s07, s08, s09;
    ulong32 r1, r2;
    /*
     * Buffering: the stream cipher produces output data by
     * blocks of 640 bits. buf[] contains such a block, and
     * "ptr" is the index of the next output byte.
     */
    unsigned char buf[80];
    unsigned ptr;
} sosemanuk_state;
```
?

src/headers/tomcrypt_cipher.h
```diff
@@ -1031,6 +1031,37 @@ int salsa20_test(void);
...
+
+int sosemanuk_schedule(sosemanuk_key_context *kc, unsigned char *key, size_t key_len);
```
karel-m: Wouldn't it be better (more consistent) to name this `sosemanuk_setup` instead of `sosemanuk_schedule`?

src/headers/tomcrypt_cipher.h
```diff
@@ -1031,6 +1031,37 @@ int salsa20_test(void);
...
+int sosemanuk_init(sosemanuk_run_context *rc, sosemanuk_key_context *kc, unsigned char *iv, size_t iv_len);
+int sosemanuk_crypt(sosemanuk_run_context *rc, const unsigned char *in, size_t data_len, unsigned char *out);
+int sosemanuk_keystream(sosemanuk_run_context *rc, unsigned char *out, size_t out_len);
```
karel-m: General comment to`size_t`: I think we should consistently use `unsigned long` (which is used for length at nearly all places in ltc) instead of `size_t`, although the `size_t` is more correct. There is a separate ticket for turning `unsigned long` into `size_t` and we should do it in all places at once (after we discuss the new troubles it might mean especially on 64bit MS Windows where `unsigned long` and `size_t` and 32-bit vs. 64-bit).

karel-m:
Larry, here is my re-definition of `LTC_INLINE` which should work also with your code:
```C
#if defined(__GNUC__) || defined(__xlc__)
   #define LTC_INLINE __inline__
#elif defined(_MSC_VER) || defined(__HP_cc)
   #define LTC_INLINE __inline
#elif defined(__STDC_VERSION__) && __STDC_VERSION__ >= 199901L
   #define LTC_INLINE inline
#else
   #define LTC_INLINE
#endif
```
It should replace current `LTC_INLINE` in `tomcrypt_cfg.h`.

Tested with: gcc (v 3.x-7.x, also with -std=c90), clang, MSVC, IBM xlc (AIX), HP cc (HP-UX).

src/stream/sosemanuk/sosemanuk.c
```diff
@@ -715,13 +585,8 @@ static ulong32 mul_ia[] = {
...
  * Compute the next block of bits of output stream. This is equivalent
  * to one full rotation of the shift register.
- *
- * If SOSEMANUK_SPEED is defined, this function takes an extra parameter
- * "counter". The function then returns the sum of all produced
- * 32-bit words, in an "ulong32". That sum prevents the compiler from
- * optimizing out part of the computation.
  */
-static INLINE void sosemanuk_internal(sosemanuk_run_context *rc)
+static LTC_INLINE void sosemanuk_internal(sosemanuk_state *ss)
```
sjaeckel: This should be called `_sosemanuk_internal` ( @karel-m it looks like `helper.pl` doesn't detect that because of the additional keyword)

src/stream/sosemanuk/sosemanuk.c
```diff
@@ -879,8 +724,8 @@ static INLINE void sosemanuk_internal(sosemanuk_run_context *rc)
  * or in2[] is not allowed. Total overlap (out == in1 and/or out == in2)
  * is allowed.
  */
-static INLINE void xorbuf(const unsigned char *in1, const unsigned char *in2,
-    unsigned char *out, size_t data_len)
+static LTC_INLINE void xorbuf(const unsigned char *in1, const unsigned char *in2,
```
sjaeckel: same

src/stream/sosemanuk/sosemanuk.c
karel-m: I have fixed helper.pl (in develop)

src/stream/sosemanuk/sosemanuk.c
```diff
@@ -0,0 +1,830 @@
...
+        WUP(w5, w0, w2, w4, cc + 1); \
+        WUP(w6, w1, w3, w5, cc + 2); \
+        WUP(w7, w2, w4, w6, cc + 3); \
+    } while (0)
+
+    unsigned char wbuf[32];
+    ulong32 w0, w1, w2, w3, w4, w5, w6, w7;
+    int i = 0;
+
+    /*
```
karel-m:
```C
   LTC_ARGCHK(ss  != NULL);
   LTC_ARGCHK(key != NULL);
```

src/stream/sosemanuk/sosemanuk.c
```diff
@@ -0,0 +1,830 @@
...
+#define FSF(zc, S, i0, i1, i2, i3, i4, o0, o1, o2, o3)  do { \
+        KA(zc, r ## i0, r ## i1, r ## i2, r ## i3); \
+        S(r ## i0, r ## i1, r ## i2, r ## i3, r ## i4); \
+        SERPENT_LT(r ## o0, r ## o1, r ## o2, r ## o3); \
+        KA(zc + 4, r ## o0, r ## o1, r ## o2, r ## o3); \
+    } while (0)
+
+    ulong32 r0, r1, r2, r3, r4;
+    unsigned char ivtmp[16];
+
```
karel-m:
```C
   LTC_ARGCHK(ss != NULL);
   LTC_ARGCHK(iv != NULL);
   LTC_ARGCHK(iv_len > 0); // or not?
```

src/stream/sosemanuk/sosemanuk.c
```diff
@@ -0,0 +1,830 @@
...
+ * @param rc       The Sosemanuk run context
+ * @param in       Data in
+ * @param out      Data out
+ * @param data_len Length of data
+ * @return CRYPT_OK on success
+ */
+int sosemanuk_crypt(sosemanuk_state *ss,
+                        const unsigned char *in, unsigned long data_len, unsigned char *out)
+{
+    if (ss->ptr < (sizeof ss->buf)) {
```
karel-m:
```C
   LTC_ARGCHK(ss  != NULL);
   LTC_ARGCHK(in  != NULL);
   LTC_ARGCHK(out != NULL);
```

src/stream/sosemanuk/sosemanuk.c
```diff
@@ -0,0 +1,830 @@
...
+/*
+ * Cipher operation, as a PRNG: the provided output buffer is filled with
+ * pseudo-random bytes as output from the stream cipher.
+ * @param rc       The Sosemanuk run context
+ * @param out      Data out
+ * @param out_len  Length of output
+ * @return CRYPT_OK on success
+ */
+int sosemanuk_keystream(sosemanuk_state *ss, unsigned char *out, unsigned long out_len)
+{
```
karel-m: Why do not we do the same stuff as by other stream ciphers?
```C
   if (outlen == 0) return CRYPT_OK; /* nothing to do */
   LTC_ARGCHK(out != NULL);
   XMEMSET(out, 0, outlen);
   return sosenamuk_crypt(sa, out, outlen, out);
```

src/stream/sosemanuk/sosemanuk.c
sjaeckel: * `LTC_ARGCHK(iv_len > 0);` seems not to be required
* I'd write this a bit easier

```
unsigned char ivtmp[16] = {0};

if (iv_len > 0) {
   iv_len = iv_len > 16 ? 16 : iv_len;
   XMEMCPY(ivtmp, iv, iv_len);
}
```

src/stream/sosemanuk/sosemanuk.c
```diff
@@ -0,0 +1,830 @@
...
+    FSS(80, S4, 0, 1, 4, 2, 3, 1, 3, 0, 2);
+    FSS(84, S5, 1, 3, 0, 2, 4, 3, 2, 1, 0);
+    FSS(88, S6, 3, 2, 1, 0, 4, 3, 2, 4, 1);
+    FSF(92, S7, 3, 2, 4, 1, 0, 0, 1, 2, 3);
+    ss->s03 = r0;
+    ss->s02 = r1;
+    ss->s01 = r2;
+    ss->s00 = r3;
+
+    ss->ptr = sizeof ss->buf;
```
sjaeckel: I know there are some reasons why someone would want `sizeof` to not look like a function, but it's all-over like that, so please an you change all `sizeof foo` into `sizeof(foo)`

@karel-m would it probably make sense to add a check of that to helper.pl?

karel-m: detection added to develop as b7e35e9f

src/stream/sosemanuk/sosemanuk.c
karel-m: If `IV` is not required we can either allow "not calling" `sosemanuk_setiv` or we have to check in `sosemanuk_crypt` that `sosemanuk_setiv`  was called.

buggywhip:
> On 5Nov, 2017, at 10:49 AM, karel-m <notifications@github.com> wrote:
> 
> If IV is not required we can either allow "not calling" sosemanuk_setiv or we have to check in sosemanuk_crypt that sosemanuk_setiv was called.
> 

Instead, consider...

I defined Sosemanuk in my Python wrapper as a class so this is easy, but the following behavior can be emulated in C with a flag in the context/state.

I allow you to pass in the key and the IV during setup(), actually, when you instantiate an instance.  If you fail to pass an IV, I default to 0.  As the final step in setup(), I call the method setiv().  If you want to call setiv() separately you may.  

If you want to reuse the key/instance for a subsequent crypt you should call setiv().  If you call setiv() without an argument, again I default to 0.

I'd like to suggest such for LTC and be applied to all the streams.  ...and perhaps to the blocks later.  ...with RC4 being a special case.

karel-m:
I think that the following is a quite simple approach similar to what we do in `salsa20_*`:
* in `sosemanuk_setup` set `ss->ptr = NULL`
* the proper value of `ss->ptr` is set during `sosemanuk_setiv`
* in `sosemanuk_crypt` check  `LTC_ARGCHK(ss->ptr != NULL)`

As you probably know I am not convinced that a unified API for stream ciphers will give us some benefits. You are trying to unify API for ciphers - without IV, with IV, with IV+counter, some have rounds parameter etc. - which might make sense on higher level (like your python bindings or my perl bindings) but it can still be quite easily implemented with our current ltc interface.

buggywhip:
> On 5Nov, 2017, at 3:21 AM, karel-m <notifications@github.com> wrote:
> 
> @karel-m commented on this pull request.
> 
> In src/stream/sosemanuk/sosemanuk.c <https://github.com/libtom/libtomcrypt/pull/322#discussion_r148954573>:
> 
> > +    } while (0)
> +
> +    /*
> +     * Last Serpent round. Contrary to the "true" Serpent, we keep
> +     * the linear transformation for that last round.
> +     */
> +#define FSF(zc, S, i0, i1, i2, i3, i4, o0, o1, o2, o3)  do { \
> +        KA(zc, r ## i0, r ## i1, r ## i2, r ## i3); \
> +        S(r ## i0, r ## i1, r ## i2, r ## i3, r ## i4); \
> +        SERPENT_LT(r ## o0, r ## o1, r ## o2, r ## o3); \
> +        KA(zc + 4, r ## o0, r ## o1, r ## o2, r ## o3); \
> +    } while (0)
> +
> +    ulong32 r0, r1, r2, r3, r4;
> +    unsigned char ivtmp[16];
> +
>    LTC_ARGCHK(ss != NULL);
>    LTC_ARGCHK(iv != NULL);
>    LTC_ARGCHK(iv_len > 0); // or not?
The Sosemanuk spec calls for an IV of 128 bits; the spec is silent on other sizes.  The eSTREAM reference implementation truncates the IV if longer than 16 bytes and pads with nulls out to 16.  

I can accept padding to 16 if something short was submitted but am uncomfortable with truncating.  It could be the tip of a larger problem.  Thus I submit...

    LTC_ARGCHK(ss != NULL);
    LTC_ARGCHK(iv != NULL);
    LTC_ARGCHK(iv_len > 0 && iv_len <= 16);

and 

    if (iv_len >= sizeof(ivtmp)) {
        XMEMCPY(ivtmp, iv, sizeof(ivtmp));
    } else {
        if (iv_len > 0)
            XMEMCPY(ivtmp, iv, iv_len);
        XMEMSET(ivtmp + iv_len, 0, (sizeof(ivtmp)) - iv_len);
    }

becomes simply

    XMEMCPY(ivtmp, iv, iv_len);
    XMEMSET(ivtmp + iv_len, 0, (sizeof(ivtmp)) - iv_len);

src/stream/sosemanuk/sosemanuk.c
```diff
@@ -0,0 +1,816 @@
...
+        S(r ## i0, r ## i1, r ## i2, r ## i3, r ## i4); \
+        SERPENT_LT(r ## o0, r ## o1, r ## o2, r ## o3); \
+        KA(zc + 4, r ## o0, r ## o1, r ## o2, r ## o3); \
+    } while (0)
+
+    ulong32 r0, r1, r2, r3, r4;
+    unsigned char ivtmp[16] = {0};
+
+    LTC_ARGCHK(ss != NULL);
+    LTC_ARGCHK(iv != NULL);
```
karel-m: If zero-length IV is acceptable we should perhaps also accept: `sosemanuk_setiv(&state, NULL, 0)` so perhaps something like `LTC_ARGCHK(iv != NULL || ivlen ==0);`

src/stream/sosemanuk/sosemanuk.c
```diff
@@ -0,0 +1,816 @@
...
+    LTC_ARGCHK(ivlen >= 0 && ivlen <= 16);
+
+    XMEMCPY(ivtmp, iv, ivlen);
```
karel-m: I'd say: `if (ivlen > 0) XMEMCPY(ivtmp, iv, ivlen);`

src/stream/sosemanuk/sosemanuk.c
```diff
@@ -0,0 +1,816 @@
...
+ * @param datalen  Length of data
+ * @return CRYPT_OK on success
+ */
+int sosemanuk_crypt(sosemanuk_state *ss,
+                        const unsigned char *in, unsigned long datalen, unsigned char *out)
+{
+
+    LTC_ARGCHK(ss  != NULL);
+    LTC_ARGCHK(in  != NULL);
+    LTC_ARGCHK(out != NULL);
```
karel-m: `LTC_ARGCHK(ss->ptr != 666); /* check whether sosemanuk_setiv was called */`

src/stream/sosemanuk/sosemanuk.c
```diff
@@ -0,0 +1,824 @@
...
+        SERPENT_LT(r ## o0, r ## o1, r ## o2, r ## o3); \
+        KA(zc + 4, r ## o0, r ## o1, r ## o2, r ## o3); \
+    } while (0)
+
+    ulong32 r0, r1, r2, r3, r4;
+    unsigned char ivtmp[16] = {0};
+
+    LTC_ARGCHK(ss != NULL);
+    LTC_ARGCHK(ivlen >= 0 && ivlen <= 16);
+    LTC_ARGCHK((iv != NULL && ivlen > 0) || (iv == NULL && ivlen == 0));
```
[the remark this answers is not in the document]
sjaeckel: :+1: `LTC_ARGCHK(iv != NULL || ivlen == 0);`

src/stream/sosemanuk/sosemanuk.c
```diff
@@ -0,0 +1,824 @@
...
+        S(r ## i0, r ## i1, r ## i2, r ## i3, r ## i4); \
+        SERPENT_LT(r ## o0, r ## o1, r ## o2, r ## o3); \
+        KA(zc + 4, r ## o0, r ## o1, r ## o2, r ## o3); \
+    } while (0)
+
+    ulong32 r0, r1, r2, r3, r4;
+    unsigned char ivtmp[16] = {0};
+
+    LTC_ARGCHK(ss != NULL);
+    LTC_ARGCHK(ivlen >= 0 && ivlen <= 16);
```
sjaeckel: as we check `ivlen` in the next statement this should be enough here: `LTC_ARGCHK(ivlen <= 16);`

probably then also swap these two lines

PR #326, opened 2017-11-09
fixes #323 ccm_process fails to process input buffer longer than 256

### LINKED ISSUE #323: ccm_process fails to process input buffer longer than 256 bytes
### Prerequisites

* [ ] Checked the developer manual
* [ ] Checked that your issue isn't already filed: https://github.com/issues?utf8=✓&q=repo%3Alibtom%2Flibtomcrypt
* [ ] Checked that your issue isn't related to TomsFastMath's limitation that PK operations can by default only be done with max. 2048bit keys

### Description

If you call `ccm_process` with `ptlen > 256` it loops for ever at this line https://github.com/libtom/libtomcrypt/blob/develop/src/encauth/ccm/ccm_process.c#L49 because `y` is just `unsigned char` and should be `unsigned long`

PR later

### Version

latest develop

karel-m:
see #323

### DIFF
--- a/src/encauth/ccm/ccm_process.c
+++ b/src/encauth/ccm/ccm_process.c
@@ -24,7 +24,8 @@ int ccm_process(ccm_state *ccm,
                 unsigned char *ct,
                 int direction)
 {
-   unsigned char  y, z, b;
+   unsigned char z, b;
+   unsigned long y;
    int err;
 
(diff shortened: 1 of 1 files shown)


sjaeckel:
ping @jforissier

jforissier:
@sjaeckel thanks for the heads up!

PR #327, opened 2017-11-10
Fix ccm_memory() cleaning user-supplied key

sjaeckel:

This fixes a segfault caused by `ccm_memory()` with `LTC_CLEAN_STACK` enabled.

### DIFF
--- a/src/encauth/ccm/ccm_memory.c
+++ b/src/encauth/ccm/ccm_memory.c
@@ -333,6 +333,9 @@ int ccm_memory(int cipher,
 
    if (skey != uskey) {
       cipher_descriptor[cipher].done(skey);
+#ifdef LTC_CLEAN_STACK
+      zeromem(skey,   sizeof(*skey));
+#endif
    }
 
    if (direction == CCM_ENCRYPT) {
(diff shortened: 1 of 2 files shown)


src/encauth/ccm/ccm_memory.c
```diff
@@ -380,7 +380,9 @@ int ccm_memory(int cipher,
    fastMask = 0;
 #endif
    mask = 0;
-   zeromem(skey,   sizeof(*skey));
+   if (skey != uskey) {
+      zeromem(skey,   sizeof(*skey));
```
karel-m: There is already `if (skey != uskey)` couple of lines above (line 334).

Should not it be better to move `zeromem(skey,   sizeof(*skey))` there?

sjaeckel: so you mean like this?
```
    cipher_descriptor[cipher].done(skey);
#ifdef LTC_CLEAN_STACK
    zeromem(skey,   sizeof(*skey));
#endif
```
would be fine for me

karel-m: Yes, that's it.

PR #329, opened 2017-11-26
Correct SHA3 block sizes

karel-m:
### Checklist


sjaeckel:
As this PR would lead (again) to a bump of the major ABI version I thought about investing some time to implement a work-around or at least catch the corner cases so the ABI version could be kept the same.
As I'm not sure if that's worth the effort, I thought probably one of you guys @rofl0r @stapelberg @sebastianas or someone else who reads this can give their opinion resp. technical reasoning which way we should go...
Thanks in advance!

stapelberg:
As per https://codesearch.debian.net/search?q=sha3_%5B0-9%5D%2B_desc&perpkg=1, no software in Debian seems to use sha3_[0-9]+_desc, so this should be an easy transition.

karel-m:
Looking at my own code I am in doubts whether increasing MAXBLOCKSIZE is really necessary.

sjaeckel:
> Looking at my own code I am in doubts whether increasing MAXBLOCKSIZE is really necessary.
> View changes

the ABI bump is necessary because of the changed size of `hmac_state` (and all the other typedefs which contain `MAXBLOCKSIZE`) therefore your hack unfortunately doesn't really help...



> As per https://codesearch.debian.net/search?q=sha3_%5B0-9%5D%2B_desc&perpkg=1, no software in Debian seems to use sha3_[0-9]+_desc, so this should be an easy transition.

thanks for the info!

PR #330, opened 2017-11-28
Fix docs build in Sosemanuk section.

J08nY:
The latex build of `doc/crypt.tex` is currently broken by unescaped `_` characters in the Sosemanuk section of it, this fixes the docs.


sjaeckel:
We should really build the docs in CI...

PR #334, opened 2017-12-06
Add Rabbit stream cipher

buggywhip:
### Checklist

### DIFF
--- a/doc/crypt.tex
+++ b/doc/crypt.tex
@@ -1391,6 +1391,65 @@ To do multiple encryptions and decryptions with the same key, you will want to s
 you do not need to re-run \textit{sosemanuk\_setup()} again, unless of course, you called
 \textit{sosemanuk\_done()}.
 
+\mysection{Rabbit}
+
+\textit{Rabbit}, along with Salsa20, Sosemanuk, and HC-128, was named one of the winners
+in the EU eSTREAM competition.  Rabbit is a simple and fast cipher suitable for resource
+constrained platforms.
+
+This implementation of Rabbit will accept a key of up to 16 bytes (128 bits), and an
+initialization vector of up to 8 bytes (64 bits).  If the provided key or IV is shorter,
+they will be null padded to that length.  There is no rounds option.
+
+The calls to \textit{rabbit\_crypt()} or \textit{rabbit\_keystream()} may be for any
+number bytes although their generation will be more efficient if the requests are for
+multiples of 16 bytes.
+
+You begin initializing Rabbit by creating a state and calling \textit{rabbit\_setup()}
+with a 128-bit key.
+\begin{verbatim}
+rabbit_state st;
+err = rabbit_setup(&st, key, keylen);
+\end{verbatim}
+
+Set the IV using \textit{rabbit\_setiv() with a 64-bit IV.
+\begin{verbatim}
+err = rabbit_setiv(&st, iv, iv_len);
+\end{verbatim}
+
+For the actual encryption or decryption, call:
+\begin{verbatim}
+err = rabbit_crypt(&st, in_buffer, inlen, out_buffer);
+\end{verbatim}
+
+If you just want a random stream of bytes initialize the cipher with a truly random \textit{key}
+(16 bytes), a truly random \textit{iv} (8 bytes). After that you can
+get a stream of pseudo--random bytes via:
+\begin{verbatim}
+err = rabbit_keystream(&st, out_buffer, length);
+\end{verbatim}
+
+When finished you should wipe the key and the state:
+\begin{verbatim}
+err = rabbit_done(&st);
+\end{verbatim}
+
+Technically, it is possible to initialize Rabbit with only a call to \textit{rabbit\_setup()},
+skipping the call to \textit{rabbit\_setiv()}.  This is not recommended for interoperability reasons,
+but there may be special cases like closed systems or custom protocols where doing so might be
+appropriate.  Please note that skipping \textit{rabbit\_setiv()} will give results different than
+calling \textit{rabbit\_setiv()} passing a zero-valued IV.  You should call rabbit_setiv() unless
+you have a special reason for not doing so.
+
+To do multiple encryptions and decryptions with the same key, you can reset the algorithm
+using \textit{rabbit\_setiv()} if you saved the state and did not wipe it with \textit{rabbit\_done()}.
+You will want to use a different IV but you do not need to call \textit{rabbit\_setup()} a 2nd time,
+unless of course, you skipped calling \textit{rabbit\_setiv()}.
+
+For more information, see: \newline
+\hspace{4em}- \url{http://www.ecrypt.eu.org/stream/p3ciphers/rabbit/rabbit_p3.pdf} \newline
+\hspace{4em}- \url{https://tools.ietf.org/html/rfc4503}
+
 \mysection{RC4}
 
 For more information about RC4 see \url{https://en.wikipedia.org/wiki/RC4}.
@@ -6568,8 +6627,9 @@ This ensures that \textit{N} is set to a random MPI in the range $1 \le N < limi
 
 \subsection{Zero'ing data}
 
-As widely know optimizing-compilers are sometimes allowed to remove an invocation of \textit{memset(out, 0, outlen)}, which could result
-in sensitive data beeing not zero'ed out. Therefore LibTomCrypt implements a variant of this routine which won't be optimized-away.
+Optimizing compilers are sometimes allowed to remove an invocation of \textit{memset(out, 0, outlen)},
+which could result in sensitive data not being zero'ed out. LibTomCrypt, therefore, implements a variant
+of this routine, \textit{zeromem(out, outlen)}, which won't be optimized away.
 
 \index{zeromem()}
 \begin{verbatim}
--- a/makefile.mingw
+++ b/makefile.mingw
@@ -190,12 +190,12 @@ src/prngs/rc4.o src/prngs/rng_get_bytes.o src/prngs/rng_make_prng.o src/prngs/so
 src/prngs/sprng.o src/prngs/yarrow.o src/stream/chacha/chacha_crypt.o src/stream/chacha/chacha_done.o \
 src/stream/chacha/chacha_ivctr32.o src/stream/chacha/chacha_ivctr64.o \
 src/stream/chacha/chacha_keystream.o src/stream/chacha/chacha_setup.o src/stream/chacha/chacha_test.o \
-src/stream/rc4/rc4_stream.o src/stream/rc4/rc4_test.o src/stream/salsa20/salsa20_crypt.o \
-src/stream/salsa20/salsa20_done.o src/stream/salsa20/salsa20_ivctr64.o \
-src/stream/salsa20/salsa20_keystream.o src/stream/salsa20/salsa20_setup.o \
-src/stream/salsa20/salsa20_test.o src/stream/sober128/sober128_stream.o \
-src/stream/sober128/sober128_test.o src/stream/sosemanuk/sosemanuk.o \
-src/stream/sosemanuk/sosemanuk_test.o
+src/stream/rabbit/rabbit.o src/stream/rc4/rc4_stream.o src/stream/rc4/rc4_test.o \
+src/stream/salsa20/salsa20_crypt.o src/stream/salsa20/salsa20_done.o \
+src/stream/salsa20/salsa20_ivctr64.o src/stream/salsa20/salsa20_keystream.o \
+src/stream/salsa20/salsa20_setup.o src/stream/salsa20/salsa20_test.o \
+src/stream/sober128/sober128_stream.o src/stream/sober128/sober128_test.o \
+src/stream/sosemanuk/sosemanuk.o src/stream/sosemanuk/sosemanuk_test.o
 
 #List of test objects to compile
 TOBJECTS=tests/base32_test.o tests/base64_test.o tests/cipher_hash_test.o tests/common.o \
--- a/makefile.msvc
+++ b/makefile.msvc
@@ -183,12 +183,12 @@ src/prngs/rc4.obj src/prngs/rng_get_bytes.obj src/prngs/rng_make_prng.obj src/pr
 src/prngs/sprng.obj src/prngs/yarrow.obj src/stream/chacha/chacha_crypt.obj src/stream/chacha/chacha_done.obj \
 src/stream/chacha/chacha_ivctr32.obj src/stream/chacha/chacha_ivctr64.obj \
 src/stream/chacha/chacha_keystream.obj src/stream/chacha/chacha_setup.obj src/stream/chacha/chacha_test.obj \
-src/stream/rc4/rc4_stream.obj src/stream/rc4/rc4_test.obj src/stream/salsa20/salsa20_crypt.obj \
-src/stream/salsa20/salsa20_done.obj src/stream/salsa20/salsa20_ivctr64.obj \
-src/stream/salsa20/salsa20_keystream.obj src/stream/salsa20/salsa20_setup.obj \
-src/stream/salsa20/salsa20_test.obj src/stream/sober128/sober128_stream.obj \
-src/stream/sober128/sober128_test.obj src/stream/sosemanuk/sosemanuk.obj \
-src/stream/sosemanuk/sosemanuk_test.obj
+src/stream/rabbit/rabbit.obj src/stream/rc4/rc4_stream.obj src/stream/rc4/rc4_test.obj \
+src/stream/salsa20/salsa20_crypt.obj src/stream/salsa20/salsa20_done.obj \
+src/stream/salsa20/salsa20_ivctr64.obj src/stream/salsa20/salsa20_keystream.obj \
+src/stream/salsa20/salsa20_setup.obj src/stream/salsa20/salsa20_test.obj \
+src/stream/sober128/sober128_stream.obj src/stream/sober128/sober128_test.obj \
+src/stream/sosemanuk/sosemanuk.obj src/stream/sosemanuk/sosemanuk_test.obj
 
 #List of test objects to compile
 TOBJECTS=tests/base32_test.obj tests/base64_test.obj tests/cipher_hash_test.obj tests/common.obj \
--- a/src/headers/tomcrypt_cipher.h
+++ b/src/headers/tomcrypt_cipher.h
@@ -1031,8 +1031,6 @@ int salsa20_test(void);
 
 #endif /* LTC_SALSA20 */
 
-
-
 #ifdef LTC_SOSEMANUK
 
 typedef struct {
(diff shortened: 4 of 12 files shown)


karel-m:
There is no gain in starting using `uint32_t` instead of `ulong32` and `uint8_t` instead of `unsigned char`. 

This is what I wrote half a year ago (something similar I have also written off-github to you).

**I am extremely strongly against creating dependency on `stdint.h`.** (unless there is an option to define something like `-DLTC_WITHOUT_STDINT_H`)

I lost this battle in libtommath and the result is that I have to keep a specially patched `tommath.h` for my perl bindings which is a pain for me. Frankly we did not gain anything in libtommath by starting using `stdint.h`.

And please do not give me those stories about how old is C99, on my Visual Studio 2008 there is no `stdint.h` so it only forces me to keep my stdint.h-less fork of libtomcrypt (as I already have to do with libtommath).

In short: either `me` or `stdint.h` 😃

buggywhip:
Funny, earlier this morning I was thinking the 'add Rabbit' PR was overloaded with <stdint.h> issues that might be better served by a separate PR.
> I have to keep a specially patched tommath.h for my perl bindings which is a pain for me. 
> 
Interesting, I didn't realize and it confirms.  Thanks.


I propose changing Rabbit back to ulong32 as doing so would NOT cause an immediate problem.  It has seemingly worked well for some time now.  Further, I propose a new PR to discuss <stdint.h> issues and how best to address them.  ..or not.  

I contend it is far more important to add Rabbit than it is to hang <stdint.h> on this poor bunny's tail.  ;-)


.

makefile.mingw
```diff
@@ -190,12 +190,12 @@ src/prngs/rc4.o src/prngs/rng_get_bytes.o src/prngs/rng_make_prng.o src/prngs/so
 src/prngs/sprng.o src/prngs/yarrow.o src/stream/chacha/chacha_crypt.o src/stream/chacha/chacha_done.o \
 src/stream/chacha/chacha_ivctr32.o src/stream/chacha/chacha_ivctr64.o \
 src/stream/chacha/chacha_keystream.o src/stream/chacha/chacha_setup.o src/stream/chacha/chacha_test.o \
-src/stream/rc4/rc4_stream.o src/stream/rc4/rc4_test.o src/stream/salsa20/salsa20_crypt.o \
-src/stream/salsa20/salsa20_done.o src/stream/salsa20/salsa20_ivctr64.o \
-src/stream/salsa20/salsa20_keystream.o src/stream/salsa20/salsa20_setup.o \
-src/stream/salsa20/salsa20_test.o src/stream/sober128/sober128_stream.o \
-src/stream/sober128/sober128_test.o src/stream/sosemanuk/sosemanuk.o \
-src/stream/sosemanuk/sosemanuk_test.o
+src/stream/rabbit/rabbit.o src/stream/rc4/rc4_stream.o src/stream/rc4/rc4_test.o \
```
rofl0r: wouldnt it make more sense to add the only new file in an own line, instead of juggling all the items in 6 different lines around ? that makes it hard to see what was actually changed

makefile.msvc
```diff
@@ -183,12 +183,12 @@ src/prngs/rc4.obj src/prngs/rng_get_bytes.obj src/prngs/rng_make_prng.obj src/pr
 src/prngs/sprng.obj src/prngs/yarrow.obj src/stream/chacha/chacha_crypt.obj src/stream/chacha/chacha_done.obj \
 src/stream/chacha/chacha_ivctr32.obj src/stream/chacha/chacha_ivctr64.obj \
 src/stream/chacha/chacha_keystream.obj src/stream/chacha/chacha_setup.obj src/stream/chacha/chacha_test.obj \
-src/stream/rc4/rc4_stream.obj src/stream/rc4/rc4_test.obj src/stream/salsa20/salsa20_crypt.obj \
-src/stream/salsa20/salsa20_done.obj src/stream/salsa20/salsa20_ivctr64.obj \
-src/stream/salsa20/salsa20_keystream.obj src/stream/salsa20/salsa20_setup.obj \
-src/stream/salsa20/salsa20_test.obj src/stream/sober128/sober128_stream.obj \
-src/stream/sober128/sober128_test.obj src/stream/sosemanuk/sosemanuk.obj \
-src/stream/sosemanuk/sosemanuk_test.obj
+src/stream/rabbit/rabbit.obj src/stream/rc4/rc4_stream.obj src/stream/rc4/rc4_test.obj \
```
rofl0r: likewise

makefile.mingw
sjaeckel: changes to OBJECTS in the makefiles are automated and also checked in travis, therefore you can ignore them

sjaeckel:
> I lost this battle in libtommath and the result is that I have to keep a specially patched tommath.h for my perl bindings which is a pain for me. Frankly we did not gain anything in libtommath by starting using stdint.h.
> 
> And please do not give me those stories about how old is C99, on my Visual Studio 2008 there is no stdint.h so it only forces me to keep my stdint.h-less fork of libtomcrypt (as I already have to do with libtommath).

I still don't get it why you don't create your own stdint.h in e.g. `perl-CryptX/src/win32` and add this to your Include path when required...

makefile.mingw
rofl0r: if that is so, maybe it'd be cleaner to have those autogenerated changes in a single file which is included from all the other makefiles

sjaeckel: I also already thought of that, but 

1. this can't be used from the `makefile.msvc` as that one needs `.obj` file endings
2. this can't be used from `makefile.unix` AFAIK posix make doesn't support the `include` statement

which leaves us with `makefile.mingw` and `makefile_include.mk` which could use it... yeah it'd be one file less... then we should probably think if it'd make sense to migrate the `makefile.mingw` to be able to use `makefile_include.mk`... (which should be doable)

rofl0r: maybe you could use a variable for the extension, e.g. rabbit.$(OEXT)

karel-m: we can simply keep just one file name per line like:
```
...
src/stream/salsa20/salsa20_crypt.o \
src/stream/salsa20/salsa20_done.o \
src/stream/salsa20/salsa20_ivctr64.o \
src/stream/salsa20/salsa20_keystream.o \
src/stream/salsa20/salsa20_setup.o \
src/stream/salsa20/salsa20_test.o \
...
```

then the diffs will be quite easy to read, but it will be quite a lot of lines

karel-m: > wouldnt it make more sense to add the only new file in an own line

Maybe it is not obvious but the object files list is updated via script with algorithm that sorts all obj files and prints them so that every line is not longer than 100 chars (it always was like this). If anybody has a better idea it won't be hard to implement.

src/headers/tomcrypt_cipher.h
```diff
@@ -1057,7 +1055,29 @@ int sosemanuk_test(void);
...
+} rabbit_ctx;
+
+typedef struct {
+   rabbit_ctx master_ctx;
+   rabbit_ctx work_ctx;
+   unsigned char    ks[16];       /* last keystream block containing unused bytes */
+   ulong32   unused;       /* count fm right */
+} rabbit_state;
+
+int rabbit_setup(rabbit_state* rs, const unsigned char *key, ulong32 keysize);
```
karel-m: keysize, ivsize, length, msglen should be unsigned long

src/stream/rabbit/rabbit.c
```diff
@@ -0,0 +1,488 @@
...
+
+#ifdef LTC_RABBIT
+
+/* local/private prototypes */
+static LTC_INLINE ulong32 _rabbit_g_func(ulong32 x);
+static LTC_INLINE void _rabbit_next_state(rabbit_ctx *p_instance);
+
+/* -------------------------------------------------------------------------- */
+
+#define SWAP32(v) \
```
karel-m: cannot we just used already existing macro `BSWAP`?

makefile.mingw
sjaeckel: > but it will be quite a lot of lines

currently more than 420 lines... I'm against this, even 

> maybe you could use a variable for the extension, e.g. rabbit.$(OEXT)

true, then we could use this for all makefiles besides makefile.unix ...

src/stream/rabbit/rabbit.c
```diff
@@ -0,0 +1,488 @@
...
+   unsigned char* buffer;
+   ulong32 i;
+
+   LTC_ARGCHK(rs != NULL);
+   LTC_ARGCHK(input != NULL);
+   LTC_ARGCHK(output != NULL);
+
+   /* get right-sized keystream */
+   buffer = XMALLOC(inlen);
+   rabbit_keystream(rs, buffer, inlen);
```
sjaeckel: 1. this ignores the return value which it most likely shouldn't
2. this is the only stream cipher that uses the pattern that `crypt()` calls `keystream()` and not vice-versa, I'd prefer to have it everywhere the same

src/stream/rabbit/rabbit.c
```diff
@@ -0,0 +1,488 @@
...
+
+#ifdef ENDIAN_LITTLE
+#define U32TO32_LITTLE(v) (v)
+#endif
+
+#ifdef ENDIAN_BIG
+#define U32TO32_LITTLE(v) SWAP32(v)
+#endif
+
+#define U8TO32_LITTLE(p) \
```
sjaeckel: I guess we also already have a macro for this :)

src/stream/rabbit/rabbit.c
```diff
@@ -0,0 +1,488 @@
...
+
+   LTC_ARGCHK(rs != NULL);
+   LTC_ARGCHK(key != NULL);
+   LTC_ARGCHK(keylen <= 16);
+
+   /* init state */
+   XMEMSET(rs, 0, sizeof(rabbit_state));
+
+   /* pad key in tmpkey */
+   XMEMCPY(&tmpkey, key, keylen);
```
sjaeckel: Preferably `XMEMCPY(tmpkey, ...`, alternatively `XMEMCPY(&tmpkey[0], ...` but not like this :)

src/stream/rabbit/rabbit.c
```diff
@@ -0,0 +1,488 @@
...
+   /* Temporary variables */
+   ulong32 i0, i1, i2, i3, i;
+   unsigned char  tmpiv[8] = {0};
+
+   LTC_ARGCHK(rs != NULL);
+   LTC_ARGCHK(iv != NULL);
+   LTC_ARGCHK(ivlen <= 8);
+
+   /* pad iv in tmpiv */
+   XMEMCPY(&tmpiv, iv, ivlen);
```
sjaeckel: same

doc/crypt.tex
```diff
@@ -1391,6 +1391,57 @@ \chapter{Stream Ciphers}
...
+information.
+
+You begin initializing Rabbit by creating a state and calling \textit{rabbit\_setup()}
+with a 128-bit key.
+\begin{verbatim}
+rabbit_state rs;
+err = rabbit_setup(&rs, key, keylen);
+\end{verbatim}
+
+Finish initializing with a 64-bit iv.
```
sjaeckel: Probably better 'Finish initialization...'?

makefile.mingw
rofl0r: i'm assuming makefile.unix is for BSD make compatibility. don't they support variable expansion too ?

sjaeckel: [posix make](http://pubs.opengroup.org/onlinepubs/009695399/utilities/make.html) supports variable expansion but no include statement

sjaeckel: also I don't think it's a priority to change some automated task, so someone can better review it...

buggywhip:
> +#define SWAP32(v) \
> cannot we just used already existing macro BSWAP?
> 
perhaps, but I prefer we keep '32' as part of the name.  ...but it might not matter as Steffen is suggesting changing another macro, the only usage of SWAP32.

makefile.mingw
karel-m: makefile.unix also works with IBM's make (on AIX) and HP's make (on HP-UX) both of them are neither GNU make nor BSD make

IMO we should move make(file) related ideas into a separate issue/pr

src/stream/rabbit/rabbit.c
sjaeckel: the right one to use here is `LOAD32L`

src/stream/rabbit/rabbit.c
sjaeckel: please simply use `BSWAP`

sjaeckel:
> but it might not matter as Steffen is suggesting changing another macro, the only usage of SWAP32.

no, that's another one

src/stream/rabbit/rabbit.c
sjaeckel: oh I just realized... this section should be something like

```
#if defined(ENDIAN_LITTLE)
#define U32TO32_LITTLE(v) (v)
#elif defined(ENDIAN_BIG)
#define U32TO32_LITTLE(v) BSWAP(v)
#endif
```

src/stream/rabbit/rabbit.c
```diff
@@ -0,0 +1,440 @@
...
+   XMEMCPY(tmpkey, key, keylen);
+
+   /* Generate four subkeys */
+   LOAD32L(k0, tmpkey+ 0);
+   LOAD32L(k1, tmpkey+ 4);
+   LOAD32L(k2, tmpkey+ 8);
+   LOAD32L(k3, tmpkey+12);
+
+   /* done with tmpkey, wipe it */
+   XMEMSET(tmpkey, 0, sizeof(tmpkey));
```
sjaeckel: I'd say in case that's really worth a comment that should be `zeromem()` instead, as this would be the perfect example of a `memset()` being optimized away because the value that is zero'ed isn't referenced afterwards.

Also you can surround it with `#ifdef LTC_CLEAN_STACK`

src/stream/rabbit/rabbit.c
```diff
@@ -0,0 +1,442 @@
...
+/* -------------------------------------------------------------------------- */
+
+/* IV setup */
+int rabbit_setiv(rabbit_state* st, const unsigned char *iv, unsigned long ivlen)
+{
+   ulong32 i0, i1, i2, i3, i;
+   unsigned char  tmpiv[8] = {0};
+
+   LTC_ARGCHK(st != NULL);
+   LTC_ARGCHK(iv != NULL);
```
karel-m:
```C
   LTC_ARGCHK(iv != NULL || ivlen == 0);
```

src/stream/rabbit/rabbit.c
```diff
@@ -0,0 +1,442 @@
...
+   LTC_ARGCHK(ivlen <= 8);
+
+   /* pad iv in tmpiv */
+   XMEMCPY(tmpiv, iv, ivlen);
```
karel-m: if (ivlen > 0) XMEMCPY(tmpiv, iv, ivlen);

karel-m:
It seems that the following two cases give different results:

1/ skip rabbit_setiv
```C
rabbit_setup(&st, k, k_len);
rabbit_crypt(&st, pt, pt_len, ct);
```

2/ call rabbit_setiv with zero length IV
```C
rabbit_setup(&st, k, k_len);
rabbit_setiv(&st, zero, 0);
rabbit_crypt(&st, pt, pt_len, ct);
```

I am not sure what specification says about "no-IV" case. I found this when checking interoperability with another rabbit perl module (Crypt::Rabbit) which uses approach 1/.

We should clarify "no-IV" case in the documentation.

buggywhip:
> On 14Dec, 2017, at 12:43 AM, karel-m <notifications@github.com> wrote:
> 
> It seems that the following two cases give different results:
> 
> 1/ skip rabbit_setiv
> 
> rabbit_setup(&st, k, k_len);
> rabbit_crypt(&st, pt, pt_len, ct);
> 2/ call rabbit_setiv with zero length IV
> 
> rabbit_setup(&st, k, k_len);
> rabbit_setiv(&st, zero, 0);
> rabbit_crypt(&st, pt, pt_len, ct);
> I am not sure what specification says about "no-IV" case. I found this when checking interoperability with another rabbit perl module (Crypt::Rabbit) which uses approach 1/.
> 
> We should clarify "no-IV" case in the documentation.
> 
Good observation, and yes, you WILL get different answers when you skip the call vs making the call passing zeros.  This is because calling setiv() doesn't just load an IV, it also scrambles the state variables.  The LTC implementation started life as the eSTREAM reference implementation and looking again at the LTC code and the original reference code, neither supports conditional execution of the state scrambling logic based on a zero valued iv.  Passing zero and skipping the call WILL give different results.

Next, Rabbit state maintains two contexts: master and work.  setiv() in the reference implementation is where the master context with its partially scrambled key gets copied into the work context.  Without a call to setiv() another implementation will, at a minimum, need to perform that copy somewhere else.  ...and don't forget the additional scrambling.  Results will most likely deviate fm the designer's intent.

Rereading the paper submitted to eSTREAM, http://www.ecrypt.eu.org/stream/p3ciphers/rabbit/rabbit_p3.pdf, I find on page 3 the ambiguous phrase "...takes a 128- bit secret key and a 64-bit IV (if desired) as input...".  Beyond that the specification is silent WRT skipping the call vs passing zeros.

Also, looking at the 89 test vectors submitted to eSTREAM, there are no test cases without an iv although the vast majority of those cases had iv values of zero.  One might assume, on the basis of these vectors, there is an expectation to call setiv() even if only to pass a zero value.

Finally, the security of Rabbit is likely affected.  setiv() has setup logic that further scrambles the key independent of the value of an iv.  ...that wouldn't take place without the call.

Without additional information to the contrary I conclude a call to setiv() is mandatory. 

Next questions:  Fidelity or interoperability?  What do the authors say?  How do other implementations behave?  How are folks actually using it?  ???

And until, I submit the doc for setiv() in crypt.tex should be extended to say, "Set the IV using \textit{rabbit\_setiv() with a 64-bit IV, even when the value is zero."

buggywhip:
> On 14Dec, 2017, at 2:56 AM, Larry Bugbee <bugbee@mac.com> wrote:
> 
> Also, looking at the 89 test vectors submitted to eSTREAM, there are no test cases without an iv although the vast majority of those cases had iv values of zero.  One might assume, on the basis of these vectors, there is an expectation to call setiv() even if only to pass a zero value.


After all that I forgot to mention that all 89 test vectors passed WITH a call to setiv() passing zero.  ...mandatory for fidelity.

buggywhip:
Sorry, there's more.  

There are TWO sets of test vectors available fm eSTREAM.  One is available at
http://www.ecrypt.eu.org/stream/svn/viewcvs.cgi/ecrypt/trunk/submissions/sosemanuk/unverified.test-vectors?rev=210.  These are the 89 vectors mentioned previously.

The second set is packaged inside the source code ZIP file available at http://www.ecrypt.eu.org/stream/e2-rabbit.html.  Of this set of 6 vectors, 3 have a key and an expected value, but no IV.  Here one might infer you can run without the call to setiv().  (One side note: one of the 6 vectors is in error.  The 3rd line of out1 should read "96 D6 73 16 88 D1 68 DA 51 D4 0C 70 C3 A1 16 F4", not "EF 9A 69 71 8B 82 49 A1 A7 3C 5A 6E 5B 90 45 95".)

-----

And then there is the **Informational** RFC4503 (https://tools.ietf.org/html/rfc4503).  In section 2.4 it says, "After the key setup, the resulting inner state is saved as a master state. Then the IV setup is run to obtain the first encryption starting state."  This suggests making the call, but then section 3.2 says, in part, "It is possible to run Rabbit without the IV setup."  ...operative word, "possible". 

So, while technically you can run Rabbit without the call to setiv(), this ambiguity is not conducive to good interoperability.  

-----

Since LTC is a provider of primitives, I believe we should support both modes of operation and ensure there are no side effects beyond those encountered in the reference implementation.  ...that our doc highlight this distinction/ambiguity so users of Rabbit know to be sure all are on the same wavelength.  Personally, I think we should encourage they "always call setiv()".  ...that skipping the IV call should be reserved for protocol designers and closed systems.

karel-m:
I am for supporting both cases:
* rabbit_setup + rabbit_crypt
* rabbit_setup + rabbit_setiv(&st, NULL, 0) + rabbit_crypt

Note that `rabbit_setiv(&st, NULL, 0)` does not work now, it requires 2 changes I propose above.

Once we allow skipping `rabbit_setiv` we should make sure that `rabbit_setup` initializes the whole state (currently it for example does not set `st->unused = 0`). We should perhaps even add a test for "skip setiv" scenario so that we see valgring & co. results.

src/stream/rabbit/rabbit.c
karel-m: ping ^^

karel-m:
```diff
- XMEMCPY(tmpiv, iv, ivlen);
+ if (iv && ivlen > 0) XMEMCPY(tmpiv, iv, ivlen);
```

PR #335, opened 2017-12-06
Minor cleanup

sjaeckel:
Minor cleanup&lil'feature PR

The only thing really worth a review is the newly introduced `copy_or_zeromem()`

The rest is just cosmetics.

Reviews are as always very welcome!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,7 +39,7 @@ script:
 env:
   - |
     BUILDSCRIPT=".ci/meta_builds.sh"
-    BUILDNAME="META_BUILS"
+    BUILDNAME="META_BUILDS"
     BUILDOPTIONS="-DGMP_DESC"
   - |
     BUILDSCRIPT=".ci/valgrind.sh"
--- a/doc/crypt.tex
+++ b/doc/crypt.tex
@@ -5500,7 +5500,7 @@ in the same manner as the other data types except they use list of objects known
 \index{ltc\_asn1\_list structure}
 \begin{verbatim}
 typedef struct {
-   int                    type;
+   ltc_asn1_type          type;
    void                  *data;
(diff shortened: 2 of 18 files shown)


karel-m:
Ad `copy_or_zeromem`
* I am not sure if we want it to be part of public API
* if it'll will be part of public API then the name is quite mysterious (I do not have a better name)
* if `copy_or_zeromem` is handy in CCM then it should be perhaps also used in other encauth cases

sjaeckel:
> if copy_or_zeromem is handy in CCM then it should be perhaps also used in other encauth cases

I factored this out as I use it in my SIV PR as well, but I thought it doesn't really belong there, so I created this PR with other changes that belong to the features introduced there

> I am not sure if we want it to be part of public API

I'll make it private for now

> if it'll will be part of public API then the name is quite mysterious (I do not have a better name)

I think the name is okay'ish... probably one of the following would be better, not sure though:

1. `mem_copy_or_zero`
2. `memcpy_or_zero`
3. `copy_or_zero`
4. `copy_or_clean`

karel-m:
If you make it private then no need to care about the name.

sjaeckel:
> If you make it private then no need to care about the name.

well I prefer to have a proper name now so we don't have to think about it in case we want to make it public ;)

karel-m:
IIUC we are doing: _"copy src to dest or zero dest (all that in constant time)"_ which sounds to me more like `copy_or_zero_dest` but I can live with `copy_or_zeromem` as well - I leave it up to you.

Another thing:
```diff
- coz &= 0x1;
+ coz = (coz == 0) ? 0 : 1;
```
so that unexpected/accidental `coz = 2` means zeromem (accidentally zeroing is better that copying)

PR #337, opened 2017-12-07
prevent undefined behavior with LTC_PTHREAD

sjaeckel:
While reading through through [`man pthread_mutex_init`](https://linux.die.net/man/3/pthread_mutex_init) I stumbled over this:
"Attempting to initialize an already initialized mutex results in undefined behavior."

This patch should prevent this undefined behavior.


rofl0r:
now that i think about it, it seems curious that you need mutexes at all. you pass a struct containing the entire context to each of the prng function, so this seems quite thread-safe to me.

sjaeckel:
> ... so this seems quite thread-safe to me

really? even in case you're sharing one prng state in multiple threads?

rofl0r:
if you share state among threads you must of course protect it with locks if you modify it, but this is the first time i see the library doing the locking "for you". it is generally considered to be the responsibility of the library user to prevent concurrent updates.

sjaeckel:
> but this is the first time i see the library doing the locking "for you"

well that's already in there for a long time, there are also some static mutexes for e.g. the cipher/hash/prng registry or in the ecc code for the tables... just look it up...

I also think that it's up to the library-user to take care of that, but as it's not enabled by default he already does that by compiling ltc with pthread-support ;)

PR #341, opened 2018-01-21
Review of prngs

sjaeckel:
I had a look at the prng's at one point and forgot about these changes.

Regarding the last commit in the row "add possibility to seed PRNG as if it's imported" - I'm not 100% sure if that's a good approach or if it makes things too complicated & complex, but I wanted to put it up for discussion.

Thanks for your feedback to all of these changes.

### DIFF
--- a/doc/crypt.tex
+++ b/doc/crypt.tex
@@ -3822,10 +3822,14 @@ key, and any hash that produces at least a 256--bit output.  However, to make th
 it has been fixed to those choices.
 
 Fortuna is more secure than Yarrow in the sense that attackers who learn parts of the entropy being
-added to the PRNG learn far less about the state than that of Yarrow.  Without getting into to many
+added to the PRNG learn far less about the state than that of Yarrow.  Without getting into too many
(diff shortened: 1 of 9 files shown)


karel-m:
Ad export/import: what about turning all prngs' export functions consistently into:
```C
int ALGNAME_prng_export(unsigned char *out, unsigned long *outlen, prng_state *prng)
{
   unsigned long len = ALGNAME_prng_desc.export_size;

   LTC_ARGCHK(prng   != NULL);
   LTC_ARGCHK(out    != NULL);
   LTC_ARGCHK(outlen != NULL);

   if (*outlen < len) {
      *outlen = len;
      return CRYPT_BUFFER_OVERFLOW;
   }

   if (ALGNAME_prng_read(out, len, prng) != len) {
      return CRYPT_ERROR_READPRNG;
   }

   *outlen = len;
   return CRYPT_OK;
}
```

sjaeckel:
Fine like that?

Regarding docs, I'll update them in the next PR regarding the PRNG's.

PR #342, opened 2018-01-22
Add new file API to GCM

sjaeckel:

This adds two new API functions `gcm_file()` and `gcm_filehandle()` and a new demo `aesgcm`.

The demo is GNU only for now and real _demo quality_ ;)

I've tried to catch error-cases as good as possible in the new API functions, please have especially a look at the error handling.

### DIFF
--- a/.ci/build.sh
+++ b/.ci/build.sh
@@ -1,5 +1,14 @@
 #!/bin/bash
-echo "$1 ($2, $3)..."
+
+if echo "$1" | grep -q "STOCK"
+then
+   MAKE_OPTIONS="$3 all"
+   CFLAGS="$CFLAGS -Wno-missing-braces"
+else
+   MAKE_OPTIONS="$3 all_test"
+fi
+
+echo "$1 ($2, $MAKE_OPTIONS)..."
 
 make clean 1>/dev/null 2>/dev/null
 
@@ -12,7 +21,7 @@ else
   MAKE_JOBS=8
 fi
 
-CFLAGS="$2 $CFLAGS $4" EXTRALIBS="$5" make -j$MAKE_JOBS -f $3 all_test 1>gcc_1.txt 2>gcc_2.txt
+make -j$MAKE_JOBS -f $MAKE_OPTIONS CFLAGS="$2 $CFLAGS $4" EXTRALIBS="$5" 1>gcc_1.txt 2>gcc_2.txt
 mret=$?
 cnt=$(wc -l < gcc_2.txt)
 # ignore 1 line since ar prints to stderr instead of stdout and ar is called for
--- a/.gitignore
+++ b/.gitignore
@@ -22,6 +22,8 @@ doc/crypt.pdf
 doc/refman.pdf
 
 # *nix/windows test executables
+aesgcm
+aesgcm.exe
 constants
 constants.exe
 ltcrypt
--- a/demos/aesgcm.c
+++ b/demos/aesgcm.c
@@ -1,0 +1,142 @@
+/* LibTomCrypt, modular cryptographic library -- Tom St Denis
+ *
+ * LibTomCrypt is a library that provides various cryptographic
+ * algorithms in a highly modular and flexible manner.
+ *
+ * The library is free for all purposes without any express
+ * guarantee it works.
+ */
+
+/**
+  @file aesgcm.c
+  AES128-GCM demo - file en-&decryption, Steffen Jaeckel
+*/
+
+#define _GNU_SOURCE
+
+#include <tomcrypt.h>
+#include <stdio.h>
+#include <stdint.h>
+#include <sys/stat.h>
+#include <sys/types.h>
+#include <dirent.h>
+#include <string.h>
+#include <errno.h>
+#include <fcntl.h>
+#include <unistd.h>
+
+
+static off_t fsize(const char *filename)
+{
+   struct stat st;
+
+   if (stat(filename, &st) == 0) return st.st_size;
+
+   return -1;
+}
+
+static int mv(const char *old_name, const char *new_name)
+{
+   int fd;
+   if (rename(old_name, new_name) == -1) return -1;
+   fd = open(new_name, 0);
+   if (fd == -1) return -1;
+   if (fsync(fd) != 0) goto OUT;
+   syncfs(fd);
+OUT:
+   close(fd);
+   return 0;
+}
+
+/* https://stackoverflow.com/a/23898449 */
+static void scan_hex(const char* str, uint8_t* bytes, size_t blen)
+{
+   uint8_t  pos;
+   uint8_t  idx0;
+   uint8_t  idx1;
+
+   const uint8_t hashmap[] =
+   {
+     0x00, 0x01, 0x02, 0x03, 0x04, 0x05, 0x06, 0x07, /* 01234567 */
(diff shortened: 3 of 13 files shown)


sjaeckel:
@karel-m Should we probably update `LTC_EASY` to contain `GCM`?
I'll open another PR then with a suggestion.

karel-m:
I am not using `LTC_EASY` at all but my opinion here is: keep `LTC_EASY` as it is now (without GCM)

karel-m:
MSVC failure:
```
src\encauth\gcm\gcm_file.c(20) : fatal error C1083: Cannot open include file: 'unistd.h': No such file or directory
NMAKE : fatal error U1077: 'cl' : return code '0x2'
Stop.
```

sjaeckel:
> src\encauth\gcm\gcm_file.c(20) : fatal error C1083: Cannot open include file: 'unistd.h': No such file or directory

```
diff --git a/src/encauth/gcm/gcm_file.c b/src/encauth/gcm/gcm_file.c
index f3c88fb..ade1de3 100644
--- a/src/encauth/gcm/gcm_file.c
+++ b/src/encauth/gcm/gcm_file.c
@@ -19,4 +19,2 @@
 
-#include <unistd.h>
-
 /**
diff --git a/src/encauth/gcm/gcm_filehandle.c b/src/encauth/gcm/gcm_filehandle.c
index ea6fb3f..53c85d2 100644
--- a/src/encauth/gcm/gcm_filehandle.c
+++ b/src/encauth/gcm/gcm_filehandle.c
@@ -19,3 +19,7 @@
 
+#if defined(_WIN32) || defined(_WIN32_WCE)
+#define ftruncate _chsize
+#else
 #include <unistd.h>
+#endif
 ```

fine like that?

**Comment by karel-m** [comment]
* gcm_file.c fix is fine
* the `#ifdef` IMO will not work well with gcc+mingw64

Too much I/O magic in `gcm_filehandle` I have to think it over

**Comment by sjaeckel** [comment]
> the #ifdef IMO will not work well with gcc+mingw64

you're right, it has to be `#if (defined(_WIN32) || defined(_WIN32_WCE)) && !defined(__GNUC__)`

> Too much I/O magic in gcm_filehandle I have to think it over

yeah, I already thought about it but couldn't find another solution.
Probably you can find something better.
It's basically the first FILE API that does en-/decryption, all the other FILE API's were for hashes or MAC's so they only had to read the file and then store some bytes...

**Comment by karel-m** [comment]
also  `mkstemp` in `file_test` is a troublemaker when compiling with MSVC

**Comment by sjaeckel** [comment]
> also mkstemp in file_test is a troublemaker when compiling with MSVC

as mentioned it's GNU only, I don't care about MSVC for that demo :)

**Comment by karel-m** [comment]
and of course `"/tmp/ltc_test_XXXXXX"` is not windows friendly as well

**Comment by karel-m** [comment]
Well, the thing is that it is not **a demo** it is **a test**, thus I am not able to `nmake ... test & .\test.exe`

**Comment by sjaeckel** [comment]
> Well, the thing is that it is not a demo it is a test, thus I am not able to nmake ... test & .\test.exe

ah, right.

**Comment by karel-m** [comment]
Here is a complete working MSVC fix:
```diff
diff --git a/src/encauth/gcm/gcm_file.c b/src/encauth/gcm/gcm_file.c
index f3c88fb3..ade1de36 100644
--- a/src/encauth/gcm/gcm_file.c
+++ b/src/encauth/gcm/gcm_file.c
@@ -17,8 +17,6 @@
 #ifdef LTC_GCM_MODE
 #ifndef LTC_NO_FILE
 
-#include <unistd.h>
-
 /**
   Process a file.
 
diff --git a/src/encauth/gcm/gcm_filehandle.c b/src/encauth/gcm/gcm_filehandle.c
index ea6fb3f9..e30fdf49 100644
--- a/src/encauth/gcm/gcm_filehandle.c
+++ b/src/encauth/gcm/gcm_filehandle.c
@@ -17,7 +17,11 @@
 #ifdef LTC_GCM_MODE
 #ifndef LTC_NO_FILE
 
+#if defined(_MSC_VER)
+#define ftruncate _chsize
+#else
 #include <unistd.h>
+#endif
 
 /**
   Process a filehandle.
diff --git a/tests/file_test.c b/tests/file_test.c
index c8ca2ece..54e9d4b2 100644
--- a/tests/file_test.c
+++ b/tests/file_test.c
@@ -9,7 +9,12 @@
 /* test file related functions */
 
 #include <tomcrypt_test.h>
+
+#if defined(_MSC_VER)
+#define unlink _unlink
+#else
 #include <unistd.h>
+#endif
 
 int file_test(void)
 {
@@ -110,13 +115,10 @@ int file_test(void)
    {
       unsigned char iv[32] = { 0x00, 0x01, 0x02, 0x03, 0x04, 0x05, 0x06, 0x07, 0x08, 0x09, 0x0A, 0x0B, 0x0C, 0x0D, 0x0E, 0x0F,
                                 0x10, 0x11, 0x12, 0x13, 0x14, 0x15, 0x16, 0x17, 0x18, 0x19, 0x1A, 0x1B, 0x1C, 0x1D, 0x1E, 0x1F };
-      int tmp, res;
-      char tmp_X_enc[] = "/tmp/ltc_test_XXXXXX";
-      char tmp_X_dec[] = "/tmp/ltc_test_XXXXXX";
+      int res;
+      char tmp_X_enc[] = "tests/tempfile.enc";
+      char tmp_X_dec[] = "tests/tempfile.dec";
 
-      tmp = mkstemp(tmp_X_enc);
-      if (tmp == -1) return CRYPT_FILE_NOTFOUND;
-      close(tmp);
       res = 666;
       if ((err = gcm_file(iaes,
                      key,    32,
@@ -132,12 +134,6 @@ int file_test(void)
          goto GCM_OUT_1;
       }
 
-      tmp = mkstemp(tmp_X_dec);
-      if (tmp == -1) {
-         err = CRYPT_FILE_NOTFOUND;
-         goto GCM_OUT_1;
-      }
-      close(tmp);
       res = 666;
       if ((err = gcm_file(iaes,
                      key,    32,
```

The rest of my comments later.

karel-m:
As mentioned above the *_file API is a different story then *_file API for hashes and macs.

To be honest I do not use even existing *_file API as it has some drawbacks (e.g. on MS Windows this API does not support unicode/UTF16 filenames). We should keep in mind that ltc is a crypto library.

Other thing is that we are here in fact proposing a new format of GCM/AES encrypted file:

```
|ciphertext|tag-16-bytes|
```

For example why not to use?

```
|iv-16-bytes|ciphertext|tag-16-bytes|
```

I would even propose to support encrypting file with arbitrary length secret and deriving IV + KEY via pbkdf2 and using format like:

```
|pbkdf2-salt|ciphertext|tag-16-bytes|
```

I simply want to say that gcm_file API is perhaps something that should be rather handled on application level.

The other thing to consider here is whether it would not be better to use chacha20poly1305 instead of GCM as it has much better (== higher) limits on plain/ciphertext size.

As for the implementation I would prefer to avoid seeking because current approach does not allow to implement encrypting STDIN to STDOUT - something like:

```
tar cv inputdir | gzip | aesgcm -e - output.tag.gz.gcm secretkey
```

Which I believe can be done as there are utilities like crypt/mcrypt that support this mode.

I am for moving/incorporating gcm_filehandle.c + gcm_file.c into demos/aesgcm.c and keeping it just as a demo for now.

sjaeckel:
You're right! This decision&implementation should really stay on application level.

PR #353, opened 2018-03-02
Minor improvements

sjaeckel:
Preparation for further work


src/headers/tomcrypt_pk.h
```diff
@@ -9,14 +9,14 @@
...
+   PK_PUBLIC   = 0x0000,
+   PK_PRIVATE  = 0x0001,
+   PK_STD      = 0x1000,
+   PK_STD_PUB  = PK_STD | PK_PUBLIC,
+   PK_STD_PRIV = PK_STD | PK_PRIVATE,
 };
 
-/* Indicates standard output formats that can be read e.g. by OpenSSL or GnuTLS */
-#define PK_STD          0x1000
-
```
karel-m: Please revert the changes above. There are even more key types in my new ecc branch:
```C
enum {
   PK_PUBLIC=0,
   PK_PRIVATE=1
};

/* Indicates standard output formats that can be read e.g. by OpenSSL or GnuTLS */
#define PK_STD          0x1000
/* Indicates compressed public ECC key */
#define PK_COMPRESSED   0x2000
/* Indicates ECC key with the curve specified by OID */
#define PK_CURVEOID     0x4000
```
Let's discuss it later.

sjaeckel: why can't you just put them in there?

sjaeckel: btw. it doesn't even really clash...
I just successfully rebased the ecc-branch on top of this one and there was only a trivial merge-conflict that could be auto-resolved by my mergetool.
Compilation works without a warning and the tests succeed as well.

sjaeckel: and here the patch that integrates your new defines in the enum:
```
diff --git a/src/headers/tomcrypt_pk.h b/src/headers/tomcrypt_pk.h
index fcac2fa..f8332be 100644
--- a/src/headers/tomcrypt_pk.h
+++ b/src/headers/tomcrypt_pk.h
@@ -12,13 +12,15 @@
 enum public_key_type {
-   PK_PUBLIC   = 0x0000,
-   PK_PRIVATE  = 0x0001,
-   PK_STD      = 0x1000,
-   PK_STD_PUB  = PK_STD | PK_PUBLIC,
-   PK_STD_PRIV = PK_STD | PK_PRIVATE,
+   PK_PUBLIC      = 0x0000,
+   PK_PRIVATE     = 0x0001,
+   /* Indicates standard output formats that can be read e.g. by OpenSSL or GnuTLS */
+   PK_STD         = 0x1000,
+   /* Indicates compressed public ECC key */
+   PK_COMPRESSED  = 0x2000,
+   /* Indicates ECC key with the curve specified by OID */
+   PK_CURVEOID    = 0x4000,
+
+   PK_STD_PUB     = PK_STD | PK_PUBLIC,
+   PK_STD_PRIV    = PK_STD | PK_PRIVATE,
 };
 
-/* Indicates compressed public ECC key */
-#define PK_COMPRESSED   0x2000
-/* Indicates ECC key with the curve specified by OID */
-#define PK_CURVEOID     0x4000
 int rand_prime(void *N, long len, prng_state *prng, int wprng);
```

karel-m: Better except:
```
   PK_STD_PUB     = PK_STD | PK_PUBLIC,
   PK_STD_PRIV    = PK_STD | PK_PRIVATE,
```
is a bit odd as there can also be something like:
```
PK_NNN_PUB = PK_BULIC|PK_COMPRESSED|PK_CURVEOID
PK_NNN_PUB = PK_BULIC|PK_COMPRESSED
PK_NNN_PUB = PK_BULIC|PK_CURVEOID
etc.
```
and `enum public_key_type` should be perhaps `enum key_type`

sjaeckel: > is a bit odd as there can also be something like: ...

which I think is fine, we should create combinations for all implemented combinations, so someone doesn't start to combine wildly and wonders why it doesn't work.

> and enum public_key_type should be perhaps enum key_type

even the wikipedia article is called https://en.wikipedia.org/wiki/Public-key_cryptography so I think that name should be fine :)

karel-m: > which I think is fine, we should create combinations for all implemented combinations, so someone doesn't start to combine wildly and wonders why it doesn't work.

I see your point but I do not like it. I prefer to have exactly one constant for one bit.

> even the wikipedia article is called https://en.wikipedia.org/wiki/Public-key_cryptography so I think that name should be fine :)

Yes but having something like:
```C
int  ecc_export(unsigned char *out, unsigned long *outlen, public_key_type type, ecc_key *key);
```
might be a bit confusing as the user may think that the function exports just public key.

karel-m:
Idea: what about to turn `scan_hex` into `src/misc/base16/base16_decode.c` + adding also `base16_encode.c`?

src/headers/tomcrypt_pk.h
sjaeckel: > I see your point but I do not like it. I prefer to have exactly one constant for one bit.

Mkay



> may think that the function exports just public key

...and when that person looks at the definition it becomes obvious that this enum also includes private keys :)

I think `key_type` is not specific enough and I can't find another good name...

PR #356, opened 2018-03-07
Keccak

karel-m:
### Checklist


karel-m:
The original request (from a user of my perl bindings) with some reasoning why it may be a good idea to add keccak.

PR #363, opened 2018-03-23
Improve Fortuna PRNG

sjaeckel:

This updates the Fortuna implementation to be more in-line with the original specification.

It adds the `fortuna_add_random_event()` function, an alternative to `fortuna_add_entropy()` and makes `fortuna_update_seed()` public.

Additionally it fixes a bug in the reseed rate-limit which should only happen after `pool0_len >= 64` _AND_ `rate-limit expired` which was an _OR_ before.


karel-m:
Not that I am against this PR, I only need more time for review, stay tuned.

src/prngs/fortuna.c
```diff
@@ -61,6 +61,29 @@ static void _fortuna_update_iv(prng_state *prng)
...
+   FILETIME CurrentTime;
+   ULARGE_INTEGER ul;
+   GetSystemTimeAsFileTime(&CurrentTime);
+   ul.LowPart  = CurrentTime.dwLowDateTime;
+   ul.HighPart = CurrentTime.dwHighDateTime;
+   cur_time = ul.QuadPart;
+   cur_time -= CONST64(116444736000000000); /* subtract epoch in microseconds */
+   cur_time /= 1000; /* nanoseconds -> microseconds */
+#endif
+   return cur_time / 100;
```
tyzoid: `cur_time` here is uninitialized if neither `_POSIX_C_SOURCE >= 200112L` or `_WIN32` is defined. I'd recommend either initializing it to `0` on line 68, or adding an additional `#else` before line 82, with an `#error` trigger to positively alert to a problem.

src/headers/tomcrypt_custom.h
```diff
@@ -358,11 +358,29 @@
...
 #define LTC_FORTUNA_WD    10
 #endif
 
+/* make sure only one of
+ *   LTC_FORTUNA_RESEED_RATELIMIT_STATIC
+ * and
+ *   LTC_FORTUNA_RESEED_RATELIMIT_TIMED
+ * is defined.
+ */
+#undef LTC_FORTUNA_RESEED_RATELIMIT_TIMED
```
tyzoid: We should warn if we `#undef` this constant here, because it indicates a fault in user-code.

src/prngs/fortuna.c
tyzoid: As pointed out by @sjaeckel on IRC, this isn't a problem, because `src/headers/tomcrypt_custom.h` on line 362 checks to make sure that these constants are defined.

src/headers/tomcrypt_custom.h
sjaeckel: done

PR #364, opened 2018-03-25
Add curve25519

sjaeckel:
* ~~documentation is added or updated~~ will be done in an upcoming PR

This adds a LTC API on top of the original curve25519 (ed25519&X25519) implementations of djb.

The PR is based on top of the feature/curve25519_originals branch which contains the original sources + their amalgamation.
This PR contains all the changes that I've made to the original sources + the new API.

As soon as we agree on the API I'll add docs.

[Edit]
btw. I've changed the base-branch to feature/curve25519_originals so this review contains only the major changes to the sources&added functions
should I also open a PR for feature/curve25519_originals or do we simply change the target branch to develop before merging?


eduardsui:
Any news about this? Is really useful in TLS 1.3 key share.

sjaeckel:
@karel-m @eduardsui & whoever else feels like... you're all welcome to review, after I've got the OK on the API I'll finish&add the docs.

karel-m:
@sjaeckel is the branch `feature/curve25519` anyhow special? I am unable to do `git checkout`.

```
$ git fetch origin
$ git checkout feature/curve25519
error: pathspec 'feature/curve25519' did not match any file(s) known to git.
```

sjaeckel:
> @sjaeckel is the branch feature/curve25519 anyhow special? I am unable to do git checkout.

```
~/tmp $ git clone https://github.com/libtom/libtomcrypt
Cloning into 'libtomcrypt'...
remote: Counting objects: 17259, done.
remote: Compressing objects: 100% (79/79), done.
remote: Total 17259 (delta 45), reused 74 (delta 33), pack-reused 17145
Receiving objects: 100% (17259/17259), 19.89 MiB | 1.72 MiB/s, done.
Resolving deltas: 100% (13064/13064), done.
Checking connectivity... done.
~/tmp $ cd libtomcrypt/
~/tmp/libtomcrypt $ git co feature/curve25519
Branch feature/curve25519 set up to track remote branch feature/curve25519 from origin.
Switched to a new branch 'feature/curve25519'
~/tmp/libtomcrypt $ git config alias.co
checkout
```

not that I'm aware of

sjaeckel:
\*ping\*

karel-m:
First of all some link to documents we might want to follow:
* https://tools.ietf.org/html/rfc7748 - Elliptic Curves for Security (X25519 and X448)
* https://tools.ietf.org/html/rfc8032 - Edwards-Curve Digital Signature Algorithm (EdDSA)
* https://tools.ietf.org/html/rfc8410 - priv/pub key formats (ASN.1)
* https://tools.ietf.org/html/rfc8411 - related to OIDs
* https://tools.ietf.org/html/rfc8418 - EC Diffie-Hellman with X25519 and X448
* https://tools.ietf.org/html/rfc8419 - EdDSA - Cryptographic Message Syntax (CMS)

We are not going to deal with Curve448 in this PR but we should keep it in mind so that in the future the maths stuff can be shared between 25519/448.

Ad 25519 API I am not sure about naming convention:
* `ed25519_sign` 
* `ed25519_verify`
* `x25519_shared_secret`

In our old `ecc_*` world we use
* `ecc_sign_hash` (despite the fact that industry standard name here is "ecdsa")
* `ecc_verify_hash`
* `ecc_shared_secret` (despite the fact that industry standard name here is "ecdh")

So why not this:
* `ec25519_sign_hash`  (+ later perhaps also `ec448_sigh_hash`)
* `ec25519_verify_hash`
* `ec25519_shared_secret`

Things are IMO getting more tricky with signatures. According to EdDSA/rfc8032 there are:
* ed25519_sign_ph = Pre-Hashed-EdDSA
* ed25519_sign = Pure-EdDSA
* ed25519_sign_ctx  = some odd stuff

The Pre-Hashed-EdDSA fits well our current interface `ec25519_sign_hash` / `ec25519_verify_hash`. However, the Pure-EdDSA IMO will need init-update-done style API like: `ec25519_sign_init` / `ec25519_sign_update` / `ec25519_sign_done` (+ similar set of functions for verification). We can of course decide to only support Pure-EdDSA created in "one-go" (e.g. `ec25519_sign_pure`).

As for the keys we will probably need some import/export functions supporting ASN.1 defined in rfc8410. Considering our recently face-lifted ecc API I would prefer renaming `ed25519_export` / `ed25519_import` to something like:

```C
int  ec25519_set_key(const unsigned char *in, unsigned long inlen, int type, ecc_key *key);
int  ec25519_get_key(unsigned char *out, unsigned long *outlen, int type, const ecc_key *key);
```
And maybe rename `ed25519_set_key` to `ec25519_set_keypair`.

karel-m:
An interesting/surprising quote from https://tools.ietf.org/html/rfc8410#section-4
```
Both [RFC7748] and [RFC8032] define the public key value as being a
byte string.  It should be noted that the public key is computed
differently for each of these documents; thus, the same private key
will not produce the same public key.
```

sjaeckel:
> An interesting/surprising quote from https://tools.ietf.org/html/rfc8410#section-4

that's exactly the reason why there's a separate `ed25519_X()` and `x25519_X()` API ...

... I was too lazy to include the key conversion :-) ...

karel-m:
As for the keys they need to be very carefully documented as it is really a mess.

Here are some IMO useful links related to ec25519 keys:
* https://blog.mozilla.org/warner/2011/11/29/ed25519-keys/ (discussion only about ed25519)
* https://github.com/jedisct1/libsodium/issues/77 (this is interesting as it is a request + discussion about unifying ed25519 / x25519 - which I have mentioned above)
* https://signal.org/docs/specifications/xeddsa/ (this is how signal handles eddsa/eddh with the same key)
* https://moderncrypto.org/mail-archive/curves/2015/000376.html (a thread about unifying ed25519/x25519 public keys)

Slightly OT: as for the actual implementation I have found:
* https://tweetnacl.cr.yp.to/software.html
* it implements Curve25519 and Ed25519 and looks simple / concise (= maintainable)
* to me it looks like a slightly better candidate to include into libtomcrypt than the current reference implementation

sjaeckel:
Alright, I've made the move to tweetnacl. (that was easier than I expected :-) )

Regarding the key conversion I'm not 100% sure where we should head to... this is all pretty unclear to me until now, AFAIU everything is possible (ed->x and x->ed conversion) but it's not sure if you really should do it in one way or the other regarding cryptographic security...

You're proposing to provide an `ec25519` API where I'm not sure if that's really useful as we don't really know what the user wants to do and what its environment is... especially since everyone seems to say "if you can afford it, use an Ed key for signatures and an X key for shared secret".

In case we add it it'd mean that we have sign+verify+shared_secret in ec25519 space, but we still have to provide make_key+import+export+set_keypair in Ed and X space!

I'd be fine to provide API functions to do the conversion, but for now I tend to stay with only two different API's.

Your thoughts? Everybody's welcome to comment on this :-)

sjaeckel:
@karel-m do you probably have the time to look at this in the next few days?

Regarding the things you mentioned:

* **tweetnacl:** that was a good idea to move away to this implementation!
* **ec25519 API:** see my comment above, I've moved tweetnacl already to the correct folder, so there's at least only 3 of them when we add ec25519 API's...
* **key conversion:** also see my comment above, I currently don't need it and I think it can be added at a later point of time.
* **other EdDSA modes:** as you proposed they're going to get different API's and as I currently also don't need this functionality it can be added later when required/requested.

karel-m:
Unfortunately I am do not have enough time to make a thorough check but after a quick skim I like it; although I might have some suggestions later.

I am for merging this into `develop` + creating a new issue tracking TODOs related to curve25519 (doc update, check interoperability with other libs, key conversion etc.)

sjaeckel:
thanks!

> although I might have some suggestions later.

I'm looking forward to that!

> doc update

I will do now and merge afterwards (if I don't do it now, I'll do it never :D )

> check interoperability with other libs

will add an issue

> key conversion

will add an issue

sjaeckel:
While starting to add the documentation, I already stumbled over some parts but I will open a separate PR soon with some API changes&additions + documentation

PR #365, opened 2018-03-25
Make base64_decode relaxed mode less relaxed

karel-m:
### Checklist

### DIFF
--- a/doc/crypt.tex
+++ b/doc/crypt.tex
@@ -6550,6 +6550,8 @@ int base64_decode(   const char *in,
 \end{verbatim}
 
 The function \textit{base64\_decode} works in a relaxed way which allows decoding some inputs that do not strictly follow the standard.
+The relaxed mode ignores white-spaces (\textit{CR}, \textit{LF}, \textit{TAB}, \textit{space}), does not care about trailing \textit{=}
+and also ignores the last input byte in case it is \textit{NUL}.
 If you want to be strict during decoding you can use:
 \index{base64\_strict\_decode()}
 \begin{verbatim}
--- a/doc/crypt.tex
+++ b/doc/crypt.tex
@@ -6578,6 +6578,11 @@ ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789-_
 Those characters are sometimes also called URL and filename safe alphabet.
 The interface is analogous to \textit{base64\_xxxx} functions in previous chapter.
 
+\index{base64url\_encode()}
+\index{base64url\_strict\_encode()}
+\index{base64url\_decode()}
+\index{base64url\_strict\_decode()}
+\index{base64url\_sane\_decode()}
 \begin{verbatim}
 int base64url_encode(const unsigned char *in,  unsigned long len,
                                     char *out, unsigned long *outlen);
--- a/doc/crypt.tex
+++ b/doc/crypt.tex
@@ -6549,9 +6549,8 @@ int base64_decode(   const char *in,
                   unsigned long *outlen);
 \end{verbatim}
 
-The function \textit{base64\_decode} works in a relaxed way which allows decoding some inputs that do not strictly follow the standard.
-The relaxed mode ignores white-spaces (\textit{CR}, \textit{LF}, \textit{TAB}, \textit{space}), does not care about trailing \textit{=}
-and also ignores the last input byte in case it is \textit{NUL}.
+The function \textit{base64\_decode} works in a dangerously relaxed way which allows decoding some inputs that do not strictly follow the standard.
+
 If you want to be strict during decoding you can use:
 \index{base64\_strict\_decode()}
 \begin{verbatim}
@@ -6559,6 +6558,16 @@ int base64_strict_decode(   const char *in,
                          unsigned long  len,
                          unsigned char *out,
                          unsigned long *outlen);
+\end{verbatim}
+
+There is also so called sane mode that ignores white-spaces (\textit{CR}, \textit{LF}, \textit{TAB}, \textit{space}),
+does not care about trailing \textit{=} and also ignores the last input byte in case it is \textit{NUL}.
+\index{base64\_sane\_decode()}
+\begin{verbatim}
+int base64_sane_decode(   const char *in,
+                       unsigned long  len,
+                       unsigned char *out,
+                       unsigned long *outlen);
 \end{verbatim}
 
 \subsection{URL--safe 'base64url' encoding}
@@ -6581,6 +6590,9 @@ int base64url_decode(   const char *in,  unsigned long len,
 
 int base64url_strict_decode(   const char *in,  unsigned long len,
                             unsigned char *out, unsigned long *outlen);
+
+int base64url_sane_decode(   const char *in,  unsigned long len,
+                          unsigned char *out, unsigned long *outlen);
 \end{verbatim}
 
 \mysection{Base32 Encoding and Decoding}
--- a/src/misc/base64/base64_decode.c
+++ b/src/misc/base64/base64_decode.c
@@ -145,7 +145,7 @@ static int _base64_decode_internal(const char *in,  unsigned long inlen,
 
 #if defined(LTC_BASE64)
 /**
-   Relaxed base64 decode a block of memory
+   Dangerously relaxed base64 decode a block of memory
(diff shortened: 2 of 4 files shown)


karel-m:
> @sjaeckel wrote off-list:
> I think we should try to keep this super relaxed mode if possible, as that's the one that is compatible with the old 1.17 behavior, right?

IMO the current super relaxed mode may be dangerous when you are decoding data from external sources (e.g. from a user request) because we are "able" to decode nearly any invalid input data. I am open to discussion what other characters should be ignored in relaxed mode.

sjaeckel:
> IMO the current super relaxed mode may be dangerous when you are decoding data from external sources (e.g. from a user request)...

I fully agree! ... and that's also what is said in [RFC 4648](https://tools.ietf.org/html/rfc4648#section-12) ...

But I also think that we shouldn't break existing code or at least give the possibility to enable the old behavior if required... so we've several options I guess...

1. add a new API function `base64_insane_decode()` and fix the original
1. add a new API function `base64_sane_decode()` and keep the original
1. add a way to set the level of sanity checking... 
a. globally via a global 
b. globally via `base64_set_decode_sanity()` 
c. on the API level via exporting the sanity parameter
d. at compile time
1. or a combination of some of the above or something else I didn't think of?

karel-m:
From the options above I prefer `base64_sane_decode()` approach

sjaeckel:
> From the options above I prefer `base64_sane_decode()` approach

:+1: 

`base64_sane_decode()` shall it be

so we have now 3 possible values `enum(insane, relaxed, strict)` where the default `base64_decode()` is `insane` :D

karel-m:
`base64_sane_decode` or `base64_relaxed_decode`?

sjaeckel:
I think `base64_sane_decode()` is more appropriate as `base64_relaxed_decode()` is basically lying because `base64_decode()` is more relaxed...

src/misc/base64/base64_decode.c
```diff
@@ -89,20 +95,29 @@ static int _base64_decode_internal(const char *in,  unsigned long inlen,
 
    g = 0; /* '=' counter */
    for (x = y = z = t = 0; x < inlen; x++) {
+       if (in[x] == 0 && x == (inlen - 1) && mode != strict) {
```
sjaeckel: can you please add parentheses?

src/misc/base64/base64_decode.c
```diff
@@ -118,7 +133,7 @@ static int _base64_decode_internal(const char *in,  unsigned long inlen,
 
    if (y != 0) {
       if (y == 1) return CRYPT_INVALID_PACKET;
-      if ((y + g) != 4 && is_strict && map != map_base64url) return CRYPT_INVALID_PACKET;
+      if ((y + g) != 4 && (mode == strict) && map != map_base64url) return CRYPT_INVALID_PACKET;
```
sjaeckel: here as well please

src/misc/base64/base64_decode.c
```diff
@@ -156,6 +171,20 @@ int base64_strict_decode(const char *in,  unsigned long inlen,
 {
    return _base64_decode_internal(in, inlen, out, outlen, map_base64, strict);
 }
+
+/**
+   Relaxed base64 decode a block of memory
```
sjaeckel: Probably this (and the URL variant) comment should be changed?

tests/base64_test.c
```diff
@@ -47,39 +49,68 @@ int base64_test(void)
...
-         {"vuiSPKIl8PiR5O-rC4z9_xTQKZ0=", 1},
-         {"vuiS*PKIl8P*iR5O-rC4*z9_xTQKZ0", 0},
-         {"vuiS*PKIl8P*iR5O-rC4*z9_xTQKZ0=", 0},
-         {"vuiS*PKIl8P*iR5O-rC4*z9_xTQKZ0==", 0},
-         {"vuiS*PKIl8P*iR5O-rC4*z9_xTQKZ0===", 0},
-         {"vuiS*PKIl8P*iR5O-rC4*z9_xTQKZ0====", 0},
-         {"vuiS*=PKIl8P*iR5O-rC4*z9_xTQKZ0=", 0},
-         {"vuiS*==PKIl8P*iR5O-rC4*z9_xTQKZ0=", 0},
-         {"vuiS*===PKIl8P*iR5O-rC4*z9_xTQKZ0=", 0},
+         {"vuiSPKIl8PiR5O-rC4z9_xTQKZ0", strict},                       /* 0 */
```
sjaeckel: :+1:

tests/base64_test.c
```diff
@@ -47,39 +49,68 @@ int base64_test(void)
...
           DO(base64url_strict_decode(url_cases[x].s, slen1, tmp, &l1));
-       else
+          DO(do_compare_testvector(tmp, l1, special_case, sizeof(special_case) - 1, "base64url_strict_decode", x));
+          DO(base64url_sane_decode(url_cases[x].s, slen1, tmp, &l1));
+          DO(do_compare_testvector(tmp, l1, special_case, sizeof(special_case) - 1, "base64url_sane_decode/strict", x));
+          DO(base64url_decode(url_cases[x].s, slen1, tmp, &l1));
+          DO(do_compare_testvector(tmp, l1, special_case, sizeof(special_case) - 1, "base64url_decode/strict", x));
+       }
+       else if(url_cases[x].flag == relaxed) {
+          DO(base64url_strict_decode(url_cases[x].s, slen1, tmp, &l1) == CRYPT_INVALID_PACKET ? CRYPT_OK : CRYPT_FAIL_TESTVECTOR);
```
sjaeckel: I already thought about it while writing the last base16 testcase... should we add this?

```
diff --git a/tests/common.h b/tests/common.h
index 1ba8820..f63e26b 100644
--- a/tests/common.h
+++ b/tests/common.h
@@ -18,2 +18,3 @@ extern prng_state yarrow_prng;
 #define DOX(x, str) do { fprintf(stderr, "%s - %s:\n", #x, (str)); run_cmd((x), __LINE__, __FILE__, #x, (str)); } while (0)
+#define SHOULD_FAIL(x) do { fprintf(stderr, "%s:\n", #x); run_cmd((x) != CRYPT_OK ? CRYPT_OK : CRYPT_FAIL_TESTVECTOR, __LINE__, __FILE__, #x, NULL); } while (0)
 #else
@@ -21,2 +22,3 @@ extern prng_state yarrow_prng;
 #define DOX(x, str) do { run_cmd((x), __LINE__, __FILE__, #x, (str)); } while (0)
+#define SHOULD_FAIL(x) do { run_cmd((x) != CRYPT_OK ? CRYPT_OK : CRYPT_FAIL_TESTVECTOR, __LINE__, __FILE__, #x, NULL); } while (0)
 #endif
```

doc/crypt.tex
```diff
@@ -6579,6 +6590,9 @@ \subsection{URL--safe 'base64url' encoding}
 
 int base64url_strict_decode(   const char *in,  unsigned long len,
                             unsigned char *out, unsigned long *outlen);
+
+int base64url_sane_decode(   const char *in,  unsigned long len,
```
sjaeckel: should we add the URL variants to the index?

tests/base64_test.c
karel-m: Maybe, but perhaps as `SHOULD_FAIL(x, expected_retval)`

src/misc/base64/base64_decode.c
karel-m: done

doc/crypt.tex
karel-m: done

tests/base64_test.c
sjaeckel: I also thought about `EXPECT(x, e)` but in my opinion that's another use-case.
Most cases where we do negative tests shouldn't care about the error-code returned, it should just be ensured that the result is not OK.

src/misc/base64/base64_decode.c
```diff
@@ -130,7 +145,7 @@ static int _base64_decode_internal(const char *in,  unsigned long inlen,
 
 #if defined(LTC_BASE64)
 /**
-   Relaxed base64 decode a block of memory
+   Dangerously relaxed base64 decode a block of memory
```
sjaeckel: :) well put!

karel-m:
Is it ok?

Issue #216, opened 2017-05-11
Length limits of encauth etc.


sjaeckel:
As @karel-m pointed out in https://github.com/libtom/libtomcrypt/pull/211/files#r115594690 we have a possible issue in `chacha20poly1305`

> Now when I am looking at the implementation the max length of our `chacha20poly1305` is `2^64` bytes; however, in theory `chacha20poly1305` has max length limit `2^70` bytes. The potentioal trouble might occur due to the fact that we do not check for reaching `2^64`.

We should verify the existing implementations regarding such limitations.

The following encauth implementations exist, they should be marked as resolved after their implementation has been verified and/or fixed

- [ ] `ccm`
- [ ] `chacha20poly1305`
- [ ] `eax`
- [x] `gcm` fixed in #129 
- [ ] `ocb`
- [ ] `ocb3`

Should this list contain further elements, e.g. MAC's?

sjaeckel:
CFRG is working on defining those limits in https://datatracker.ietf.org/doc/draft-irtf-cfrg-aead-limits/

Integration will be postponed until the RFC is finished.

PR #369, opened 2018-03-26
Add aesgcm demo

sjaeckel:
This is a follow-up of #342

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,6 +22,8 @@ doc/crypt.pdf
 doc/refman.pdf
 
 # *nix/windows test executables
+aesgcm
+aesgcm.exe
 constants
 constants.exe
 ltcrypt
--- a/demos/aesgcm.c
+++ b/demos/aesgcm.c
@@ -1,0 +1,146 @@
+/* LibTomCrypt, modular cryptographic library -- Tom St Denis
+ *
+ * LibTomCrypt is a library that provides various cryptographic
+ * algorithms in a highly modular and flexible manner.
+ *
+ * The library is free for all purposes without any express
+ * guarantee it works.
+ */
+
+/**
+  @file aesgcm.c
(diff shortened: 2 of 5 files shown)


karel-m:
This is basically fine, I only prefer that ` demos/*.c` are self-containing demos so that a libtomcrypt user can simply cut&paste&tweak the code.  What about putting all the stuff into `demos/aesgcm.c`?

sjaeckel:
> This is basically fine, I only prefer that demos/*.c are self-containing demos so that a libtomcrypt user can simply cut&paste&tweak the code. What about putting all the stuff into demos/aesgcm.c?

as this originates in a PR which tried to introduce those as new API functions I thought it'd make sense to leave them separated.

karel-m:
OK, I'll leave it up to you.

sjaeckel:
I liked your points regarding the output-format and algorithm selection, so I thought it could probably be refactored at one point into a more general solution for all AEAD algorithms :)

karel-m:
> a more general solution for all AEAD algorithms

Keep in mind that not all AEAD algs are suitable for file encryption due to max. size limit. GCM has limit approx. 2^39, ChaCha20Poly1305 has limit AFAIK 2^64 (or a little bit more), the other AEAD algorithms are IMO unusable for large file encryption.

sjaeckel:
> Keep in mind that not all AEAD algs are suitable for file encryption due to max. size limit

algo | max AE
--- | ---
OCB3 | 2^48 blocks (2^55 bits or 4 petabytes)
CCM | 2^64 blocks
GCM | 2^39 - 256 bits
EAX | ?
ChaCha20Poly1305 | 2^64 blocks

true, you shouldn't use them for really large files... and we should think about creating a table with a summary of this, but this reminds me of #216 :)

PR #370, opened 2018-03-27
fix base64_encode to properly add NUL termination

karel-m:
this one is really interesting

### DIFF
--- a/src/misc/base64/base64_encode.c
+++ b/src/misc/base64/base64_encode.c
@@ -70,7 +70,7 @@ static int _base64_encode_internal(const unsigned char *in,  unsigned long inlen
    }
 
    /* append a NULL byte */
-   *p = '\0';
+   *p++ = '\0';
 
    /* return ok */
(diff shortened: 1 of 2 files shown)


sjaeckel:
True, very nice one!

it's somehow a nice mix of "NUL-termination" and "return actual length in length-pointer"

and I like especially the following use-case in the version before this fix...
```
p = malloc(1234);
base64_decode(foo, &bar, ...);
p = realloc(p, bar);
```

karel-m:
The fix will probably break a lot of code.

For consistency it might be better **not** to add terminating NUL byte in all `base(16|32|64)_encode` functions.

karel-m:
Or we can "break" for consistency `base(16|32)_encode` in the same way as `base64_encode`.

karel-m:
I think it would be better to apply #370 instead of this PR.

sjaeckel:
closed in favor of #371

PR #371, opened 2018-03-28
consistent NUL byte handling in baseNN_encode

karel-m:
alternative to #370


src/misc/base16/base16_encode.c
```diff
@@ -49,12 +49,12 @@ int base16_encode(const unsigned char *in,  unsigned long  inlen,
       *outlen = x;
       return CRYPT_BUFFER_OVERFLOW;
    }
-   *outlen = x;
+   x--;
```
sjaeckel: that's exactly why I went with the way it was... `outlen` having two different return values for the same input...

sjaeckel:
can you please add a sentence regarding this behavior in the doc of base16/32/64?

PR #324, opened 2017-11-09
Added missing 'const' qualifier to many functions' parameters 

TrinityCoder:
I have found several functions taking non-cost pointers, while the only thing they do is return size of some keys. 

I believe that functions like these really should take their arguments as `const rsa_key*` to preserve const-correctness and not being forced to write (i. e. in C++) things like

`inline int GetKeySize() const { return ::rsa_get_size(const_cast<rsa_key*>(&key_)); }`

If someone comes across another function(s) which should have `const` parameter but do not, you can write here and we can fix them all at once instead of splitting that into several commits.

Thanks!

**_EDIT:_** As was pointed out by several commenters, there are many more functions in which
a non-const pointer parameter is not used to change the pointed-to object, therefore the pointer can be declared as a const pointer. See below for the full discussion.

PR #381, opened 2018-04-10
add const to the hashes' compress functions

felixdoerre:
Adds a const qualifer to all hashes compress functions and thus removing the necessity for the const-cast in tomcrypt_hash.h


sjaeckel:
This should've been in #324

sjaeckel:
@fperrad what does your linter say?

PR #383, opened 2018-04-28
new fortuna: fix getting current time

karel-m:
Initially there was missing `#include <windows.h>` but it turned out that ns - ms conversion was not correct (fixed and explained in comments).

But there is still one issue related to the fact that `prng->fortuna.wd` is only `unsigned long` (which means 32bit on MS Windows) but we are assigning into it a number of 100ms intervals since 1 Jan 1970.

MSVC warns `conversion from 'ulong64' to 'unsigned long', possible loss of data` here:
```C
unsigned long now = _fortuna_current_time();
```

### DIFF
--- a/src/headers/tomcrypt_custom.h
+++ b/src/headers/tomcrypt_custom.h
@@ -364,6 +364,15 @@
 /* time-based rate limit of the reseeding */
 #define LTC_FORTUNA_RESEED_RATELIMIT_TIMED
 
+/* with non-glibc or glibc 2.17+ prefer clock_gettime over gettimeofday */
+#if defined(__GLIBC__) && defined(__GLIBC_PREREQ)
+#if __GLIBC_PREREQ(2, 17)
+  #define LTC_CLOCK_GETTIME
+#endif
+#elif defined(_POSIX_C_SOURCE) && _POSIX_C_SOURCE >= 200112L
+  #define LTC_CLOCK_GETTIME
+#endif
+
 #else
 
 #ifndef LTC_FORTUNA_WD
--- a/src/prngs/fortuna.c
+++ b/src/prngs/fortuna.c
@@ -93,7 +93,7 @@ static int _fortuna_reseed(prng_state *prng)
    int           err, x;
 
 #ifdef LTC_FORTUNA_RESEED_RATELIMIT_TIMED
-   unsigned long now = _fortuna_current_time();
+   ulong64 now = _fortuna_current_time();
    if (now == prng->fortuna.wd)
       return CRYPT_OK;
 #else
--- a/src/prngs/fortuna.c
+++ b/src/prngs/fortuna.c
@@ -8,6 +8,16 @@
  */
 #include "tomcrypt.h"
 
+#ifdef LTC_FORTUNA_RESEED_RATELIMIT_TIMED
+#if defined(_WIN32)
+  #include <windows.h>
+#elif defined(LTC_CLOCK_GETTIME)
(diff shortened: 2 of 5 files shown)


karel-m:
I have experienced another fortuna related trouble on CentOS 6
```
libtomcrypt.a(fortuna.o): In function `_fortuna_reseed':
fortuna.c:(.text+0xa72): undefined reference to `clock_gettime'
libtomcrypt.a(fortuna.o): In function `fortuna_ready':
fortuna.c:(.text+0x16c6): undefined reference to `clock_gettime'
```

It turns out that using `clock_gettime` requires linking with `-lrt` on Centos 6 (= RHEL6). Therefore I switched to `gettimeofday` (I know it is perhaps not preferred nowadays  but it does not introduce any linking troubles).

karel-m:
@sjaeckel should we change `prng->fortuna.wd` from `unsigned long` to `ulong64` ?

sjaeckel:
> @sjaeckel should we change prng->fortuna.wd from unsigned long to ulong64 ?

why not!

> Therefore I switched to gettimeofday

[man gettimeofday](https://linux.die.net/man/2/gettimeofday) says `The  time  returned  by gettimeofday() is affected by discontinuous jumps in the system time (e.g., if the system administrator manually changes the system time).` ... I'd really prefer to have it as a fall-back, not as the default, especially since that's caused by some soon end-of-life distribution ...

[man clock_gettime](https://linux.die.net/man/3/clock_gettime) says `Link with -lrt (only for glibc versions before 2.17).` therefore I think we can safely add something like this in `tomcrypt_custom.h`:

```
#if defined(_POSIX_C_SOURCE) && _POSIX_C_SOURCE >= 200112L
#if defined(__GLIBC_PREREQ)
#if __GLIBC_PREREQ(2, 17)
#define LTC_CLOCK_GETTIME
#endif /* __GLIBC_PREREQ(2, 17) */
#else
/* here we catch all POSIX-compatible, but non-glibc systems */
#define LTC_CLOCK_GETTIME
#endif /* defined(__linux__) && defined(__GLIBC_PREREQ) */
#endif /* defined(_POSIX_C_SOURCE) && _POSIX_C_SOURCE >= 200112L */
```

probably @rofl0r can verify if that's okay as he's using musl as libc IIRC :)

src/prngs/fortuna.c
```diff
@@ -8,6 +8,12 @@
  */
 #include "tomcrypt.h"
 
+#if defined(_WIN32)
+  #include <windows.h>
+#else
+  #include <sys/time.h>
```
tyzoid: `sys/time.h` is not available on embedded micros, so there should be some `#if defined` to see if the user has defined their own time functions.

src/headers/tomcrypt_custom.h
```diff
@@ -364,6 +364,15 @@
 /* time-based rate limit of the reseeding */
 #define LTC_FORTUNA_RESEED_RATELIMIT_TIMED
 
+/* with non-glibc or glibc 2.17+ prefer clock_gettime over gettimeofday */
+#if defined(__GLIBC__) && defined(__GLIBC_PREREQ)
+#if __GLIBC_PREREQ(2, 17)
+  #define LTC_CLOCK_GETTIME
+#endif
+#else
```
sjaeckel: shouldn't this better be `#elif defined(_POSIX_C_SOURCE) && _POSIX_C_SOURCE >= 200112L`?

karel-m:
@sjaeckel  can I merge this pr?

PR #387, opened 2018-05-03
Add XSalsa20

buggywhip:
Add XSalsa20 to the suite of stream ciphers.

### Checklist
* [X] documentation is added or updated
* [X] tests are added or updated

### DIFF
--- a/doc/crypt.tex
+++ b/doc/crypt.tex
@@ -1301,48 +1301,84 @@ At the end you have to terminate the state:
 err = chacha_done(&st);
 \end{verbatim}
 
-\mysection{Salsa20}
+\mysection{Salsa20 and XSalsa20}
 
-\textit{Salsa20} is the forerunner of the ChaCha stream cipher.  The ChaCha cipher is
-Salsa20 with a few minor tweaks to further improve its strength, and in so doing, increase its
-speed performance by about 5 percent.  Unless you need Salsa20 for some reason, you should
-probably choose ChaCha instead.
+\textit{Salsa20} was Daniel Bernstein's submission to the EU eSTREAM
+competition where a reduced-round version, \textit{Salsa20/12}, was named
+one of the winners.  A third version, \textit{Salsa20/8}, was also evaluated.
+\vspace{1mm}
 
-In April 2008 \textit{Salsa20/12} was named one of the winners in the EU eSTREAM competition.
-Salsa20 was originally submitted by Daniel Bernstein with 20 rounds of strength but the
-12-round reduced-round version was deemed to have sufficient strength and declared a winner.
-Even the 8-round reduced-round version, Salsa20/8, has withstood attack.
+While 20 rounds is the conservative default number of rounds, eSTREAM deemed
+12 rounds to be a decent balance between strength and better performance.
+The 8-round version, while still secure as of this writing, is faster but
+does not enjoy the same margin of safety.  Regardless of the number of rounds,
+\textit{Salsa20} accepts either a 128- or a 256-bit key, a 64-bit IV, and a
+64-bit counter.
+\vspace{1mm}
 
-For more information about Salsa20 see \url{https://en.wikipedia.org/wiki/Salsa20}.
+\textit{XSalsa20} is yet another variant of \textit{Salsa20} designed to accept
+only a 256-bit key and a longer 192-bit nonce, initialization being the only
+difference between \textit{XSalsa20} and \textit{Salsa20}.  Even the
+\textit{salsa20\_state} is the same.  Thereafter, salsa20\_crypt(),
+salsa20\_keystream(), and salsa20\_done() are used unaltered.
+salsa20\_ivctr64() is NOT used with xsalsa20\_setup().
+\vspace{1mm}
 
-Supported key size: 16 or 32 bytes (128 or 256 bits).
+To initialize \textit{Salsa20} for 8, 12, or 20 rounds with a 128- or a
+256-bit key (16 or 32 bytes), a 64-bit IV (8 bytes), and counter (typically
+zero), use:
 
-You can initialize Salsa20 with 64bit \textit{nonce} + 64bit \textit{counter}:
 \begin{verbatim}
 salsa20_state st;
+ulong64 counter = 0;
 err = salsa20_setup(&st, key, key_len, rounds);
-err = salsa20_ivctr64(&st, nonce, 8, initial_64bit_ctr);
+err = salsa20_ivctr64(&st, nonce, 8, counter);
 \end{verbatim}
 
-The \textit{salsa20\_setup} takes the number of rounds as a parameter -- choose 20 (the default)
-if you are not sure.  As always never ever use the same key + nonce pair more than once.
+To initialize \textit{XSalsa20} for the recommended 20 rounds with a 256-bit
+key (32 bytes) and a 192-bit nonce (24 bytes), use: 
 
-For the actual encryption or decryption you have to call:
+\begin{verbatim}
+salsa20_state st;
+err = xsalsa20_setup(&st, key, key_len, nonce, nonce_len, rounds);
+\end{verbatim}
+
+Both \textit{Salsa20} and \textit{XSalsa20} use the following functions.  To
+encrypt or decrypt call:
 \begin{verbatim}
 err = salsa20_crypt(&st, in_buffer, in_len, out_buffer);
 \end{verbatim}
 
-If you just want a random stream of bytes initialize the cipher with a truly random \textit{key}
-(32 bytes), a truly random \textit{nonce} (8 bytes) and zero initial counter. After that you can
-get a stream of pseudo--random bytes via:
+For a random keystream initialize the cipher with a truly random \textit{key}
+and random \textit{nonce} after which you can get a stream of
+pseudo--random bytes via:
 \begin{verbatim}
 err = salsa20_keystream(&st, out_buffer, out_len);
 \end{verbatim}
 
-When finished you should wipe the state:
+Finally, when finished you should wipe the state.
 \begin{verbatim}
 err = salsa20_done(&st);
 \end{verbatim}
+
+For both \textit{Salsa20} and \textit{XSalsa20} rounds must be an even number
+and if set to 0 the default number of rounds, 20, will be used.
+\vspace{1mm}
+
+If you define \textit{LTC_XSALSA20} to include \textit{XSalsa20} in a minimal
+\textit{libtomcrypt} library build, you must also define \textit{LTC_SALSA20}.
+\vspace{1mm}
+
+As always, never ever use the same key + nonce/IV pair more than once.
+\vspace{1mm}
+
+For more information about Salsa20 see
+\url{https://en.wikipedia.org/wiki/Salsa20}.
+\vspace{1mm}
+
+For more information about XSalsa20 see
+\url{https://cr.yp.to/snuffle/xsalsa-20081128.pdf}.
+\vspace{1mm}
 
 \mysection{Sosemanuk}
 
--- a/src/headers/tomcrypt_cipher.h
+++ b/src/headers/tomcrypt_cipher.h
@@ -1031,6 +1031,15 @@ int salsa20_test(void);
 
 #endif /* LTC_SALSA20 */
 
+#ifdef LTC_XSALSA20
+
+int xsalsa20_setup(salsa20_state *st, const unsigned char *key,   unsigned long keylen,
+                                      const unsigned char *nonce, unsigned long noncelen,
+                                      int rounds);
+int xsalsa20_test(void);
+
+#endif /* LTC_XSALSA20 */
+
 #ifdef LTC_SOSEMANUK
 
 typedef struct {
--- a/src/stream/salsa20/salsa20_crypt.c
+++ b/src/stream/salsa20/salsa20_crypt.c
@@ -62,7 +62,7 @@ int salsa20_crypt(salsa20_state *st, const unsigned char *in, unsigned long inle
    LTC_ARGCHK(st        != NULL);
    LTC_ARGCHK(in        != NULL);
    LTC_ARGCHK(out       != NULL);
-   LTC_ARGCHK(st->ivlen == 8);
+   LTC_ARGCHK(st->ivlen == 8 || st->ivlen == 24);
 
    if (st->ksleft > 0) {
       j = MIN(st->ksleft, inlen);
--- a/src/stream/salsa20/xsalsa20_test.c
+++ b/src/stream/salsa20/xsalsa20_test.c
@@ -1,0 +1,94 @@
+/* LibTomCrypt, modular cryptographic library -- Tom St Denis
+ *
+ * LibTomCrypt is a library that provides various cryptographic
+ * algorithms in a highly modular and flexible manner.
+ *
+ * The library is free for all purposes without any express
+ * guarantee it works.
+ */
+
+/* The implementation is based on:
+ * "Extending the Salsa20 nonce", https://cr.yp.to/snuffle/xsalsa-20081128.pdf
+ * "Salsa20 specification", http://cr.yp.to/snuffle/spec.pdf
+ * and salsa20-ref.c version 20051118
+ * Public domain from D. J. Bernstein
+ */
+
+#include "tomcrypt.h"
+
+#ifdef LTC_XSALSA20
+
+#ifdef LTC_SHA256
+int _sha256(unsigned char *hash, const unsigned char *data, const int datalen) {
+   hash_state md;
+   sha256_init(&md);
+   sha256_process(&md, data, datalen);
+   sha256_done(&md, hash);
+   return CRYPT_OK;
+}
+#endif
+
+int xsalsa20_test(void)
+{
+#ifndef LTC_TEST
+   return CRYPT_NOP;
+#else
+
+    /***************************************************************************
+     * verify a round trip:
+     */
+    {
+        const unsigned char key[]   = {0x1b,0x27,0x55,0x64,0x73,0xe9,0x85,0xd4,0x62,0xcd,0x51,0x19,0x7a,0x9a,0x46,0xc7,0x60,0x09,0x54,0x9e,0xac,0x64,0x74,0xf2,0x06,0xc4,0xee,0x08,0x44,0xf6,0x83,0x89};
+        const unsigned char nonce[] = {0x69,0x69,0x6e,0xe9,0x55,0xb6,0x2b,0x73,0xcd,0x62,0xbd,0xa8,0x75,0xfc,0x73,0xd6,0x82,0x19,0xe0,0x03,0x6b,0x7a,0x0b,0x37};
+        const void *msg             = "Kilroy was here!";
+        unsigned char msglen = 17;                  /* includes trailing NULL */
+        int rounds = 20;
+        unsigned char ciphertext[17];
+        unsigned char msg2[17];
+        salsa20_state st;
+        int err;
+
+        if ((err = xsalsa20_setup(&st, key, 32, nonce, 24, rounds)) != CRYPT_OK)  return err;
+        if ((err = salsa20_crypt(&st, msg, msglen, ciphertext))     != CRYPT_OK)  return err;
+        if ((err = salsa20_done(&st))                               != CRYPT_OK)  return err;
+
+        if ((err = xsalsa20_setup(&st, key, 32, nonce, 24, rounds)) != CRYPT_OK)  return err;
+        if ((err = salsa20_crypt(&st, ciphertext, msglen, msg2))    != CRYPT_OK)  return err;
+        if ((err = salsa20_done(&st))                               != CRYPT_OK)  return err;
+
+        if (compare_testvector(msg, msglen, msg2, msglen, "XSALSA20-TV1", 1))  return CRYPT_FAIL_TESTVECTOR;
+    }
+
+#ifdef LTC_SHA256
+   /***************************************************************************
+    * verify correct generation of a keystream
+    */
+   {
+       const unsigned char key[]        = {0x1b,0x27,0x55,0x64,0x73,0xe9,0x85,0xd4,0x62,0xcd,0x51,0x19,0x7a,0x9a,0x46,0xc7,0x60,0x09,0x54,0x9e,0xac,0x64,0x74,0xf2,0x06,0xc4,0xee,0x08,0x44,0xf6,0x83,0x89};
+       const unsigned char nonce[]      = {0x69,0x69,0x6e,0xe9,0x55,0xb6,0x2b,0x73,0xcd,0x62,0xbd,0xa8,0x75,0xfc,0x73,0xd6,0x82,0x19,0xe0,0x03,0x6b,0x7a,0x0b,0x37};
+       const unsigned char expecthash[] = {0x6a,0x60,0x57,0x65,0x27,0xe0,0x00,0x51,0x6d,0xb0,0xda,0x60,0x46,0x20,0xf6,0xd0,0x95,0x65,0x45,0x39,0xf4,0x86,0x83,0x43,0x64,0xdf,0xd9,0x5a,0x6f,0x3f,0xbe,0xb7};
+       int rounds = 20;
+       unsigned char keystream[91101];
+       unsigned long keystreamlen = 91101;
+       unsigned char hash[32];
+       salsa20_state st;
+       int err;
+
+       if ((err = xsalsa20_setup(&st, key, 32, nonce, 24, rounds))   != CRYPT_OK)  return err;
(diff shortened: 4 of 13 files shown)


karel-m:
Hi Larry,

some cosmetics - this:
```
/* ref:         HEAD -> develop */
/* git commit:  af67321bf3cde1a470c679e459ebb8189e38c9bd */
/* commit time: 2018-04-13 09:42:47 +0200 *
```
should be replaced with:
```
/* ref:         $Format:%D$ */
/* git commit:  $Format:%H$ */
/* commit time: $Format:%ai$ */
```

The `xsalsa20_ivctr64` looks to me like a candidate for misuse (with potential security consequences). I'd leave it out till somebody comes with a reasonable use case.

I am not sure whether to introduce `xsalsa20_state`; I think using `salsa20_state` with `xsalsa20_*` functions is ok.

The last doubt I have (for now) is related to `unsigned char *subkey` - is it necessary?

buggywhip:
> should be replaced with:
> 
> /* ref:         $Format:%D$ */
> /* git commit:  $Format:%H$ */
> /* commit time: $Format:%ai$ */
will do
> The xsalsa20_ivctr64 looks to me like a candidate for misuse (with potential security consequences). I'd leave it out till somebody comes with a reasonable use case.
> 
I too struggled with the security implications.  It was that or try to keep parallel APIs.  I won't feel bad if we yank it.
> I am not sure whether to introduce xsalsa20_state; I think using salsa20_state with xsalsa20_*functions is ok.
> 
And you can because xsalsa20_state is defined to salsa20_state in the same way we map xsalsa20_crypt(), _keystream(), and _done() to their salsa20 counterparts.  I suggest, however, keeping xsalsa20_state to reduce confusion.
> The last doubt I have (for now) is related to unsigned char *subkey - is it necessary?  
> 
Not really so long as the user doesn't want a hybrid xsalsa/salsa where the subkey is used more like an ephemeral key.  sha256(key+nonce) can accomplish that but the doubleround function is lightweight and more attractive for smaller devices.  I can argue this both ways; it comes down to what is our objective?  

I will wait a few days.  Others might want to chime in.

-----

I have one additional concern.  The "#define xsalsa20_state salsa20_state" appears to kill getting xsalsa20_state_size fm misc/crypt/crypt_sizes.c.  I still want to play with it some more but python users might just have to remember to use salsa20_state.  ...but all the xsalsa mappings work just fine elsewhere.  Confusing?

karel-m:
Ad `crypt_sizes.c` maybe
```C
#ifdef LTC_SALSA20
    _SZ_STRINGIFY_T(salsa20_state),
#endif
#ifdef LTC_XSALSA20
    _SZ_STRINGIFY_T(xsalsa20_state),
#endif
```

src/stream/xsalsa20/xsalsa20_setup.c
```diff
@@ -0,0 +1,138 @@
...
+   if (subkey != NULL) XMEMCPY(subkey, secondkey, sizeof(secondkey));
+   return CRYPT_OK;
+}
+
+
+#endif
+
+/* ref:         $Format:%D$ */
+/* git commit:  $Format:%H$ */
+/* commit time: $Format:%ai$ */
```
sjaeckel: can you please add a newline on all newly added files

src/headers/tomcrypt_cipher.h
```diff
@@ -1031,6 +1031,21 @@ int salsa20_test(void);
...
 
+#ifdef LTC_XSALSA20
+
+int xsalsa20_setup(salsa20_state *st, const unsigned char *key,   unsigned long keylen,
+                                      const unsigned char *nonce, unsigned long noncelen,
+                                      int rounds, unsigned char *subkey);
+int xsalsa20_test(void);
+
+#define xsalsa20_state     salsa20_state
+#define xsalsa20_ivctr64   salsa20_ivctr64
```
sjaeckel: Do we really have to provide this?
IIUC you could use it, but it's dangerous and therefor: will it even be used?

src/headers/tomcrypt_cipher.h
```diff
@@ -1031,6 +1031,21 @@ int salsa20_test(void);
 
 #endif /* LTC_SALSA20 */
 
+#ifdef LTC_XSALSA20
+
+int xsalsa20_setup(salsa20_state *st, const unsigned char *key,   unsigned long keylen,
+                                      const unsigned char *nonce, unsigned long noncelen,
+                                      int rounds, unsigned char *subkey);
```
sjaeckel: shouldn't there be a `unsigned long *subkeylen`? :)

src/stream/xsalsa20/xsalsa20_setup.c
```diff
@@ -0,0 +1,138 @@
...
+       salsa20_setup(st, secondkey, 32, rounds);
+       salsa20_ivctr64(st, &nonce[16], 8, 0);
+   */
+   st->rounds = rounds;
+   st->ksleft = 0;
+   st->ivlen = 8;           /* set switch to say nonce/IV has been loaded */
+
+   /* copy out subkey if not NULL */
+   if (subkey != NULL) XMEMCPY(subkey, secondkey, sizeof(secondkey));
+   return CRYPT_OK;
```
sjaeckel: this is missing a 
```
#ifdef LTC_CLEAN_STACK 
zeromem(secondkey, ...
zeromem(x, ...
...
```

src/stream/xsalsa20/xsalsa20_setup.c
```diff
@@ -0,0 +1,138 @@
...
+   /* ...this...
+       salsa20_setup(st, secondkey, 32, rounds);
+       salsa20_ivctr64(st, &nonce[16], 8, 0);
+   */
+   st->rounds = rounds;
+   st->ksleft = 0;
+   st->ivlen = 8;           /* set switch to say nonce/IV has been loaded */
+
+   /* copy out subkey if not NULL */
+   if (subkey != NULL) XMEMCPY(subkey, secondkey, sizeof(secondkey));
```
sjaeckel: Instead of copying why not work with the provided buffer if given?
```
unsigned char secondkey_buf[32], *secondkey = secondkey_buf;
...
if (subkey != NULL) secondkey = subkey;
...
#ifdef LTC_CLEAN_STACK
zeromem(secondkey_buf, ...
```

src/stream/xsalsa20/xsalsa20_setup.c
```diff
@@ -0,0 +1,138 @@
...
+   @param subkey    [out] The subkey, NULL if not want a copy
+   @return CRYPT_OK if successful
+*/
+int xsalsa20_setup(salsa20_state *st,
+                   const unsigned char *key, unsigned long keylen,
+                   const unsigned char *nonce,    unsigned long noncelen,
+                   int rounds, unsigned char *subkey) {
+   unsigned char secondkey[32];
+   ulong32       x[64];                     /* input to & output fm doubleround */
+   int sti[] = {0, 5, 10, 15, 6, 7, 8, 9};  /* indices used to build subkey fm x */
```
sjaeckel: `const int`

src/stream/xsalsa20/xsalsa20_test.c
```diff
@@ -0,0 +1,109 @@
...
+{
+#ifndef LTC_TEST
+   return CRYPT_NOP;
+#else
+
+    /***************************************************************************
+     * verify a round trip:
+     */
+    {
+        unsigned char key[]   = {0x1b,0x27,0x55,0x64,0x73,0xe9,0x85,0xd4,0x62,0xcd,0x51,0x19,0x7a,0x9a,0x46,0xc7,0x60,0x09,0x54,0x9e,0xac,0x64,0x74,0xf2,0x06,0xc4,0xee,0x08,0x44,0xf6,0x83,0x89};
```
sjaeckel: `const`

src/stream/xsalsa20/xsalsa20_test.c
```diff
@@ -0,0 +1,109 @@
...
+        unsigned char nonce[] = {0x69,0x69,0x6e,0xe9,0x55,0xb6,0x2b,0x73,0xcd,0x62,0xbd,0xa8,0x75,0xfc,0x73,0xd6,0x82,0x19,0xe0,0x03,0x6b,0x7a,0x0b,0x37};
```
sjaeckel: `const`

src/stream/xsalsa20/xsalsa20_test.c
```diff
@@ -0,0 +1,109 @@
...
+
+        if (compare_testvector(msg, msglen, msg2, msglen, "XSALSA20-TV1", 1))  return CRYPT_FAIL_TESTVECTOR;
+    }
+
+   /***************************************************************************
+      verify correct generation of subkey
+      ref: stream3.c/out in nacl-20110221/tests
+   */
+   {
+       unsigned char key[]      = {0x1b,0x27,0x55,0x64,0x73,0xe9,0x85,0xd4,0x62,0xcd,0x51,0x19,0x7a,0x9a,0x46,0xc7,0x60,0x09,0x54,0x9e,0xac,0x64,0x74,0xf2,0x06,0xc4,0xee,0x08,0x44,0xf6,0x83,0x89};
```
sjaeckel: `const`

src/stream/xsalsa20/xsalsa20_test.c
```diff
@@ -0,0 +1,109 @@
...
+       unsigned char nonce[]    = {0x69,0x69,0x6e,0xe9,0x55,0xb6,0x2b,0x73,0xcd,0x62,0xbd,0xa8,0x75,0xfc,0x73,0xd6,0x82,0x19,0xe0,0x03,0x6b,0x7a,0x0b,0x37};
```
sjaeckel: `const`

src/stream/xsalsa20/xsalsa20_test.c
```diff
@@ -0,0 +1,109 @@
...
+       unsigned char subkey[32] = {0};
+       unsigned char expect[]   = {0xdc,0x90,0x8d,0xda,0x0b,0x93,0x44,0xa9,0x53,0x62,0x9b,0x73,0x38,0x20,0x77,0x88,0x80,0xf3,0xce,0xb4,0x21,0xbb,0x61,0xb9,0x1c,0xbd,0x4c,0x3e,0x66,0x25,0x6c,0xe4};
```
sjaeckel: `const`

src/stream/xsalsa20/xsalsa20_test.c
```diff
@@ -0,0 +1,109 @@
...
+
+       if ((err = xsalsa20_setup(&st, key, 32, nonce, 24, rounds, subkey)) != CRYPT_OK)  return err;
+       if (compare_testvector(subkey, sizeof(subkey), expect, sizeof(expect), "XSALSA20-TV2", 1))               return CRYPT_FAIL_TESTVECTOR;
+   }
+
+   /***************************************************************************
+      verify correct generation of a keystream
+   */
+   {
+       unsigned char key[]        = {0x1b,0x27,0x55,0x64,0x73,0xe9,0x85,0xd4,0x62,0xcd,0x51,0x19,0x7a,0x9a,0x46,0xc7,0x60,0x09,0x54,0x9e,0xac,0x64,0x74,0xf2,0x06,0xc4,0xee,0x08,0x44,0xf6,0x83,0x89};
```
sjaeckel: `const` ... and all the others :D

src/stream/xsalsa20/xsalsa20_setup.c
buggywhip: done

src/headers/tomcrypt_cipher.h
buggywhip: removed

src/headers/tomcrypt_cipher.h
buggywhip: added

src/stream/xsalsa20/xsalsa20_setup.c
buggywhip: implemented

src/stream/xsalsa20/xsalsa20_setup.c
buggywhip: implemented

src/stream/xsalsa20/xsalsa20_setup.c
buggywhip: done

src/stream/xsalsa20/xsalsa20_test.c
buggywhip: done

src/stream/xsalsa20/xsalsa20_test.c
buggywhip: done

src/stream/xsalsa20/xsalsa20_test.c
buggywhip: done

src/stream/xsalsa20/xsalsa20_test.c
buggywhip: done

src/stream/xsalsa20/xsalsa20_test.c
buggywhip: done

src/headers/tomcrypt_cipher.h
```diff
@@ -1031,6 +1031,20 @@ int salsa20_test(void);
 
 #endif /* LTC_SALSA20 */
 
+#ifdef LTC_XSALSA20
+
+int xsalsa20_setup(salsa20_state *st, const unsigned char *key,   unsigned long keylen,
```
karel-m: shouldn't we use `xsalsa20_state` here?

src/stream/salsa20/salsa20_ivctr64.c
```diff
@@ -15,7 +15,7 @@
 
 #include "tomcrypt.h"
 
-#ifdef LTC_SALSA20
+#if defined(LTC_SALSA20) || defined(LTC_XSALSA20)
```
karel-m: this can be IMO reverted to `#ifdef LTC_SALSA20`

src/stream/xsalsa20/xsalsa20_setup.c
```diff
@@ -0,0 +1,145 @@
...
+   @param key       The secret key
+   @param keylen    The length of the secret key, must be 32 (octets)
+   @param nonce     The nonce
+   @param noncelen  The length of the nonce, must be 24 (octets)
+   @param rounds    Number of rounds (must be evenly divisible by 2, default is 20)
+   @param subkey    [out] The subkey, NULL if not want a copy
+   @param subkeylen [out] number of bytes copied, NULL if not want a copy
+   @return CRYPT_OK if successful
+*/
+int xsalsa20_setup(salsa20_state *st, const unsigned char *key, unsigned long keylen,
```
karel-m: `salsa20_state` > `xsalsa20_state`

karel-m:
I still do not see a use case for `subkey`. How can one reuse the value?

I am for: `xsalsa20_setup(&st, key, key_len, nonce, nonce_len, rounds)`

If there is a demand for `subkey` we can introduce `xsalsa20_setup_ex(&st, key, key_len, nonce, nonce_len, rounds, &subkey, &subkeylen)`

src/stream/xsalsa20/xsalsa20_setup.c
```diff
@@ -0,0 +1,145 @@
...
+   const int sti[] = {0, 5, 10, 15, 6, 7, 8, 9};  /* indices used to build subkey fm x */
+   int i = 0;
+
+   LTC_ARGCHK(st        != NULL);
+   LTC_ARGCHK(secondkey != NULL);
+   LTC_ARGCHK(key       != NULL);
+   LTC_ARGCHK(keylen    == 32);
+   LTC_ARGCHK(nonce     != NULL);
+   LTC_ARGCHK(noncelen  == 24);
+   LTC_ARGCHK((subkey != NULL && subkeylen != NULL)  ||
```
sjaeckel:
```
LTC_ARGCHK(((subkey == NULL) && (subkeylen == NULL)) ||
           ((subkey != NULL) && (subkeylen != NULL) && (*subkeylen >= 32)));
```

src/stream/xsalsa20/xsalsa20_setup.c
sjaeckel: not sure though if we should remove the `xsalsa20_state` define and go with only `salsa20_state*` arguments... it'd be cleaner I guess...

src/headers/tomcrypt_cipher.h
sjaeckel: see my comment below

src/stream/salsa20/salsa20_ivctr64.c
sjaeckel: :+1:

sjaeckel:
> I still do not see a use case for subkey. How can one reuse the value?

very good point!

> I am for: xsalsa20_setup(&st, key, key_len, nonce, nonce_len, rounds)

:+1:

buggywhip:
> On 6May, 2018, at 12:56 PM, Steffen Jaeckel <notifications@github.com> wrote:
> I still do not see a use case for subkey. How can one reuse the value?
> 
> very good point!
> 
I can see using the "hashed" subkey in various situations, none of which couldn't be satisfied with another hash function.  The only redeeming value might be the doubleround's space and time.  

I doubt we would ever see a request for two reasons: XSalsa's low adoption and library "shoppers" will simply move on.  (My usage was predominately prototyping; between OpenSSL and LTC I'd carve out the pieces I needed and augment with custom code.)  

An alternative would be to implement Karel's xsalsa20_setup_ex(&st, key, key_len, nonce, nonce_len, rounds, &subkey, &subkeylen) now.  ...but as a "shopper" I'd be much more interested in XChaCha20 with all the bells and whistles.
> I am for: xsalsa20_setup(&st, key, key_len, nonce, nonce_len, rounds)
> 
> 👍
> 
All that said...  Clearly this will be the most frequent use case and is *much* cleaner.  

Done.  (offline I'll keep the existing code should we change our mind)

doc/crypt.tex
```diff
@@ -1344,6 +1344,48 @@ \chapter{Stream Ciphers}
...
 \end{verbatim}
 
+\mysection{XSalsa20}
+
+\textit{XSalsa20} is a derivation of \textit{Salsa20} that supports a 192-bit
+nonce.
+
+xsalsa20_setup() is the only call needed to completely initialize the state.
+The key size must be 32 bytes (256 bits) and nonce size must be 24 bytes (192
+bits).  Rounds must be an even number and if set to 0 it will be changed to 20.
```
sjaeckel: > and if set to 0 it will be changed to 20.

and if set to 0 the default number of rounds, 20, will be used.

src/stream/salsa20/salsa20_ivctr64.c
sjaeckel: I'd also prefer to add something like the following in the "cleanup" section of `tomcrypt_custom.h` where all the other similar cases are handled

```
#if defined(LTC_XSALSA20) && !defined(LTC_SALSA20)
#error LTC_XSALSA20 requires LTC_SALSA20
#endif
```

then all the changes to `src/stream/salsa20/salsa20_*.c` could be reverted

src/stream/salsa20/salsa20_keystream.c
```diff
@@ -27,6 +27,7 @@
 int salsa20_keystream(salsa20_state *st, unsigned char *out, unsigned long outlen)
 {
    if (outlen == 0) return CRYPT_OK; /* nothing to do */
+   LTC_ARGCHK(st->ivlen == 8 || st->ivlen == 24);
```
karel-m: there is IMO no need for this check here

karel-m:
Providing the check with `#error LTC_XSALSA20 requires LTC_SALSA20` in `tomcrypt_custom.h` I think we do not need any of these:

```C
-#ifdef LTC_SALSA20
+#if defined(LTC_SALSA20) || defined(LTC_XSALSA20)
```

src/stream/salsa20/salsa20_ivctr64.c
```diff
@@ -30,7 +30,7 @@ int salsa20_ivctr64(salsa20_state *st, const unsigned char *iv, unsigned long iv
    LTC_ARGCHK(st != NULL);
    LTC_ARGCHK(iv != NULL);
    /* Salsa20: 64-bit IV (nonce) + 64-bit counter */
-   LTC_ARGCHK(ivlen == 8);
+   LTC_ARGCHK(ivlen != 24);
```
karel-m: Are you sure here? What's the idea behind this change?

buggywhip: I want to prevent invoking salsa20_ivctr64() after calling xsalsa20_setup().  ...unnecessary, and if called the uniqueness of the 24-byte nonce could be compromised.

src/headers/tomcrypt_cipher.h
```diff
@@ -1012,7 +1012,7 @@ int chacha_test(void);
 
 #endif /* LTC_CHACHA */
 
-#ifdef LTC_SALSA20
+#if defined(LTC_SALSA20) || defined(LTC_XSALSA20)
```
karel-m: This can IMO stay as it used to be: `#ifdef LTC_SALSA20`

src/stream/salsa20/salsa20_crypt.c
```diff
@@ -15,7 +15,7 @@
 
 #include "tomcrypt.h"
 
-#ifdef LTC_SALSA20
+#if defined(LTC_SALSA20) || defined(LTC_XSALSA20)
```
karel-m: This can IMO stay as it used to be: `#ifdef LTC_SALSA20`

src/stream/salsa20/salsa20_done.c
```diff
@@ -9,7 +9,7 @@
 
 #include "tomcrypt.h"
 
-#ifdef LTC_SALSA20
+#if defined(LTC_SALSA20) || defined(LTC_XSALSA20)
```
karel-m: This can IMO stay as it used to be: `#ifdef LTC_SALSA20`

src/stream/salsa20/salsa20_keystream.c
```diff
@@ -15,7 +15,7 @@
 
 #include "tomcrypt.h"
 
-#ifdef LTC_SALSA20
+#if defined(LTC_SALSA20) || defined(LTC_XSALSA20)
```
karel-m: This can IMO stay as it used to be: `#ifdef LTC_SALSA20`

karel-m:
Larry, there is one more `#if defined(LTC_SALSA20) || defined(LTC_XSALSA20)` in `salsa20_ivctr64.c` that can be IMO reverted.

buggywhip:
> On 28May, 2018, at 11:49 PM, karel-m <notifications@github.com> wrote:
> 
> @karel-m approved this pull request.
> 
> Looks good to me. Just please squash it into 2 commits: 1/ update makefiles + 2/ xsalsa20
> 
Thanks, will do after one more doc update to add one sentence:

"If you define \textit{LTC_XSalsa20} to include \textit{XSalsa20} in a minimal
\textit{libtomcrypt} library build, you must also define \textit{LTC_Salsa20}."

I'll push it tomorrow.

doc/crypt.tex
```diff
@@ -1301,49 +1301,85 @@ \chapter{Stream Ciphers}
...
+Finally, when finished you should wipe the state.
 \begin{verbatim}
 err = salsa20_done(&st);
 \end{verbatim}
 
+For both \textit{Salsa20} and \textit{XSalsa20} rounds must be an even number
+and if set to 0 the default number of rounds, 20, will be used.
+\vspace{1mm}
+
+If you define \textit{LTC_XSalsa20} to include \textit{XSalsa20} in a minimal
```
sjaeckel: `LTC_XSALSA20`

doc/crypt.tex
```diff
@@ -1301,49 +1301,85 @@ \chapter{Stream Ciphers}
...
+\textit{libtomcrypt} library build, you must also define \textit{LTC_Salsa20}.
```
sjaeckel: `LTC_SALSA20`

src/stream/salsa20/xsalsa20_test.c
```diff
@@ -0,0 +1,89 @@
...
+ * Public domain from D. J. Bernstein
+ */
+
+#include "tomcrypt.h"
+
+#ifdef LTC_XSALSA20
+
+int _sha256(unsigned char *hash, const unsigned char *data, const int datalen) {
+   hash_state md;
+   sha256_init(&md);
```
sjaeckel: pretty paranoid, but this would fail to build if someone wants to build w/o sha256

I think it'd be better to use the `hash_memory()` function instead of this `_sha256()` ... but I just realized then again we'd have a dependency on the `LTC_HASH_HELPERS` ... btw. how's this handled in all the other tests? :D 

Simply put this function and the second test-case in a `#ifdef LTC_SHA256` please

PR #390, opened 2018-05-09
Fixes and improvements

sjaeckel:
Some fixes and improvements I came across


src/headers/tomcrypt_pk.h
```diff
@@ -775,10 +775,10 @@ int der_printable_value_decode(int v);
 
 /* UTF-8 */
 #if (defined(SIZE_MAX) || __STDC_VERSION__ >= 199901L || defined(WCHAR_MAX) || defined(__WCHAR_MAX__) || defined(_WCHAR_T) || defined(_WCHAR_T_DEFINED) || defined (__WCHAR_TYPE__)) && !defined(LTC_NO_WCHAR)
-#include <wchar.h>
 #if defined(__WCHAR_MAX__)
 #define LTC_WCHAR_MAX __WCHAR_MAX__
-#elif defined(WCHAR_MAX)
```
karel-m: This part should be IMO like this (indented for clarity)
```C
   #if defined(__WCHAR_MAX__)
     #define LTC_WCHAR_MAX __WCHAR_MAX__
   #else
     #include <wchar.h>
     #if defined(WCHAR_MAX)
       #define LTC_WCHAR_MAX WCHAR_MAX
     #endif
   #endif
```

sjaeckel: Indentation okay, but the additional `#if defined(WCHAR_MAX)` can be left out in my opinion as of http://pubs.opengroup.org/onlinepubs/009695399/basedefs/wchar.h.html
or do you know a `wchar.h` that doesn't define `WCHAR_MAX`?

karel-m: OK, it looks like you are right.

PR #392, opened 2018-05-19
makefile.shared: Respect LIBTOOL.

orbea:
I am using the libtool implementation called slibtool which seems to work just fine with libtomcrypt.

https://git.midipix.org/cgit.cgi/slibtool
https://git.midipix.org/cgit.cgi/slibtool/plain/README

he problem is that the standard way to use slibtool is to export `LIBTOOL` in `MAKEFLAGS`.
```
export MAKEFLAGS='LIBTOOL=slibtool'
```
Which will work with almost all autotools builds. However libtomcrypt doesn't respect this variable and wants `LT` instead.

If the makefile.shared can use `LIBTOOL` instead of `LT` this would make it a lot easier for anyone that doesn't want the default libtool.

Please see PR https://github.com/libtom/libtommath/pull/114 for more information.

On a side note, libtomcrypt is a good example of how slibtool is faster than libtool, with an AMD FX(tm)-6350 Six-Core cpu using `-j6` and `make -f makefile.shared`.

GNU libtool
```
real	0m43.242s
user	2m9.735s
sys	0m38.435s
```
slibtool
```
real	0m12.180s
user	0m42.789s
sys	0m10.406s
```

### DIFF
--- a/makefile.shared
+++ b/makefile.shared
@@ -16,19 +16,19 @@
 
 PLATFORM := $(shell uname | sed -e 's/_.*//')
 
-ifndef LT
+ifndef LIBTOOL
   ifeq ($(PLATFORM), Darwin)
-    LT:=glibtool
+    LIBTOOL:=glibtool
   else
-    LT:=libtool
+    LIBTOOL:=libtool
   endif
 endif
 ifeq ($(PLATFORM), CYGWIN)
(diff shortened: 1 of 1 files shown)


rofl0r:
hah. proof that the libtool abomination only slows down your build by a huge factor and adds possible issues, while providing zero benefit.

PR #395, opened 2018-05-31
Add a new header file for private functions

sjaeckel:
This adds `tomcrypt_private.h` with all the declarations that were protected by `LTC_SOURCE` before.

Additionally this makes `blake2[sb]_init()` private.
I'd accompany that change by marking both functions as deprecated in the upcoming 1.18.2 release.
[Edit]: I just realized we can't simply deprecate them w/o having our source complain as well on compilation... I'll have to think about it a bit more...


src/headers/tomcrypt_private.h
```diff
@@ -0,0 +1,328 @@
...
+
+/* tomcrypt_mac.h */
+
+int ocb3_int_ntz(unsigned long x);
+void ocb3_int_xor_blocks(unsigned char *out, const unsigned char *block_a, const unsigned char *block_b, unsigned long block_len);
+
+
+/* tomcrypt_math.h */
+
+#if !defined(DESC_DEF_ONLY) && defined(LTC_SOURCE)
```
karel-m: drop ` && defined(LTC_SOURCE)`

src/headers/tomcrypt_private.h
```diff
@@ -0,0 +1,328 @@
...
+   PKA_EC,
+   PKA_EC_PRIMEF
+};
+
+/*
+ * Internal Types
+ */
+
+typedef struct Oid {
+    const unsigned long *OID;
```
karel-m: `unsigned long OID[16];`

makefile.unix
```diff
@@ -297,7 +299,7 @@ clean:
 install: $(LIBMAIN_S) $(HEADERS)
```
karel-m: maybe `install: $(LIBMAIN_S) $(HEADERS_PUB)`

karel-m:
consider also changing `makefile_include.mk` like this:
```diff
- install -p -m 644 $(HEADERS) $(DESTDIR)$(INCPATH)
+ install -p -m 644 $(HEADERS_PUB) $(DESTDIR)$(INCPATH)
```

karel-m:
One more `LTC_SOURCE` to eliminate is in `doc/Doxyfile`
```
PREDEFINED             = LTC_SOURCE \
                         LTM_DESC \
                         GMP_DESC \
                         TFM_DESC
```

src/headers/tomcrypt_private.h
sjaeckel: done

src/headers/tomcrypt_private.h
sjaeckel: oops, that one slipped in from my pkcs8 refactor... :)

makefile.unix
sjaeckel: I removed it entirely, as it's already an indirect dependency of LIBMAIN_S

sjaeckel:
> I'd accompany that change by marking both functions as deprecated in the upcoming 1.18.2 release.
> [Edit]: I just realized we can't simply deprecate them w/o having our source complain as well on compilation... I'll have to think about it a bit more...

regarding the deprecation I think we shouldn't do that in 1.18.2 but only in develop (as already done for the math initializers). To prevent warnings in our internal code I'd leave the `blake2[sb]_init()` API as deprecated public and make it call a newly added `_blake2[sb]_init()` which contain the implementation. Okay?

ksherlock:
I'm probably the only one, but I use `blake2s_init()` function from user code to generate hashes with non-standard digest sizes (128, 160, 224, and 256 are the most common but it can generate digests of 8-512 bits. There's also a configurable key parameter).

sjaeckel:
Oh okay, then we'll leave them in the public API!

PR #398, opened 2018-06-02
Improve base16 api

sjaeckel:
As said in the commit-message:

```
probably we want to add more options in the future

I could think of support for some options of `xxd` resp. `hexdump`
```

e.g. the little-endian encoding of 2**n byte values

### DIFF
--- a/doc/crypt.tex
+++ b/doc/crypt.tex
@@ -6710,12 +6710,12 @@ To encode a binary string in base16 call:
 \begin{verbatim}
 int base16_encode(const unsigned char *in,  unsigned long  inlen,
                                  char *out, unsigned long *outlen,
-                                 int  caps);
+                        unsigned int   options);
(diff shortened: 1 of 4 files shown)


doc/crypt.tex
```diff
@@ -6710,12 +6710,12 @@ \subsection{URL--safe 'base64url' encoding}
 \begin{verbatim}
 int base16_encode(const unsigned char *in,  unsigned long  inlen,
                                  char *out, unsigned long *outlen,
-                                 int  caps);
+                        unsigned long  options);
```
karel-m: I would prefer `unsigned int` which is 32bit pretty everywhere. With `unsigned long` there might be a trouble that it is only 32bit on Windows/x86 but 64bit on many other platforms.

karel-m:
I have merged this one so that I can continue with merging #400

PR #406, opened 2018-06-15
Sosemanuk - remove what is effectively a NOP

buggywhip:
prior return stmt was a failed attempt to initialize the remaining bytes of the state which is also negated by the policy of enforcing a call to sosemanuk_setiv() before calling sosemanuk_crypt().

### Checklist


src/stream/sosemanuk/sosemanuk.c
```diff
@@ -322,7 +322,7 @@ int sosemanuk_setup(sosemanuk_state *ss, const unsigned char *key, unsigned long
      * Initialize with a zero-value iv to ensure state is correct in the
```
sjaeckel: This comment is lying now ;)

Issue #407, opened 2018-06-15
Fix for CVE-2018-12437?


stapelberg:
mitre: https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2018-12437
paper: https://www.nccgroup.trust/us/our-research/technical-advisory-return-of-the-hidden-number-problem/

Is a fix for libtomcrypt available? Thanks.

sjaeckel:
Working on it :)

sjaeckel:
Let's discuss this in #408

sjaeckel:
Btw we're going to have another bugfix release this week.

PR #408, opened 2018-06-17
ecc_sign_hash blinding CVE-2018-12437

karel-m:
Fix for #407 aka CVE-2018-12437

### DIFF
--- a/src/pk/ecc/ecc_sign_hash.c
+++ b/src/pk/ecc/ecc_sign_hash.c
@@ -21,7 +21,7 @@ static int _ecc_sign_hash(const unsigned char *in,  unsigned long inlen,
                                 prng_state *prng, int wprng, const ecc_key *key, int sigformat)
 {
    ecc_key       pubkey;
-   void          *r, *s, *e, *p;
+   void          *r, *s, *e, *p, *b;
    int           err, max_iterations = LTC_PK_MAX_RETRIES;
    unsigned long pbits, pbytes, i, shift_right;
    unsigned char ch, buf[MAXBLOCKSIZE];
@@ -37,7 +37,7 @@ static int _ecc_sign_hash(const unsigned char *in,  unsigned long inlen,
    }
 
    /* init the bignums */
-   if ((err = mp_init_multi(&r, &s, &e, NULL)) != CRYPT_OK) {
+   if ((err = mp_init_multi(&r, &s, &e, &b, NULL)) != CRYPT_OK) {
       return err;
    }
 
@@ -72,12 +72,15 @@ static int _ecc_sign_hash(const unsigned char *in,  unsigned long inlen,
       if (mp_iszero(r) == LTC_MP_YES) {
          ecc_free(&pubkey);
       } else {
+         if ((err = rand_bn_upto(b, p, prng, wprng)) != CRYPT_OK)          { goto error; } /* b = blinding value */
          /* find s = (e + xr)/k */
-         if ((err = mp_invmod(pubkey.k, p, pubkey.k)) != CRYPT_OK)         { goto error; } /* k = 1/k */
+         if ((err = mp_mulmod(pubkey.k, b, p, pubkey.k)) != CRYPT_OK)      { goto error; } /* k = kb */
+         if ((err = mp_invmod(pubkey.k, p, pubkey.k)) != CRYPT_OK)         { goto error; } /* k = 1/kb */
          if ((err = mp_mulmod(key->k, r, p, s)) != CRYPT_OK)               { goto error; } /* s = xr */
-         if ((err = mp_add(e, s, s)) != CRYPT_OK)                          { goto error; } /* s = e +  xr */
-         if ((err = mp_mod(s, p, s)) != CRYPT_OK)                          { goto error; } /* s = e +  xr */
-         if ((err = mp_mulmod(s, pubkey.k, p, s)) != CRYPT_OK)             { goto error; } /* s = (e + xr)/k */
+         if ((err = mp_mulmod(pubkey.k, s, p, s)) != CRYPT_OK)             { goto error; } /* s = xr/kb */
+         if ((err = mp_mulmod(pubkey.k, e, p, e)) != CRYPT_OK)             { goto error; } /* e = e/kb */
+         if ((err = mp_add(e, s, s)) != CRYPT_OK)                          { goto error; } /* s = e/kb + xr/kb */
+         if ((err = mp_mulmod(s, b, p, s)) != CRYPT_OK)                    { goto error; } /* s = b(e/kb + xr/kb) = (e + xr)/k */
          ecc_free(&pubkey);
          if (mp_iszero(s) == LTC_MP_NO) {
             break;
@@ -111,7 +114,7 @@ static int _ecc_sign_hash(const unsigned char *in,  unsigned long inlen,
 error:
    ecc_free(&pubkey);
 errnokey:
-   mp_clear_multi(r, s, e, NULL);
+   mp_clear_multi(r, s, e, b, NULL);
    return err;
 }
 


rofl0r:
out of curiosity, could you describe what exactly your fix does, @karel-m ?

karel-m:
Sorry, here are some details from off-github communication.

Original disclosure:

```
NCC Group Security Advisory
https://www.nccgroup.trust
--------------------------------------------

Side Channel Based ECDSA Key Extraction in Libtomcrypt

Vendor: Libtomcrypt
Systems Affected: ECDSA signature generation 
Author: Keegan Ryan <keegan.ryan@nccgroup.trust>
Risk: Medium

Summary:
--------
An attacker capable of triggering signatures and mounting a side channel
attack against a victim can extract an ECDSA key in a practical amount of time.
The full attack can take only a few minutes on a modern system. We also suggest
countermeasures that can be applied to DSA and ECDSA signature generation to
protect against this attack.

Other cryptographic libraries are similarly affected, so eventual public
disclosure of this issue and release of a corresponding paper should be
coordinated with these parties as well.

Due to the number of affected libraries, we have not developed a working proof
of concept for Libtomcrypt, but we have developed a working proof
of concept for a library with a similar code pattern.

Location:
---------
When computing a DSA or ECDSA signature, the signer calculates
    r = x[kG]
    s = k^-1(m + rx)
over a finite field. Our attack targets non-constant time operations in
the addition step
    m + (rx).

Impact:
-------
An attacker can use this attack to extract an ECDSA private key from a victim,
and the requirements for the attacker are realistic. Consider a HTTPS server
running in a VM in a cloud environment. An unprivileged running in a colocated
VM can initiate a connection to the server to trigger a signature and use a
side channel to leak the desired information about the signature.

Details:
--------
This attack applies a known cryptanalytic technique in a new and unique way and
shows that the needed information is exposed by Libtomcrypt via well-known side channel
techniques.

In particular, we use the leaked information to construct an instance of the
Hidden Number Problem (HNP). The HNP has been used before in cryptanalytic
side channel attacks, but typically these attacks seek to leak bits of k during
the exponentiation step of computing r. Various methods exist to solve the HNP
efficiently, including a method based on highly efficient lattice reduction
algorithms.

The HNP problem can be stated as follows:
Given finite field of prime order q, unknown x \in F_q, and a series of
multipliers t_i and approximations u_i to x*t_i, recover the value of x.
Approximations can be written as
    |(t_i * x mod q) - u_i| < q * 2 ^ -l.
That is, when u_i is close to x*t_i, the absolute value on the left hand side is
small. If we can formulate our cryptanalytic attack in this way, we can use
HNP solvers to recover the value of x.

We are targeting the step of adding m to the product rx. In Libtomcrypt's ECDSA signing
code, rx has been reduced modulo q (although this is not the case in the DSA
code). We wish for our side channel to leak whether m + (rx mod q) is less than q or
not (here we are referring to addition within Z, not within Fq).

This is possible because of the non-constant time implementation of _ecc_sign_hash.
This function contains the following code snippet:

  /* find s = (e + xr)/k */
  if ((err = mp_invmod(pubkey.k, p, pubkey.k)) != CRYPT_OK)         { goto error; } /* k = 1/k */
  if ((err = mp_mulmod(key->k, r, p, s)) != CRYPT_OK)               { goto error; } /* s = xr */
  if ((err = mp_add(e, s, s)) != CRYPT_OK)                          { goto error; } /* s = e +  xr */
  if ((err = mp_mod(s, p, s)) != CRYPT_OK)                          { goto error; } /* s = e +  xr */
  if ((err = mp_mulmod(s, pubkey.k, p, s)) != CRYPT_OK)             { goto error; } /* s = (e + xr)/k */

It is clear that at each step the result is reduced modulo the curve order. This is done
with a call to mp_mod, which depends on the implementation but in libtommath does not
return in constant time.

Assume that the side channel indicates that e + xr > p. Thus we know (in our terms):
    m + (rx mod q) < q
	(rx mod q) < q - m
	|(rx mod q) - (q - m) / 2| < (q - m) / 2
The relationship to the HNP is obvious. Under our assumptions, the attacker knows
the value of m and the signature (r, s), so the attacker knows the HNP multiplier
r, approximation (q-m)/2, and bound (q-m)/2. After collecting several approximations
of this type, the attacker can use an HNP solver to calculate x, the private key.

This is a simplified view of the attack. In reality, there are limitations to the
bounds that HNP solvers can tolerate, the side channel may have errors, and it is
possible that the attacker can observe m but not control it. These obstacles can be
overcome with slight modifications to the procedure, but the fundamental idea is
the same.

The cryptanalysis has been implemented in a separate proof of concept and another
similar library has been successfully targeted using Flush+Reload to leak the desired
side-channel information. In that PoC, we demonstrate an unprivileged attacker recovering
a 256 bit ECDSA private key from a colocated process. Due to the number of libraries
with the same leaky code pattern, we have not developed a proof of concept for all
libraries, but we believe it would be straightforward to adapt the attack to compromise
this library.

Recommendation:
---------------
There are two approaches to mitigating this issues, and using both will help defend
against this sort of issue. Though both address slightly different risks, we believe
the first countermeasure is effective against a greater range of threats and its
implementation should be prioritized.

The first approach is to make modular arithmetic for computing the final value of s
branchless. Thus by ensuring there is no difference in code executed when
m + (rx mod q) < q, many side channels will be useless for recovering this data.
This will protect against currently known variations of Prime+Probe, Flush+Reload,
Spectre v1, and more. Since the time needed to compute s is dwarfed by the exponentiation
step in computing r, we believe that making such code constant time is unlikely to
incur a significant performance penalty. Additionally, although this work only
targets the modular addition, it is possible that the same information could be leaked
through modular multiplication, and making all operations involved in computing s
constant time would give greater assurance that the needed information is not being leaked.

The second approach is to use a blinding value for the sum. Even with constant time code
in place, it is theoretically possible that a side channel (such as power analysis under
a Hamming distance model) could distinguish between constant time code subtracting 0 from
the sum and subtracting q. To address this possibility, we recommend introducing a random
blinding multiplier b \in [1, q-1]. Instead of computing the reduced sum:

    (m + (r * x mod q) mod q)    and revealing if    m + (r * x mod q) < q
    
we propose computing

    (b^-1 * ((b * m mod q) + (b * r * x mod q)) mod q)    and revealing if    (bm mod q) + (brx mod q) < q.

Under these assumptions, the countermeasure is provably secure (although it does not
address if the side channel reveals different information).

Finally, we note that the countermeasures should also be applied to the DSA signature
generation algorithm. Since the sign function in dsa_sign_hash.c does not reduce the product r*x
before adding m, the DSA implementation is not vulnerable to this current attack. However,
this implementation may be vulnerable to future variants, and it would be better to protect
it at the same time as the ECDSA function.

Vendor Communication:
---------------------


About NCC Group:
--------------------
NCC Group is a global expert in cyber security and risk mitigation, 
working with businesses to protect their brand, value and reputation 
against the ever-evolving threat landscape. Our Security Consulting 
services leverage our extensive knowledge of current security 
vulnerabilities, penetration testing techniques and software development 
best practices to enable organizations to secure their applications 
against ever-present threats. At NCC Group we can boast unrivaled talent 
and recognized world-class security expertise. Bringing together best in 
class security consultancies iSEC Partners, Intrepidus Group, Matasano, 
NCC Group and NGS we have created one of the largest, most respected 
security consultancies in the world.
```

My patch is (I believe) trying to implement `The second approach is to use a blinding value for the sum.` mentioned above.

karel-m:
And to make it clear what the change is about - here is the old (vulnerable) style:
```C
/* find s = (e + xr)/k */
if ((err = mp_invmod(pubkey.k, p, pubkey.k)) != CRYPT_OK)    { goto error; } /* k = 1/k */
if ((err = mp_mulmod(key->k, r, p, s)) != CRYPT_OK)          { goto error; } /* s = xr */
if ((err = mp_add(e, s, s)) != CRYPT_OK)                     { goto error; } /* s = e +  xr */
if ((err = mp_mod(s, p, s)) != CRYPT_OK)                     { goto error; } /* s = e +  xr */
if ((err = mp_mulmod(s, pubkey.k, p, s)) != CRYPT_OK)        { goto error; } /* s = (e + xr)/k */

```

And the new (hopefully fixed) style:
```C
if ((err = rand_bn_upto(b, p, prng, wprng)) != CRYPT_OK)     { goto error; } /* b = blinding value */
/* find s = (e + xr)/k */
if ((err = mp_mulmod(pubkey.k, b, p, pubkey.k)) != CRYPT_OK) { goto error; } /* k = kb */
if ((err = mp_invmod(pubkey.k, p, pubkey.k)) != CRYPT_OK)    { goto error; } /* k = 1/kb */
if ((err = mp_mulmod(key->k, r, p, s)) != CRYPT_OK)          { goto error; } /* s = xr */
if ((err = mp_mulmod(pubkey.k, s, p, s)) != CRYPT_OK)        { goto error; } /* s = xr/kb */
if ((err = mp_mulmod(pubkey.k, e, p, e)) != CRYPT_OK)        { goto error; } /* e = e/kb */
if ((err = mp_add(e, s, s)) != CRYPT_OK)                     { goto error; } /* s = e/kb + xr/kb */
if ((err = mp_mulmod(s, b, p, s)) != CRYPT_OK)               { goto error; } /* s = b(e/kb + xr/kb) = (e + xr)/k */
```

rofl0r:
thanks! :+1:

DerDakon:
I wonder if this is the same as [CVE-2018-0495](https://nvd.nist.gov/vuln/detail/CVE-2018-0495), which is the entry that e.g. libgcrypt used and that is linked from [nccgroup](https://www.nccgroup.com/us/our-research/technical-advisory-return-of-the-hidden-number-problem/).

PR #422, opened 2018-07-03
fix clang-tidy warning: google-readability-braces-around-statements

karel-m:
This is the last patch related to my clang-tidy testing. After this patch the script `.ci/clang-tidy.sh` should not throw any warning (at least with clang-tidy 6.0.0).

It is related to `braces-around-statements` for which the clang-tidy has two plugins: `google-readability-braces-around-statements` and `readability-braces-around-statements`.

I have chosen less restrictive `google-readability-braces-around-statements`.

It allows one line statements without braces:

```C
if (something) some_func(1,2,3);
```

But does not allow:

```C
if (something)
   some_func(1,2,3);
```

Which should be rewritten as:

```C
if (something) {
   some_func(1,2,3);
}
```

### DIFF
--- a/src/encauth/ocb3/ocb3_decrypt_last.c
+++ b/src/encauth/ocb3/ocb3_decrypt_last.c
@@ -66,10 +66,11 @@ int ocb3_decrypt_last(ocb3_state *ocb, const unsigned char *ct, unsigned long ct
      /* Checksum_* = Checksum_m xor (P_* || 1 || zeros(127-bitlen(P_*))) */
      ocb3_int_xor_blocks(ocb->checksum, ocb->checksum, pt+full_blocks_len, last_block_len);
      for(x=last_block_len; x<ocb->block_len; x++) {
-       if (x == last_block_len)
+       if (x == last_block_len) {
          ocb->checksum[x] ^= 0x80;
-       else
+       } else {
          ocb->checksum[x] ^= 0x00;
+       }
      }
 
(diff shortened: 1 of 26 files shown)


src/encauth/ocb3/ocb3_decrypt_last.c
```diff
@@ -66,10 +66,12 @@ int ocb3_decrypt_last(ocb3_state *ocb, const unsigned char *ct, unsigned long ct
      /* Checksum_* = Checksum_m xor (P_* || 1 || zeros(127-bitlen(P_*))) */
      ocb3_int_xor_blocks(ocb->checksum, ocb->checksum, pt+full_blocks_len, last_block_len);
      for(x=last_block_len; x<ocb->block_len; x++) {
-       if (x == last_block_len)
+       if (x == last_block_len) {
          ocb->checksum[x] ^= 0x80;
-       else
+       }
```
sjaeckel: I somehow prefer if it's written `} else {` is that also valid? :) 

if yes, can you please change it?

sjaeckel (approved):

src/encauth/ocb3/ocb3_decrypt_last.c
karel-m: done

rofl0r:
advantage of
```c
if (something)
   some_func(1,2,3);
```
is that you can put a breakpoint on the second line

karel-m:
I have changed:
```
if (something)
   some_func(1,2,3);
```

into:
```
if (something) {
   some_func(1,2,3);
}
```
So this advantage still applies.

PR #393, opened 2018-05-23
ECC doc update

karel-m:
Add doc for changes introduced in 24c0eb84f989de52c53f7129704dd7322296f3be
* [x] new `ecc_get_curve(const char* name_or_oid, const ltc_ecc_curve** cu)`
* [x] new `ecc_set_dp(const ltc_ecc_curve *cu, ecc_key *key)`
* [x] new `ecc_generate_key(prng_state *prng, int wprng, ecc_key *key)`
* [x] new `ecc_set_key(const unsigned char *in, unsigned long inlen, int type, ecc_key *key)`
* [x] new `ecc_get_key(unsigned char *out, unsigned long *outlen, int type, const ecc_key *key)`
* [x] new `ecc_get_oid_str(char *out, unsigned long *outlen, const ecc_key *key)`
* [x] obsolete: `ltc_ecc_set_type` (replaced by `ltc_ecc_curve`)

From the last part of ECC changes:
* [x] new `ecc_export_openssl`
* [x] new `ecc_import_openssl`
* [x] new `ecc_import_x509`

It will take some time.

PR #423, opened 2018-07-03
rename ecc_set_dp >> ecc_set_curve + ecc_get_curve >> ecc_find_curve

karel-m:
as discussed in #393 

Now we have similarly named:
* `int  ecc_get_curve(const char* name_or_oid, const ltc_ecc_curve** cu);`
* `int  ecc_set_curve(const ltc_ecc_curve *cu, ecc_key *key);`

I am not sure whether `ecc_get_curve` is proper here, maybe `ecc_find_curve` would be better?

### DIFF
--- a/libtomcrypt_VS2008.vcproj
+++ b/libtomcrypt_VS2008.vcproj
@@ -2283,11 +2283,11 @@
 					>
 				</File>
 				<File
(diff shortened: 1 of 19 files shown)


sjaeckel:
find is indeed better :+1:

sjaeckel (approved):

PR #404, opened 2018-06-11
Improved pkcs8 support

karel-m:
This is a re-created #403 which got closed and for some mysterious reasons cannot be re-opened. We should share the new ECC+pcks8 with RSA as the old `rsa_import_pkcs8` does not support password protected keys.

* [x] ecc_import_pkcs8
* [x] rsa_import_pkcs8
* [ ] documentation
* [x] further tests

PR #426, opened 2018-07-05
Feature/pkcs12

sjaeckel:

add pkcs#12 API factored-out from #404 

*[Edit]* unfortunately I couldn't find any testvectors, does someone have some?


karel-m:
I would keep these two functions in `tomcrypt_private.h`. At least for now.

sjaeckel:
> I would keep these two functions in tomcrypt_private.h. At least for now.

good idea! done

karel-m (approved):

PR #429, opened 2018-07-10
Remove oid_st

### LINKED ISSUE #362: Re-work OID handling
Follow-up of OID handling of #309

sjaeckel:

This is the first PR in order to resolve #362 

In this PR we get rid of the only internally used `struct oid_st`.


src/pk/asn1/oid/pk_get_oid.c
```diff
@@ -0,0 +1,45 @@
...
+
+static const pka_oid oids[] = {
+                               { PKA_RSA, "1.2.840.113549.1.1.1" },
+                               { PKA_DSA, "1.2.840.10040.4.1" },
+                               { PKA_EC, "1.2.840.10045.2.1" },
+                               { PKA_EC_PRIMEF, "1.2.840.10045.1.1" },
+};
+
+/*
+   Returns the OID of the public key algorithm.
```
karel-m: cosmetics: just use something more general than "public key algorithm" as especially in the new PKCS#8 implementation we will used for OIDs of many other kinds.

karel-m (approved):

src/pk/asn1/oid/pk_get_oid.c
sjaeckel: very true

sjaeckel: better? fb7b8799cd0258741307eb8c89cfd85deab5ba5b

PR #431, opened 2018-07-30
Streams - add single call crypt functions

buggywhip:
Added single-call crypt functions to each of the stream ciphers using the \<cipher\>_memory() naming convention.

### Checklist

### DIFF
--- a/src/stream/chacha/chacha_test.c
+++ b/src/stream/chacha/chacha_test.c
@@ -40,25 +40,38 @@ int chacha_test(void)
    int err;
 
    len = strlen(pt);
-   /* crypt piece by piece */
+
+   /* crypt piece by piece - using chacha_ivctr32() */
    if ((err = chacha_setup(&st, k, sizeof(k), 20)) != CRYPT_OK)                            return err;
    if ((err = chacha_ivctr32(&st, n, sizeof(n), 1)) != CRYPT_OK)                           return err;
-   if ((err = chacha_crypt(&st, (unsigned char*)pt,      35,       out)) != CRYPT_OK)      return err;
+   if ((err = chacha_crypt(&st, (unsigned char*)pt,      35,       out     )) != CRYPT_OK) return err;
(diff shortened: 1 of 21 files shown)


src/stream/chacha/chacha_test.c
```diff
@@ -59,6 +59,9 @@ int chacha_test(void)
    if ((err = chacha_ivctr64(&st, n + 4, sizeof(n) - 4, 1)) != CRYPT_OK)                   return err;
    if ((err = chacha_crypt(&st, (unsigned char*)pt, len, out)) != CRYPT_OK)                return err;
    if (compare_testvector(out, len, ct, sizeof(ct), "CHACHA-TV3", 1))                      return CRYPT_FAIL_TESTVECTOR;
+   /* crypt in a single call using 64-bit counter with a value of 1 */
```
sjaeckel: Can you please add a second testcase that also hits `chacha_ivctr32()`.

sjaeckel (changes requested):
LGTM :+1:

sjaeckel:
I took the freedom to rebase everything on top of develop and clean up the duplicate commits :)

buggywhip:
> On 10Sep, 2018, at 2:37 AM, Steffen Jaeckel <notifications@github.com> wrote:
> 
> I took the freedom to rebase everything on top of develop and clean up the duplicate commits :)
> 
Thanks.  ...but the git log graph looks funny.  ...and certainly not like it was after my rebase/squash.  ???  May I assume it is fixable?  ...how?

sjaeckel (commented):
I think we should include these `xx_memory()` implementations in the single c files where there's only a single file for now!
@karel-m what do you think?
@buggywhip sorry, I didn't see that before!

karel-m:
> I think we should include these `xx_memory()` implementations in the single c files where there's only a single file for now!

I am for keeping `xx_memory()` **always** in a separate `.c` file. It has obvious advantages (smaller code) when using static linking.

sjaeckel:
Good point I didn't think about!
So let's keep it like that.

PR #433, opened 2018-08-14
makefile.shared: Support rlibtool.

orbea:
When building libtomcrypt with rlibtool instead of libtool it will fail when rlibtool fails to parse the generated libtool which does not exist.

Since rlibtool should be the default choice for most slibtool users in the future this patch will use slibtool-shared instead of rlibtool which will correctly build the shared library.

This could also help build the shared library on additional targets and hosts where the stock libtool does not have shared libraries enabled.

### DIFF
--- a/makefile.shared
+++ b/makefile.shared
@@ -16,19 +16,24 @@
 
 PLATFORM := $(shell uname | sed -e 's/_.*//')
 
+ifeq ($(LIBTOOL),rlibtool)
+  TGTLIBTOOL:=slibtool-shared
+endif
+
 ifndef LIBTOOL
   ifeq ($(PLATFORM), Darwin)
     LIBTOOL:=glibtool
   else
     LIBTOOL:=libtool
   endif
+  TGTLIBTOOL=$(LIBTOOL)
 endif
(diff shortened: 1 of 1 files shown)


orbea:
If this is acceptable I can make a PR for libtommath too.

The slibtool devs added the rlibtool symlink to address the issue of being able to dynamically determine if shared libraries or static archives should be built, but since libtommath doesn't use autotools it has no way of determining this and will fail requiring the older slibtool-shared symlink.

rofl0r:
couldn't we just remove use of libtool completely ? i can't think of a single reason why it should be used.

orbea:
I have no complaints with that.

sjaeckel:
If we can replace those {r,s,}libtool features like so versioning etc. sure, why not.
Feel free to open a PR and we'll have a look

sjaeckel:
> If we can replace those {r,s,}libtool features like so versioning etc. sure, why not.
> Feel free to open a PR and we'll have a look

As it doesn't seem like someone will open that PR I'll merge this one.

FYI I've rebased the commit on top of develop as it was based on master.

PR #436, opened 2018-09-11
Implement ecc_recover_key to recover public key from hash+signature

rmw42:
This patch implements ECC public key recovery from hash+signature.

There are only a small number of possible public keys, typically 2, corresponding to the 'r' value in an elliptic-curve signature. These can be represented by a single parity bit - as with X9.62 compressed keys - in addition to the signature, rather than pre-sharing the public key. This is used in Bitcoin/Ethereum to avoid an explicit representation of the public key, saving space in messages.

ecc_recover_key() takes a signature and message hash and an additional "recovery ID" value to specify which of the possible public keys to recover. Parameters are very similar to ecc_verify_hash(), but the public key is the output rather than an input.

See https://crypto.stackexchange.com/questions/18105/how-does-recovering-the-public-key-from-an-ecdsa-signature-work for the underlying maths.

Cheers,
---
Russ

### Checklist

### DIFF
--- a/doc/crypt.tex
+++ b/doc/crypt.tex
@@ -5549,6 +5549,22 @@ int ecc_sign_hash_rfc7518(const unsigned char *in,
 This function creates the same ECDSA signature as \textit{ecc\_sign\_hash} only the output format is different.
 The format follows \url{https://tools.ietf.org/html/rfc7518#section-3.4}, sometimes it is also called plain signature.
 
+\index{ecc\_sign\_hash_ex()}
+\begin{verbatim}
+int ecc_sign_hash_ex(const unsigned char *in,
+                           unsigned long  inlen,
+                           unsigned char *out,
+                           unsigned long *outlen,
+                              prng_state *prng,
+                                     int  wprng,
+                      ecc_signature_type  sigformat,
+                                     int *recid,
+                                 ecc_key *key);
+\end{verbatim}
+
+This function is an extended version of the ECDSA signature in \textit{ecc\_sign\_hash}, but with a choice of output formats
+and an optional output of the recovery ID for use with \textit{ecc\_recover\_key}.
+
 \subsection{Signature Verification}
 \index{ecc\_verify\_hash()}
 \begin{verbatim}
@@ -5579,6 +5595,19 @@ int ecc_verify_hash_rfc7518(const unsigned char *sig,
 This function validate the ECDSA signature as \textit{ecc\_verify\_hash} only the signature input format
 follows \url{https://tools.ietf.org/html/rfc7518#section-3.4}.
 
+\index{ecc\_verify\_hash_ex()}
+\begin{verbatim}
+int ecc_verify_hash_ex(const unsigned char *sig,
+                             unsigned long  siglen,
+                       const unsigned char *hash,
+                             unsigned long  hashlen,
+                        ecc_signature_type  sigformat,
+                                       int *stat,
+                                   ecc_key *key);
+\end{verbatim}
+
+This function validates an ECDSA signature as \textit{ecc\_verify\_hash} but with a choice of signature formats.
+
 {\bf BEWARE:} With ECC if you try to sign a hash that is bigger than your ECC key you can run into problems. The math
 will still work, and in effect the signature will still work.  With ECC keys the strength of the signature is limited
 by the size of the hash, or the size of the key, whichever is smaller.  For example, if you sign with SHA256 and a
@@ -5593,19 +5622,46 @@ int ecc_recover_key(const unsigned char *sig,
                     const unsigned char *hash,
                           unsigned long  hashlen,
                                     int  recid,
+                     ecc_signature_type  sigformat,
                                 ecc_key *key);
 \end{verbatim}
 
 This function will recover (a) public key from the ECDSA signature in the array pointed to by \textit{sig} of length \textit{siglen} octets, the message digest
 pointed to by the array \textit{hash} of length \textit{hashlen}, and the recovery id \textit{recid}. It will store the recovered
-key into \textit{key} and return CRYPT_OK if recovery succeeds, or an error if recovery fails.
-This is for compatibility with the (v,r,s) signatures used in Bitcoin/Ethereum, where public keys are not explicitly
-shared, only the parity of the public key. For curves like secp256k1, recid will take values of 0 or 1, corresponding to
-the parity of the public key's y coordinate. For curves like secp112r2, with a cofactor of 4, values 0..7 are possible,
-with the low bit corresponding to the parity and the higher bits specifying the public key's x coordinate's  multiple
+key into \textit{key} and return CRYPT\_OK if recovery succeeds, or an error if recovery fails.
+This is for compatibility with the (v,r,s) signatures used in Ethereum, where public keys are not explicitly shared,
+only the parity of the public key. For curves like secp256k1, recid will take values of 0 or 1, corresponding to the
+parity of the public key's y coordinate. For curves like secp112r2, with a cofactor of 4, values 0..7 are possible,
+with the low bit corresponding to the parity and the higher bits specifying the public key's x coordinate's multiple
 of the curve's order.
-The function \textit{ecc\_recover\_key} implements signature format according to X9.62 ECDSA, and the output is compliant for GF(p) curves.
+If the signature format contains the recovery id (currently only \textit{LTC\_ECCSIG\_ETH27}), \textit{recid} can be -1
+which signals that the recovery id from the signature blob should be used. This means an application does not need to
+extract the recovery id from such a signature in order to use this function.
+The function \textit{ecc\_recover\_key} implements multiple signature formats, and the output is compliant for GF(p) curves.
 
+\subsection{Signature Formats}
+The following signature formats are suported:
+
+\begin{figure}[hpbt]
+\index{Signature Formats}
+\begin{small}
+\begin{center}
+\begin{tabular}{|l|l|}
+\hline \textbf{sigformat} & \textbf{description} \\
+\hline LTC\_ECCSIG\_ANSIX962 & ASN.1 encoded, ANSI X9.62 \\
+\hline LTC\_ECCSIG\_RFC7518 & raw R, S values as defined in RFC7518 \\
+\hline LTC\_ECCSIG\_ETH27 & raw R, S, V values (V has 27 added) \\
+\hline
+\end{tabular}
+\end{center}
+\end{small}
+\caption{Signature Formats}
+\label{fig:sigformat}
+\end{figure}
+
+The \textit{LTC\_ECCSIG\_ETH27} format is based on the Ethereum Yellow Paper, see \url{https://github.com/ethereum/yellowpaper}
+(Appendix F). However, convention allows the use of v=0,1 as equivalent to v=27,28 and both are accepted by
+\textit{ecc\_recover\_key}.
 
 \mysection{Shared Secret (ECDH)}
 To construct a Diffie-Hellman shared secret with a private and public ECC key, use the following function:
--- a/doc/crypt.tex
+++ b/doc/crypt.tex
@@ -5585,6 +5585,28 @@ by the size of the hash, or the size of the key, whichever is smaller.  For exam
 P--192 key, you have in effect 96--bits of security. The library will not warn you if you make this mistake, so it
 is important to check yourself before using the signatures.
 
+\subsection{Public Key Recovery}
+\index{ecc\_recover\_key()}
+\begin{verbatim}
+int ecc_recover_key(const unsigned char *sig,
+                          unsigned long  siglen,
+                    const unsigned char *hash,
+                          unsigned long  hashlen,
+                                    int  recid,
+                                ecc_key *key);
+\end{verbatim}
+
+This function will recover (a) public key from the ECDSA signature in the array pointed to by \textit{sig} of length \textit{siglen} octets, the message digest
+pointed to by the array \textit{hash} of length \textit{hashlen}, and the recovery id \textit{recid}. It will store the recovered
+key into \textit{key} and return CRYPT_OK if recovery succeeds, or an error if recovery fails.
+This is for compatibility with the (v,r,s) signatures used in Bitcoin/Ethereum, where public keys are not explicitly
+shared, only the parity of the public key. For curves like secp256k1, recid will take values of 0 or 1, corresponding to
+the parity of the public key's y coordinate. For curves like secp112r2, with a cofactor of 4, values 0..7 are possible,
+with the low bit corresponding to the parity and the higher bits specifying the public key's x coordinate's  multiple
+of the curve's order.
+The function \textit{ecc\_recover\_key} implements signature format according to X9.62 ECDSA, and the output is compliant for GF(p) curves.
+
+
 \mysection{Shared Secret (ECDH)}
 To construct a Diffie-Hellman shared secret with a private and public ECC key, use the following function:
 \index{ecc\_shared\_secret()}
--- a/src/headers/tomcrypt_pk.h
+++ b/src/headers/tomcrypt_pk.h
@@ -244,6 +244,16 @@ typedef struct {
     void *k;
 } ecc_key;
 
+/** Formats of ECC signatures */
+typedef enum ecc_signature_type_ {
+   /* ASN.1 encoded, ANSI X9.62 */
+   LTC_ECCSIG_ANSIX962   = 0x0,
+   /* raw R, S values */
+   LTC_ECCSIG_RFC7518    = 0x1,
+   /* raw R, S, V (+27) values */
+   LTC_ECCSIG_ETH27      = 0x2
+} ecc_signature_type;
+
 /** the ECC params provided */
 extern const ltc_ecc_curve ltc_ecc_curves[];
 
@@ -286,25 +296,30 @@ int  ecc_decrypt_key(const unsigned char *in,  unsigned long  inlen,
                            unsigned char *out, unsigned long *outlen,
                            const ecc_key *key);
 
-int ecc_sign_hash_rfc7518(const unsigned char *in,  unsigned long inlen,
-                                unsigned char *out, unsigned long *outlen,
-                                prng_state *prng, int wprng, const ecc_key *key);
+#define ecc_sign_hash_rfc7518(in_, inlen_, out_, outlen_, prng_, wprng_, key_) \
+   ecc_sign_hash_ex(in_, inlen_, out_, outlen_, prng_, wprng_, LTC_ECCSIG_RFC7518, NULL, key_)
 
-int  ecc_sign_hash(const unsigned char *in,  unsigned long inlen,
-                         unsigned char *out, unsigned long *outlen,
-                         prng_state *prng, int wprng, const ecc_key *key);
+#define ecc_sign_hash(in_, inlen_, out_, outlen_, prng_, wprng_, key_) \
+   ecc_sign_hash_ex(in_, inlen_, out_, outlen_, prng_, wprng_, LTC_ECCSIG_ANSIX962, NULL, key_)
 
-int ecc_verify_hash_rfc7518(const unsigned char *sig,  unsigned long siglen,
-                            const unsigned char *hash, unsigned long hashlen,
-                            int *stat, const ecc_key *key);
+#define ecc_verify_hash_rfc7518(sig_, siglen_, hash_, hashlen_, stat_, key_) \
+   ecc_verify_hash_ex(sig_, siglen_, hash_, hashlen_, LTC_ECCSIG_RFC7518, stat_, key_)
 
-int  ecc_verify_hash(const unsigned char *sig,  unsigned long siglen,
-                     const unsigned char *hash, unsigned long hashlen,
-                     int *stat, const ecc_key *key);
+#define ecc_verify_hash(sig_, siglen_, hash_, hashlen_, stat_, key_) \
+   ecc_verify_hash_ex(sig_, siglen_, hash_, hashlen_, LTC_ECCSIG_ANSIX962, stat_, key_)
+
+int  ecc_sign_hash_ex(const unsigned char *in,  unsigned long inlen,
+                            unsigned char *out, unsigned long *outlen,
+                            prng_state *prng, int wprng, ecc_signature_type sigformat,
+                            int *recid, const ecc_key *key);
+
+int  ecc_verify_hash_ex(const unsigned char *sig,  unsigned long siglen,
+                        const unsigned char *hash, unsigned long hashlen,
+                        ecc_signature_type sigformat, int *stat, const ecc_key *key);
 
 int  ecc_recover_key(const unsigned char *sig,  unsigned long siglen,
                      const unsigned char *hash, unsigned long hashlen,
-                     int recid, ecc_key *key);
+                     int recid, ecc_signature_type sigformat, ecc_key *key);
 
 #endif
 
--- a/src/headers/tomcrypt_pk.h
+++ b/src/headers/tomcrypt_pk.h
@@ -302,6 +302,10 @@ int  ecc_verify_hash(const unsigned char *sig,  unsigned long siglen,
                      const unsigned char *hash, unsigned long hashlen,
                      int *stat, const ecc_key *key);
 
+int  ecc_recover_key(const unsigned char *sig,  unsigned long siglen,
+                     const unsigned char *hash, unsigned long hashlen,
+                     int recid, ecc_key *key);
+
 #endif
 
 #ifdef LTC_MDSA
--- a/src/pk/ecc/ecc_recover_key.c
+++ b/src/pk/ecc/ecc_recover_key.c
@@ -18,9 +18,20 @@
   ECC Crypto, Russ Williams
 */
 
-static int _ecc_recover_key(const unsigned char *sig,  unsigned long siglen,
-                            const unsigned char *hash, unsigned long hashlen,
-                            int recid, ecc_key *key)
+/**
+   Recover ECC public key from signature and hash
+   @param sig         The signature to verify
+   @param siglen      The length of the signature (octets)
+   @param hash        The hash (message digest) that was signed
+   @param hashlen     The length of the hash (octets)
+   @param recid       The recovery ID ("v"), can be -1 if signature contains it
+   @param sigformat   The format of the signature (ecc_signature_type)
+   @param key         The recovered public ECC key
+   @return CRYPT_OK if successful (even if the signature is not valid)
+*/
+int ecc_recover_key(const unsigned char *sig,  unsigned long siglen,
+                    const unsigned char *hash, unsigned long hashlen,
+                    int recid, ecc_signature_type sigformat, ecc_key *key)
 {
    ecc_point     *mG = NULL, *mQ = NULL, *mR = NULL;
    void          *p, *m, *a, *b;
@@ -62,11 +73,56 @@ static int _ecc_recover_key(const unsigned char *sig,  unsigned long siglen,
       goto error;
    }
 
-   /* Only ASN.1 format signatures supported for now */
-   if ((err = der_decode_sequence_multi_ex(sig, siglen, LTC_DER_SEQ_SEQUENCE | LTC_DER_SEQ_STRICT,
+   if (sigformat == LTC_ECCSIG_ANSIX962) {
+      /* ANSI X9.62 format - ASN.1 encoded SEQUENCE{ INTEGER(r), INTEGER(s) }  */
+      if ((err = der_decode_sequence_multi_ex(sig, siglen, LTC_DER_SEQ_SEQUENCE | LTC_DER_SEQ_STRICT,
                                      LTC_ASN1_INTEGER, 1UL, r,
                                      LTC_ASN1_INTEGER, 1UL, s,
                                      LTC_ASN1_EOL, 0UL, NULL)) != CRYPT_OK)                             { goto error; }
+   }
+   else if (sigformat == LTC_ECCSIG_RFC7518) {
+      /* RFC7518 format - raw (r,s) */
+      i = mp_unsigned_bin_size(key->dp.order);
+      if (siglen != (2*i)) {
+         err = CRYPT_INVALID_PACKET;
+         goto error;
+      }
+      if ((err = mp_read_unsigned_bin(r, (unsigned char *)sig,   i)) != CRYPT_OK)                       { goto error; }
+      if ((err = mp_read_unsigned_bin(s, (unsigned char *)sig+i, i)) != CRYPT_OK)                       { goto error; }
+   }
+   else if (sigformat == LTC_ECCSIG_ETH27) {
+      /* Ethereum (v,r,s) format */
+      if (key->dp.oidlen != 5   || key->dp.oid[0] != 1 || key->dp.oid[1] != 3 ||
+          key->dp.oid[2] != 132 || key->dp.oid[3] != 0 || key->dp.oid[4] != 10) {
+         /* Only valid for secp256k1 - OID 1.3.132.0.10 */
+         err = CRYPT_ERROR; goto error;
+      }
+      if (siglen != 65) { /* Only secp256k1 curves use this format, so must be 65 bytes long */
+         err = CRYPT_INVALID_PACKET;
+         goto error;
+      }
+      i = (unsigned long)sig[64];
+      if ((i>=27) && (i<31)) i -= 27; /* Ethereum adds 27 to recovery ID */
+      if (recid >= 0 && ((unsigned long)recid != i)) {
+         /* Recovery ID specified, but doesn't match signature */
+         err = CRYPT_INVALID_PACKET;
+         goto error;
+      }
+      recid = i;
+      if ((err = mp_read_unsigned_bin(r, (unsigned char *)sig,  32)) != CRYPT_OK)                       { goto error; }
+      if ((err = mp_read_unsigned_bin(s, (unsigned char *)sig+32, 32)) != CRYPT_OK)                     { goto error; }
+   }
+   else {
+      /* Unknown signature format */
+      err = CRYPT_ERROR;
+      goto error;
+   }
+
+   if (recid < 0 || (unsigned long)recid >= 2*(key->dp.cofactor+1)) {
+      /* Recovery ID is out of range, reject it */
+      err = CRYPT_INVALID_ARG;
+      goto error;
+   }
 
    /* check for zero */
    if (mp_cmp_d(r, 0) != LTC_MP_GT || mp_cmp_d(s, 0) != LTC_MP_GT ||
@@ -181,25 +237,8 @@ error:
    if (mR != NULL) ltc_ecc_del_point(mR);
    if (mQ != NULL) ltc_ecc_del_point(mQ);
    if (mG != NULL) ltc_ecc_del_point(mG);
-   mp_clear_multi(r, s, v, w, t1, t2, u1, u2, v1, v2, e, x, y, a_plus3, NULL);
+   mp_clear_multi(a_plus3, y, x, e, v2, v1, u2, u1, t2, t1, w, v, s, r, NULL);
    return err;
-}
-
-/**
-   Recover ECC public key from signature and hash
-   @param sig         The signature to verify
-   @param siglen      The length of the signature (octets)
-   @param hash        The hash (message digest) that was signed
-   @param hashlen     The length of the hash (octets)
-   @param recid       0 or 1 to select parity ("v")
-   @param key         The recovered public ECC key
-   @return CRYPT_OK if successful (even if the signature is not valid)
-*/
-int ecc_recover_key(const unsigned char *sig,  unsigned long siglen,
-                    const unsigned char *hash, unsigned long hashlen,
-                    int recid, ecc_key *key)
-{
-   return _ecc_recover_key(sig, siglen, hash, hashlen, recid, key);
 }
 
 #endif
--- a/src/pk/ecc/ecc_recover_key.c
+++ b/src/pk/ecc/ecc_recover_key.c
@@ -1,0 +1,210 @@
+/* LibTomCrypt, modular cryptographic library -- Tom St Denis
+ *
+ * LibTomCrypt is a library that provides various cryptographic
+ * algorithms in a highly modular and flexible manner.
+ *
+ * The library is free for all purposes without any express
+ * guarantee it works.
+ */
+
+#include "tomcrypt_private.h"
+
+#ifdef LTC_MECC
+
+#ifdef LTC_ECC_SHAMIR
+
+/**
+  @file ecc_recover_key.c
+  ECC Crypto, Russ Williams
+*/
+
+static int _ecc_recover_key(const unsigned char *sig,  unsigned long siglen,
+                            const unsigned char *hash, unsigned long hashlen,
+                            int recid, ecc_key *key)
+{
+   ecc_point     *mG = NULL, *mQ = NULL, *mR = NULL;
+   void          *p, *m, *a, *b;
+   void          *r, *s, *v, *w, *t1, *t2, *u1, *u2, *v1, *v2, *e, *x, *y, *a_plus3;
+   void          *mu = NULL, *ma = NULL;
+   void          *mp = NULL;
+   int           err;
+   unsigned long pbits, pbytes, i, shift_right;
+   unsigned char ch, buf[MAXBLOCKSIZE];
+
+   LTC_ARGCHK(sig  != NULL);
+   LTC_ARGCHK(hash != NULL);
+   LTC_ARGCHK(key  != NULL);
+
+   /* BEWARE: requires sqrtmod_prime */
+   if (ltc_mp.sqrtmod_prime == NULL) {
+      return CRYPT_ERROR;
+   }
+
+   /* allocate ints */
+   if ((err = mp_init_multi(&r, &s, &v, &w, &t1, &t2, &u1, &u2, &v1, &v2, &e, &x, &y, &a_plus3, NULL)) != CRYPT_OK) {
+      return err;
+   }
+
+   p = key->dp.order;
+   m = key->dp.prime;
+   a = key->dp.A;
+   b = key->dp.B;
+   if ((err = mp_add_d(a, 3, a_plus3)) != CRYPT_OK) {
+      goto error;
+   }
+
+   /* allocate points */
+   mG = ltc_ecc_new_point();
+   mQ = ltc_ecc_new_point();
+   mR = ltc_ecc_new_point();
+   if (mR == NULL || mQ  == NULL || mG == NULL) {
+      err = CRYPT_MEM;
+      goto error;
+   }
+
+   /* Only ASN.1 format signatures supported for now */
+   if ((err = der_decode_sequence_multi_ex(sig, siglen, LTC_DER_SEQ_SEQUENCE | LTC_DER_SEQ_STRICT,
+                                     LTC_ASN1_INTEGER, 1UL, r,
+                                     LTC_ASN1_INTEGER, 1UL, s,
+                                     LTC_ASN1_EOL, 0UL, NULL)) != CRYPT_OK)                             { goto error; }
+
+   /* check for zero */
+   if (mp_cmp_d(r, 0) != LTC_MP_GT || mp_cmp_d(s, 0) != LTC_MP_GT ||
+       mp_cmp(r, p) != LTC_MP_LT || mp_cmp(s, p) != LTC_MP_LT) {
+      err = CRYPT_INVALID_PACKET;
+      goto error;
+   }
+
+   /* read hash - truncate if needed */
+   pbits = mp_count_bits(p);
+   pbytes = (pbits+7) >> 3;
+   if (pbits > hashlen*8) {
+      if ((err = mp_read_unsigned_bin(e, (unsigned char *)hash, hashlen)) != CRYPT_OK)                  { goto error; }
+   }
+   else if (pbits % 8 == 0) {
+      if ((err = mp_read_unsigned_bin(e, (unsigned char *)hash, pbytes)) != CRYPT_OK)                   { goto error; }
+   }
+   else {
+      shift_right = 8 - pbits % 8;
+      for (i=0, ch=0; i<pbytes; i++) {
+        buf[i] = ch;
+        ch = (hash[i] << (8-shift_right));
+        buf[i] = buf[i] ^ (hash[i] >> shift_right);
+      }
+      if ((err = mp_read_unsigned_bin(e, (unsigned char *)buf, pbytes)) != CRYPT_OK)                    { goto error; }
+   }
+
+   /* decompress point from r=(x mod p) - BEWARE: requires sqrtmod_prime */
+   /* x = r + p*(recid/2) */
+   if ((err = mp_set(x, recid/2)) != CRYPT_OK)                                                          { goto error; }
+   if ((err = mp_mulmod(p, x, m, x)) != CRYPT_OK)                                                       { goto error; }
+   if ((err = mp_add(x, r, x)) != CRYPT_OK)                                                             { goto error; }
+   /* compute x^3 */
+   if ((err = mp_sqr(x, t1)) != CRYPT_OK)                                                               { goto error; }
+   if ((err = mp_mulmod(t1, x, m, t1)) != CRYPT_OK)                                                     { goto error; }
+   /* compute x^3 + a*x */
+   if ((err = mp_mulmod(a, x, m, t2)) != CRYPT_OK)                                                      { goto error; }
+   if ((err = mp_add(t1, t2, t1)) != CRYPT_OK)                                                          { goto error; }
+   /* compute x^3 + a*x + b */
+   if ((err = mp_add(t1, b, t1)) != CRYPT_OK)                                                           { goto error; }
+   /* compute sqrt(x^3 + a*x + b) */
+   if ((err = mp_sqrtmod_prime(t1, m, t2)) != CRYPT_OK)                                                 { goto error; }
+
+   /* fill in mR */
+   if ((err = mp_copy(x, mR->x)) != CRYPT_OK)                                                           { goto error; }
+   if ((mp_isodd(t2) && (recid%2)) || (!mp_isodd(t2) && !(recid%2))) {
+      if ((err = mp_mod(t2, m, mR->y)) != CRYPT_OK)                                                     { goto error; }
+   }
+   else {
+      if ((err = mp_submod(m, t2, m, mR->y)) != CRYPT_OK)                                               { goto error; }
+   }
+   if ((err = mp_set(mR->z, 1)) != CRYPT_OK)                                                            { goto error; }
+
+   /*  w  = r^-1 mod n */
+   if ((err = mp_invmod(r, p, w)) != CRYPT_OK)                                                          { goto error; }
+   /* v1 = sw */
+   if ((err = mp_mulmod(s, w, p, v1)) != CRYPT_OK)                                                      { goto error; }
+   /* v2 = -ew */
+   if ((err = mp_mulmod(e, w, p, v2)) != CRYPT_OK)                                                      { goto error; }
+   if ((err = mp_submod(p, v2, p, v2)) != CRYPT_OK)                                                     { goto error; }
+
+   /*  w  = s^-1 mod n */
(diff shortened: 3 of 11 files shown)


rmw42:
Hmm... some of the builds are failing :(

I'm using:
`make CFLAGS="-DUSE_LTM -DLTM_DESC -I../libtommath" EXTRALIBS="../libtommath/libtommath.a"`
locally on MacOS, with the latest LTM from GitHub... do some of the other maths providers not support `ltc_mp.ecc_mul2add`?

sjaeckel:
> do some of the other maths providers not support ltc_mp.ecc_mul2add?

good guess, but it's the missing `sqrtmod_prime()` implementation of tfm ... :-\

rmw42:
Aha! Cheers... is there a standard workaround for that, then? The logic is the same as for compressed points, which seem to require sqrtmod_prime.

rmw42:
Oh...
```
#ifndef USE_TFM
      /* XXX-FIXME: TFM does not support sqrtmod_prime */
      /* export compressed public key */
```

tests/ecc_test.c
```diff
@@ -520,6 +520,8 @@ static int _ecc_new_api(void)
       DO(ecc_verify_hash(buf, len, data16, 16, &stat, &pubkey));
       if (stat != 1) return CRYPT_FAIL_TESTVECTOR;
 
+#ifndef USE_TFM
```
sjaeckel: better to check at runtime `if (strcmp(ltc_mp.name, "TomsFastMath") != 0)` as it's possible to build for all MPI providers, with `USE_TFM`, but then select another one when invoking [`./test`](https://github.com/libtom/libtomcrypt/blob/develop/tests/test.c#L341)

rmw42: OK. Should I change the existing test for reading compressed points as well? That'll have the same issue with switching MPI provider.

rmw42:
Well, only two failures now... can't tell what the problem is from Travis CI... stdout looks to show successful completion of the tests, but then a SIGSEGV is reported after it's exited :(

rmw42:
Ahh... `ltc_mp.ecc_mul2add` is missing if built with:
`EASY (-DLTC_EASY -O2, makefile V=1 IGNORE_SPEED=1)`

karel-m:
It looks like:

1/ the whole `ecc_recovery_key` should be inside `#ifdef LTC_ECC_SHAMIR`

2/ or we might consider adding `#define LTC_ECC_SHAMIR` into `LTC_EASY`

rmw42:
Sure, happy to... just let me know what you want me to do...

sjaeckel:
Chapter 17.7.10 of the docs says: 1/ (IIUC)
I say: I don't care - but for sure I'm too lazy to re-write that Chapter and re-think all this _and_ I don't know who's even using `LTC_EASY` so I'd also go for 1/ :)

rmw42:
Sorry... the MacOS build of clang doesn't seem to do `-Wunused-variable` by default, so I missed those build errors :(

sjaeckel:
> Sorry... the MacOS build of clang doesn't seem to do `-Wunused-variable` by default, so I missed those build errors :(

No worries, that's why we have travis :)

src/pk/ecc/ecc_recover_key.c
```diff
@@ -0,0 +1,234 @@
...
+   LTC_ARGCHK(hash != NULL);
+   LTC_ARGCHK(key  != NULL);
+
+
+   /* allocate ints */
+   if ((err = mp_init_multi(&r, &s, &v, &w, &t1, &t2, &u1, &u2, &v1, &v2, &e, &x, &y, &a_plus3, NULL)) != CRYPT_OK) {
+      return err;
+   }
+
+   /* BEWARE: requires sqrtmod_prime */
```
sjaeckel, line 49: I'd prefer those checks above the allocation so you can directly return w/o the need to clean-up

src/pk/ecc/ecc_recover_key.c
```diff
@@ -0,0 +1,234 @@
...
+
+#include "tomcrypt_private.h"
+
+#ifdef LTC_MECC
+
+#ifdef LTC_ECC_SHAMIR
+
+/**
+  @file ecc_recover_key.c
+  ECC Crypto, Tom St Denis
```
sjaeckel: I guess that author isn't correct :)

sjaeckel (commented):
how about squashing all those fixup/minor-changes commits since ad8b15b10d55c3ce1e340188bf19a4ea7b69da8b together?

rmw42:
Fixed those two code review issues, squashed that and all post-makefile (ad8b15b) commits into a single commit.

src/pk/ecc/ecc_recover_key.c
```diff
@@ -0,0 +1,231 @@
...
+
+      err = CRYPT_OK;
+   }
+   else {
+      /* not found - recid is wrong or we're unable to calculate public key for some other reason */
+      err = CRYPT_INVALID_ARG;
+   }
+
+error:
+   if (mG != NULL) ltc_ecc_del_point(mG);
```
sjaeckel: can you please reverse the order of all free/clear/del-operations in relation to the allocation

src/pk/ecc/ecc_recover_key.c
```diff
@@ -0,0 +1,231 @@
...
+   if ((err = mp_set(mR->z, 1)) != CRYPT_OK)                                                            { goto error; }
+
+
+   /*  w  = r^-1 mod n */
+   if ((err = mp_invmod(r, p, w)) != CRYPT_OK)                                                          { goto error; }
+   /* v1 = sw */
+   if ((err = mp_mulmod(s, w, p, v1)) != CRYPT_OK)                                                      { goto error; }
+   /* v2 = -ew */
+   if ((err = mp_mulmod(e, w, p, v2)) != CRYPT_OK)                                                      { goto error; }
+   if ((err = mp_submod(p, v2, p, v2)) != CRYPT_OK)                                                       { goto error; }
```
sjaeckel: we need a (good) clang-format config... (or astyle or whatever...)

src/pk/ecc/ecc_recover_key.c
```diff
@@ -0,0 +1,231 @@
...
+  @file ecc_recover_key.c
+  ECC Crypto, Russ Williams
+*/
+
+static int _ecc_recover_key(const unsigned char *sig,  unsigned long siglen,
+                            const unsigned char *hash, unsigned long hashlen,
+                            int recid, ecc_key *key, int sigformat)
+{
+   ecc_point     *mG = NULL, *mQ = NULL, *mR = NULL;
+   void          *r, *s, *v, *w, *t1, *t2, *u1, *u2, *v1, *v2, *e, *x, *y, *p, *m, *a, *b, *a_plus3 = NULL, *mu = NULL, *ma = NULL;
```
sjaeckel: can you please put the declaration of `p, m, a, b` in a (one) separate line, so it's visible that they're a "different set of variables"
the same for the `NULL`-initialized ones please

src/pk/ecc/ecc_recover_key.c
```diff
@@ -0,0 +1,231 @@
...
+#ifdef LTC_ECC_SHAMIR
+
+/**
+  @file ecc_recover_key.c
+  ECC Crypto, Russ Williams
+*/
+
+static int _ecc_recover_key(const unsigned char *sig,  unsigned long siglen,
+                            const unsigned char *hash, unsigned long hashlen,
+                            int recid, ecc_key *key, int sigformat)
```
sjaeckel: there's a `sigformat` parameter which isn't used.
I see that this part is c&p'ed from `_ecc_sign/verify_hash()` but I generally don't like the idea to keep code "just in case" 
do you plant to add an API endpoint & tests for it? 
resp. would that endpoint even make sense or never be used since there exist no signatures in the wild for that use-case?

sjaeckel (commented):
:+1: LGTM
Sorry, for those additional (& nitpick) comments :) 
I didn't really check the part of the calculations for correctness!
@karel-m can you please also have a look

src/pk/ecc/ecc_recover_key.c
rmw42: I'm not sure what this means...?

Had missed the alignment of the goto, so will fix that...

src/pk/ecc/ecc_recover_key.c
rmw42: OK. It was cut/pasted from ecc_verify_hash but has grown rather unwieldy now...

src/pk/ecc/ecc_recover_key.c
rmw42: The signatures in the wild would be from Bitcoin/Ethereum, which is close to the RFC7518 format. If that were to be supported, though, it'd also make sense for `ecc_sign_hash()` to be able to create such signatures, which is probably a rather bigger change than this pull request.

If there's interest in supporting a third signature format in LTC, I'm happy to implement that. For now, it seems best to remove the unused `sigformat` parameter.

src/pk/ecc/ecc_recover_key.c
sjaeckel: yeah, it was only a personal reminder that we shouldn't have to take care manually of stuff like that :) thanks for fixing the alignment by hand

rmw42: No worries :)

rmw42:
OK, have made those fixes... I'll wait for @karel-m's review as well, and can then squash the commit(s) if desired

karel-m:
> The signatures in the wild would be from Bitcoin/Ethereum, which is close
> to the RFC7518 format. If that were to be supported, though, it'd also make
> sense for ecc_sign_hash() to be able to create such signatures, which is
> probably a rather bigger change than this pull request.

Considering the fact that the ecc related changes between libtomcrypt 1.18 > 1.19 will be quite huge it is IMO a good time for changing/extending ecc API so that it better handles different ecc signature types (if we want it it should happen before 1.19).

I may be wrong, but is not such a big change.

First, we need in `tomcrypt_pk.h` something like:

```C
enum ecc_signature_type {
   LTC_ECCSIG_ANSIX962, /* ASN1 encoded, ANSI X9.62 */
   LTC_ECCSIG_RFC7518,  /* raw R, S values */
   LTC_ECCSIG_ETH,      /* raw R, S, V values */
};
```

Then:
* rename `_ecc_sign_hash` > `ecc_sign_hash_ex` (new API function)
* rename `_ecc_verify_hash` > `ecc_verify_hash_ex` (new API function)
* after that `ecc_verify_hash`, `ecc_verify_hash_rfc7518`, `ecc_sign_hash` and `ecc_sign_hash_rfc7518` can be turned into a simple `#define ..`
* and you can return back `int sigformat` to your `_ecc_recover_key` that can be renamed to `ecc_recover_key`

I think, that it can go to this PR (for sure as a separate commit).

BTW: the name `ecc_recover_key` is not bad but it is basically a key import function; any ideas for `ecc_import_<something>` name?

src/pk/ecc/ecc_recover_key.c
```diff
@@ -0,0 +1,222 @@
...
+*/
+
+static int _ecc_recover_key(const unsigned char *sig,  unsigned long siglen,
+                            const unsigned char *hash, unsigned long hashlen,
+                            int recid, ecc_key *key)
+{
+   ecc_point     *mG = NULL, *mQ = NULL, *mR = NULL;
+   void          *p, *m, *a, *b;
+   void          *r, *s, *v, *w, *t1, *t2, *u1, *u2, *v1, *v2, *e, *x, *y;
+   void          *a_plus3 = NULL, *mu = NULL, *ma = NULL;
```
karel-m: shouldn't ` *a_plus3` be on the previous line (without `= NULL`)?

src/pk/ecc/ecc_recover_key.c
```diff
@@ -0,0 +1,222 @@
...
+   LTC_ARGCHK(sig  != NULL);
+   LTC_ARGCHK(hash != NULL);
+   LTC_ARGCHK(key  != NULL);
+
+   /* BEWARE: requires sqrtmod_prime */
+   if (ltc_mp.sqrtmod_prime == NULL) {
+      return CRYPT_ERROR;
+   }
+   /* BEWARE: requires ecc_mul2add for Shamir's trick */
+   if (ltc_mp.ecc_mul2add == NULL) {
```
karel-m: probably not needed anymore as we are inside `#ifdef LTC_ECC_SHAMIR`

src/pk/ecc/ecc_recover_key.c
```diff
@@ -0,0 +1,222 @@
...
+
+   /* allocate points */
+   mG = ltc_ecc_new_point();
+   mQ = ltc_ecc_new_point();
+   mR = ltc_ecc_new_point();
+   if (mR == NULL || mQ  == NULL || mG == NULL) {
+      err = CRYPT_MEM;
+      goto error;
+   }
+
```
karel-m, line 75: IMO no need for more than 1 empty line (also on several places below)

src/pk/ecc/ecc_recover_key.c
```diff
@@ -0,0 +1,222 @@
...
+   else {
+      /* not found - recid is wrong or we're unable to calculate public key for some other reason */
+      err = CRYPT_INVALID_ARG;
+   }
+
+error:
+   if (ma != NULL) mp_clear(ma);
+   if (mu != NULL) mp_clear(mu);
+   if (mp != NULL) {
+      mp_montgomery_free(mp);
```
karel-m: just from aesthetic reasons keep this on a single line ` if (mp != NULL) mp_montgomery_free(mp);`

karel-m:
I did a quick skim - basically looks good, couple of comment added to the code. I'll look at it more carefully over the weekend.

rmw42:
> BTW: the name `ecc_recover_key` is not bad but it is basically a key import function; any ideas for `ecc_import_<something>` name?

I can see your point, but "import" implies that there's a symmetrical "export"... this is really a side-effect of how ECDSA works, cancelling out the witness instead of the private key in order to find the public key.

I'm not sure "recover" is the best term, but it's the one that's used elsewhere (e.g. the `ecrecover` function in Solidity), and it conveys that it's bringing back something that's been lost...

src/pk/ecc/ecc_recover_key.c
rmw42: OK. I nearly did that yesterday when I tidied up the frees... it was a cut/paste from `ecc_verify_hash()`, but it looks like it's been left over from some previous debugging.

src/pk/ecc/ecc_recover_key.c
rmw42: OK

src/pk/ecc/ecc_recover_key.c
rmw42: It had the `= NULL` from `ecc_verify_hash()`, but I can't see a good reason for that. It's initialised with the rest in that giant `mp_init_multi()`, isn't optional (like ma/mu), and isn't a curve constant (like p/m/a/b).

Moved it and removed the NULL initialisation.

src/pk/ecc/ecc_recover_key.c
rmw42: OK

rmw42:
Code review changes are in. I'll get on with the broader change for adding a new signature format today, since it seems like that's worth having.

I'll squash the code review changes when I commit that, to keep it tidy.

rmw42:
Lots of squashing... commit 6cb796a is everything up to this morning, the makefiles (commit aceb17b) are moved to just after this, and then commit 27b5035 is the new stuff for extended signature format handling.

`ecc_sign_hash_ex()` now takes the signature format and an (optional) pointer-to-int to store the recovery ID, for signature formats which discard it. This is passed as NULL by the macros.

I've tidied up `ecc_sign_hash.c` and `ecc_verify_hash.c` where I've touched them, to address the code review issues previously identified with `ecc_recover_key.c`. Hopefully all three are consistent now.

I've included a bunch of tests around key recovery. I took a signature example from the ethereum/go-ethereum project to verify compatibility, but then replaced it with a newly-generated equivalent test to avoid even the slightest sniff of LGPL code. Also added a bunch of random generate/sign/verify/recover tests for all the different curves.

I've also updated the docs to add the _ex versions of the methods.

Let me know what you think! :)

rmw42:
PS - Don't ask me what the private key for the new test was. It was randomly generated, and I forgot to print it out... Guess it proves that the public key is recovered without cheating ;-)

rmw42:
Oops... `ecc_free(&privkey)` in the test but `privkey` wasn't used that early. Not caught in some build settings :(

karel-m:
Great work.

I am considering adding also something like:
```
   /* SSH + ECDSA signature format defined by RFC5656 */
   LTC_ECCSIG_RFC5656 = 0x3
```
I am pretty sure that dropbear guys (libtomcrypt prominent consumers) will appreciate it as nowadays they have to implement a quite ugly kung-fu around ECDSA sign/verify.

The SSH signature encoding is defined here https://tools.ietf.org/html/rfc5656#section-3.1.2 Technically it looks like:
```
ecdsa-sha2-<curvename><mpint_r><mpint_s>
```
There are basically 2 gotchas:

1/ we need to map `ltc_ecc_dp->oid` to `<curvename>`

2/ the bigint encoding (used for `<mpint_r><mpint_s>` part) is slightly special - see https://tools.ietf.org/html/rfc4251#section-5 (scroll to "mpint")

But if you start to feel that I am pushing you too far, just let me know. We can leave the SSH/ECDSA signature format for a separate PR.

doc/crypt.tex
```diff
@@ -5537,12 +5553,48 @@ \subsection{Signature Verification}
...
+                          unsigned long  hashlen,
+                                    int  recid,
+                                    int  sigformat,
+                                ecc_key *key);
+\end{verbatim}
+
+This function will recover (a) public key from the ECDSA signature in the array pointed to by \textit{sig} of length \textit{siglen} octets, the message digest
+pointed to by the array \textit{hash} of length \textit{hashlen}, and the recovery id \textit{recid}. It will store the recovered
+key into \textit{key} and return CRYPT_OK if recovery succeeds, or an error if recovery fails.
+This is for compatibility with the (v,r,s) signatures used in Bitcoin/Ethereum, where public keys are not explicitly
```
karel-m: could you please add a link to some Bitcoin/Ethereum specification or BIP or another document defining this signature format?

doc/crypt.tex
```diff
@@ -5507,6 +5507,22 @@ \subsection{Signature Generation}
...
+                              prng_state *prng,
+                                     int  wprng,
+                                     int  sigformat,
+                                     int *recid,
+                                 ecc_key *key);
+\end{verbatim}
+
+This function is an extended version of the ECDSA signature in \textit{ecc\_sign\_hash}, but with a choice of output formats
+and an optional output of the recovery ID for use with \textit{ecc\_recover\_key}.
+
```
karel-m, line 5567: a table with supported `sigformat` values would be handy here:

| sigformat | description |
|-------|-------------|
|  LTC_ECCSIG_ANSIX962 | ASN.1 encoded, ANSI X9.62 |
|  LTC_ECCSIG_RFC7518  | raw R, S values as defined in RFC7518 |
| LTC_ECCSIG_BLOCKCHAIN | raw V, R, S values |

tests/ecc_test.c
```diff
@@ -501,6 +520,27 @@ static int _ecc_new_api(void)
       DO(ecc_verify_hash(buf, len, data16, 16, &stat, &pubkey));
       if (stat != 1) return CRYPT_FAIL_TESTVECTOR;
 
+#ifdef LTC_ECC_SHAMIR
+      /* XXX-FIXME: ecc_recover_key currently requires mul2add */
```
karel-m: this comment can be probably left out

doc/crypt.tex
rmw42: That's remarkably difficult to find - the few docs there are differ on whether v is the first byte or the last(!), and which values are acceptable.

The Ethereum "yellow paper" is probably the best to cite, so I'll add that link. Bitcoin was originally using X9.62/DER... that means LTC_ECCSIG_ETH might actually be the best name after all.

tests/ecc_test.c
rmw42: OK

doc/crypt.tex
rmw42: I've not done any LaTeX before this, so I'm not familiar with how it all works... something like this?

```
\begin{figure}[hpbt]
\index{Signature Formats}
\begin{small}
\begin{center}
\begin{tabular}{|l|l|}
\hline \textbf{sigformat} & \textbf{description} \\
\hline LTC_ECCSIG_ANSIX962 & ASN.1 encoded, ANSI X9.62 \\
\hline LTC_ECCSIG_RFC7518 & raw R, S values as defined in RFC7518 \\
\hline LTC_ECCSIG_ETH & raw V, R, S values as defined in Yellow Paper\\
\hline
\end{tabular}
\end{center}
\end{small}
\caption{Signature Formats}
\label{fig:sigformat}
\end{figure}
```

karel-m: you have to escape `_` as `\_`

```
\begin{figure}[hpbt]
\index{Signature Formats}
\begin{small}
\begin{center}
\begin{tabular}{|l|l|}
\hline \textbf{sigformat} & \textbf{description} \\
\hline LTC\_ECCSIG\_ANSIX962 & ASN.1 encoded, ANSI X9.62 \\
\hline LTC\_ECCSIG\_RFC7518 & raw R, S values as defined in RFC7518 \\
\hline LTC\_ECCSIG\_ETH & raw V, R, S values as defined in Yellow Paper\\
\hline
\end{tabular}
\end{center}
\end{small}
\caption{Signature Formats}
\label{fig:sigformat}
\end{figure}
```

rmw42:
> But if you start to feel that I am pushing you too far, just let me know. We can leave the SSH/ECDSA signature format for a separate PR.

I think that would definitely be better as a separate PR.

I'm happy to do the work, and the `mpint` format rules look very similar to ASN.1 DER (e.g. the mandatory zero prefix if the top bit is set in an unsigned value), but I don't know how long the curve name lookups will take to implement, and I've no idea where to find examples of this signature.

doc/crypt.tex
rmw42: Cheers!
I'd got a mistake before with an unescaped _ in `CRYPT_OK`, so I've fixed that, too. Managed to get a build of the doc working - LaTeX isn't fun on a Mac - and it all looks fine now.

src/pk/ecc/ecc_recover_key.c
```diff
@@ -0,0 +1,238 @@
...
+      if ((err = mp_read_unsigned_bin(s, (unsigned char *)sig+i, i)) != CRYPT_OK)                       { goto error; }
+   }
+   else if (sigformat == LTC_ECCSIG_ETH) {
+      /* Bitcoin/Ethereum (v,r,s) format */
+      if (siglen != 65) { /* Only secp256k1 curves used */
+         err = CRYPT_INVALID_PACKET;
+         goto error;
+      }
+      i = (unsigned long)sig[64];
+      if ((i>=27) && (i<31)) i -= 27; /* Bitcoin/Ethereum add 27 to recovery ID to make it look less arbitrary */
```
karel-m: According to https://github.com/ethereum/EIPs/blob/master/EIPS/eip-155.md it looks like the value `27` is not that fixed as it looks; IIUC ~~it can be `27`, `28`, `37`, `38`~~ it can be `27` or `37` (and maybe more in the future). But I have not got deep enough into ETH yet.

rmw42: The original format is 27/28, and that's still valid even on the Ethereum main chain.

As you say, EIP-155 adds 35/36 _**plus 2x the chain ID**_ - which we cannot possibly know. That's why the library implementations tend to go with 0/1 and leave it up to the application to sort out. It's also not possible to represent 1337x2+35 (private chains) in a byte, so this format cannot be used for that.

29/30 are also possible on the original scheme, though the Yellow Paper incorrectly describes them as "infinite". However, most ETH implementations can't handle them, and since they only have about a 1-in-2^127 chance of occurring, nobody will ever see them in the life of the universe.

Basically, this entire space is an undocumented ill-conceived mess...

rmw42: This implementation will handle the output of https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_sign when the hex value is converted to big-endian binary (convention: https://github.com/ethereum/wiki/wiki/JavaScript-API#web3toascii) - the last two hex digits of the signature in the `eth_sign` example output are `1b`, which means the last byte will be `0x1b` = 27, giving a recovery ID of 0.

I wish there were a clear, normative reference for all this, but I've been looking at this space for months and still haven't seen one :(

karel-m: OK, I suggest being more explicit and rename `LTC_ECCSIG_ETH` to `LTC_ECCSIG_ETH27` so that it is clear that it means *"Ethereum +27 El Clásico"* variant. The caller of `ecc_sign_hash_ex` gets the true recovery id in `*recid` anyway so they can handle it afterwards as they wish.

rmw42: OK, makes sense. On that basis, the main change will be in `ecc_sign_hash_ex` - we should probably support both 0/1 and 27/28 equivalently in `ecc_recover_key`, which is what that line does.

rmw42: Done. I've updated the docs and tests, removed any mention of Bitcoin (since I can't see any clear reference to non-DER signatures there), and also fixed a zero-padding bug when r or s are smaller than 2^248.

rmw42:
Is there anything else I need to do for this PR?

rmw42:
Rebased to current root/develop branch, squashed signature format/code review commits.

src/pk/ecc/ecc_sign_hash.c
```diff
@@ -69,6 +84,16 @@ static int _ecc_sign_hash(const unsigned char *in,  unsigned long inlen,
       /* find r = x1 mod n */
       if ((err = mp_mod(pubkey.pubkey.x, p, r)) != CRYPT_OK)               { goto error; }
 
+      /* find recovery ID */
```
sjaeckel: as this `recid` is only used when `sigformat == LTC_ECCSIG_ETH27` shouldn't this go in the appropriate location within that `else if`?

src/pk/ecc/ecc_sign_hash.c
```diff
@@ -92,8 +117,17 @@ static int _ecc_sign_hash(const unsigned char *in,  unsigned long inlen,
       goto errnokey;
    }
 
-   if (sigformat == 1) {
-      /* RFC7518 format */
+   if (recid) *recid = v;
```
sjaeckel, line 122: same

src/pk/ecc/ecc_verify_hash.c
```diff
@@ -65,12 +84,19 @@ static int _ecc_verify_hash(const unsigned char *sig,  unsigned long siglen,
       if ((err = mp_read_unsigned_bin(r, (unsigned char *)sig,   i)) != CRYPT_OK)                       { goto error; }
       if ((err = mp_read_unsigned_bin(s, (unsigned char *)sig+i, i)) != CRYPT_OK)                       { goto error; }
    }
+   else if (sigformat == LTC_ECCSIG_ETH27) {
+      /* Ethereum (v,r,s) format */
+      if (siglen != 65) { /* Only secp256k1 curves */
```
sjaeckel: is the `siglen` the only indication of a secp256k1? looks a bit lax to me

src/pk/ecc/ecc_recover_key.c
```diff
@@ -0,0 +1,238 @@
...
+   }
+
+error:
+   if (ma != NULL) mp_clear(ma);
+   if (mu != NULL) mp_clear(mu);
+   if (mp != NULL) mp_montgomery_free(mp);
+   if (mR != NULL) ltc_ecc_del_point(mR);
+   if (mQ != NULL) ltc_ecc_del_point(mQ);
+   if (mG != NULL) ltc_ecc_del_point(mG);
+   mp_clear_multi(r, s, v, w, t1, t2, u1, u2, v1, v2, e, x, y, a_plus3, NULL);
```
sjaeckel: can you please also turn those args around?

_sidenote_ we could probably make `mp_clear_multi()` process the args from the end 'til the beginning... but I fear this implementation would be worse to implement than always reverse the order on invocation :D

src/headers/tomcrypt_pk.h
```diff
@@ -286,21 +296,30 @@ int  ecc_decrypt_key(const unsigned char *in,  unsigned long  inlen,
...
+
+#define ecc_verify_hash(sig_, siglen_, hash_, hashlen_, stat_, key_) \
+   ecc_verify_hash_ex(sig_, siglen_, hash_, hashlen_, LTC_ECCSIG_ANSIX962, stat_, key_)
 
-int  ecc_sign_hash(const unsigned char *in,  unsigned long inlen,
-                         unsigned char *out, unsigned long *outlen,
-                         prng_state *prng, int wprng, const ecc_key *key);
+int  ecc_sign_hash_ex(const unsigned char *in,  unsigned long inlen,
+                            unsigned char *out, unsigned long *outlen,
+                            prng_state *prng, int wprng, int sigformat,
```
sjaeckel: how about using the `enum` as type of the `sigformat` argument?
I'd also be fine to make that `enum` a typedef (c.f. the `ltc_asn1_...` enums)

src/pk/ecc/ecc_sign_hash.c
rmw42: It's an independent property of the signature - the `recid` output isn't needed for LTC_ECCSIG_ETH27, since that contains the value, but can be used to recover keys with a signature in a different format if separately transmitted.

I could put the calculation in an `if (recid || sigformat==LTC_ECCSIG_ETH27)` block so it's only calculated if it's going to be used?

src/pk/ecc/ecc_verify_hash.c
rmw42: There's no indication of the curve _at all_, as with the RFC7518, but since this signature is used only by applications using the secp256k1 curve, it must be 65 bytes long. If it's not, it's some other format.

src/pk/ecc/ecc_recover_key.c
rmw42: Well, going from last to first would probably not be too bad for a recursive solution, but no problem in swapping the order around. I'll do that now.

Sorry, hadn't realised that `mp_init_multi`/`mp_clear_multi` wasn't atomic.

src/headers/tomcrypt_pk.h
rmw42: I took the `padding` example in `rsa_encrypt_key_ex`, but I did originally wonder about this. Happy to change it to a typedef and let the compiler do a little bit of type checking :)

src/pk/ecc/ecc_sign_hash.c
rmw42: Done that... the calculation will now happen only if something wants the recovery ID, either as an output parameter or for the `LTC_ECCSIG_ETH27` format.

src/pk/ecc/ecc_verify_hash.c
rmw42: I've changed the comment here to clarify that logic

karel-m: When talking about RFC7518 I had a similar discussion about what we now call `LTC_ECCSIG_RFC7518` - see https://github.com/DCIT/perl-CryptX/issues/45

The thing is that by `LTC_ECCSIG_RFC7518` we mean **only the signature format** not other conditions defined by the RFC in question.

Similarly by `LTC_ECCSIG_ETH27` we should mean **the format** how the final byte sequence is constructed.

So for `LTC_ECCSIG_ETH27` we should IMO only test that `siglen` is equal to `key->dp.size * 2 + 1` and that `sig[key->dp.size * 2] >= 27`.

rmw42: The signature format in RFC7518 is at least _defined_ for different key sizes, whereas the Ethereum stuff isn't as there is only a single curve in use. There are literally no signatures in this format which aren't either 65 bytes (or, equivalently, 130 hex digits).

It's a simple enough change if you want me to make it, though it probably should also get rid of the curve OID check in `ecc_sign_hash` so there's at least some way to generate the non-standard forms. Just let me know...

karel-m: Does `ecc_recover_key` work also with other curves - especially those with `cofactor != 1`?

rmw42: Yes. The main one to test with is `secp112r2` (cofactor 4) - I've seen recovery values for that during the tests coming through as (e.g.) 6. That's with an X9.62 signature and a separate recovery value, though.

... which is a good point. Adding 27 would get rather weird for values of `v` other than 0/1.

rmw42:
Is there anything else you need me to do here?

I've got the suggested SSH+ECDSA stuff about half done in another branch, but it needs to be rebased for a pull request.

doc/crypt.tex
```diff
@@ -5549,6 +5549,22 @@ \subsection{Signature Generation}
...
 
+\index{ecc\_sign\_hash_ex()}
+\begin{verbatim}
+int ecc_sign_hash_ex(const unsigned char *in,
+                           unsigned long  inlen,
+                           unsigned char *out,
+                           unsigned long *outlen,
+                              prng_state *prng,
+                                     int  wprng,
+                                     int  sigformat,
```
sjaeckel: this should be `ecc_signature_type sigformat,` (applies to the other parts of the docs as well

sjaeckel (dismissed):
@karel-m LGTM, also fine for you?

doc/crypt.tex
rmw42: Ah! Oops, I'll change that now.

doc/crypt.tex
```diff
@@ -5579,12 +5595,71 @@ \subsection{Signature Verification}
...
+                     ecc_signature_type  sigformat,
+                                ecc_key *key);
+\end{verbatim}
+
+This function will recover (a) public key from the ECDSA signature in the array pointed to by \textit{sig} of length \textit{siglen} octets, the message digest
+pointed to by the array \textit{hash} of length \textit{hashlen}, and the recovery id \textit{recid}. It will store the recovered
+key into \textit{key} and return CRYPT\_OK if recovery succeeds, or an error if recovery fails.
+This is for compatibility with the (v,r,s) signatures used in Ethereum, where public keys are not explicitly shared,
+only the parity of the public key. For curves like secp256k1, recid will take values of 0 or 1, corresponding to the
+parity of the public key's y coordinate. For curves like secp112r2, with a cofactor of 4, values 0..7 are possible,
```
sjaeckel, line 5634: The documentation says here that key recovery is usable with secp112r2 and secp256k1, but the implementation only supports secp256k1. Shouldn't one of both be adapted? Either it is made clear in the doc that 'in theory it should be possible but is not implemented' or the implementation gets extended? Or did I miss something and/or get something wrong there?

src/pk/ecc/ecc_recover_key.c
```diff
@@ -0,0 +1,238 @@
...
+        buf[i] = ch;
+        ch = (hash[i] << (8-shift_right));
+        buf[i] = buf[i] ^ (hash[i] >> shift_right);
+      }
+      if ((err = mp_read_unsigned_bin(e, (unsigned char *)buf, pbytes)) != CRYPT_OK)                    { goto error; }
+   }
+
+   /* decompress point from r=(x mod p) - BEWARE: requires sqrtmod_prime */
+   /* x = r + p*(recid/2) */
+   if ((err = mp_set(x, recid/2)) != CRYPT_OK)                                                          { goto error; }
```
sjaeckel, line 155: `recid` is used here w/o check of any restrictions for the cases where `sigformat != LTC_ECCSIG_ETH27` is that okay?

doc/crypt.tex
rmw42: The implementation will recover keys for any of the ECC curves - this is exercised in the tests (around line 535 of `ecc_test.c`, in a loop through curve types):
```
         for (j = 0; j < 2*(1+(int)privkey.dp.cofactor); j++) {
            stat = ecc_recover_key(buf, len, data16, 16, j, LTC_ECCSIG_ANSIX962, &reckey);
            if (stat != CRYPT_OK) continue; /* last two will almost always fail, only possible if x<(prime mod order) */
            stat = _ecc_key_cmp(PK_PUBLIC, &pubkey, &reckey);
            if (stat == CRYPT_OK) found++;
         }
         if (found != 1) return CRYPT_FAIL_TESTVECTOR; /* unique match */
```

The `_ecc_test_recovery` function actually goes through all of the different curves and tests recovery - generating a `LTC_ECCSIG_RFC7518` signature and separate recovery ID (line 1084), then recovering the public key (line 1093) and checking it matches (line 1094).


The restriction on secp256k1 is only for generating `LTC_ECCSIG_ETH27` signatures, because that's the only curve used by Ethereum/Bitcoin/etc and because adding 27 to the recovery value is specific to Ethereum and gets rather confusing for curves with a cofactor other than 1, so the signature format is undefined for any other curve type.

rmw42: Oh... there _is_ an `#ifdef LTC_ECC_SECP256K` block in `_ecc_test_recovery`, because the `LTC_ECCSIG_ETH27` signature example is for secp256k1, so there's no way to run that part of the test if that curve isn't included in the build.

karel-m:
Ad `LTC_ECCSIG_ETH27` - I would like to see it slightly more general; not strictly tied to `secp256k1`. Simply an "ethereum-like" format that stores: R (left zero-padded to the curve size), S (left zero-padded to the curve size) and 1 byte with `recovery_id + 27`. @rmw42  could you please change it in this way?

src/pk/ecc/ecc_recover_key.c
```diff
@@ -0,0 +1,238 @@
...
+   if (sigformat == LTC_ECCSIG_ANSIX962) {
+      /* ANSI X9.62 format - ASN.1 encoded SEQUENCE{ INTEGER(r), INTEGER(s) }  */
+      if ((err = der_decode_sequence_multi_ex(sig, siglen, LTC_DER_SEQ_SEQUENCE | LTC_DER_SEQ_STRICT,
+                                     LTC_ASN1_INTEGER, 1UL, r,
+                                     LTC_ASN1_INTEGER, 1UL, s,
+                                     LTC_ASN1_EOL, 0UL, NULL)) != CRYPT_OK)                             { goto error; }
+   }
+   else if (sigformat == LTC_ECCSIG_RFC7518) {
+      /* RFC7518 format - raw (r,s) */
+      if ((siglen % 2) == 1) {
```
karel-m: the test here should be IMO slightly stricter: `if (siglen != (2 * key->dp.size)) {

src/pk/ecc/ecc_recover_key.c
rmw42: Yes. `recid` is always required for the function to operate, and must always be filled in by that point. If `sigformat` is `LTC_ECCSIG_ETH27` then the value can be obtained from the signature (line 99 of `ecc_recover_key.c`) instead and overwrites the passed-in `recid` (line 106).

karel-m: we should at least test whether `recid > 0`

src/pk/ecc/ecc_sign_hash.c
```diff
@@ -103,13 +139,29 @@ static int _ecc_sign_hash(const unsigned char *in,  unsigned long inlen,
...
+         /* Only valid for secp256k1 - OID 1.3.132.0.10 */
+         err = CRYPT_ERROR; goto errnokey;
+      }
+      if (*outlen < 65) { err = CRYPT_MEM; goto errnokey; }
+      zeromem(out, 65);
+      i = mp_unsigned_bin_size(r);
+      if ((err = mp_to_unsigned_bin(r, out + 32 - i)) != CRYPT_OK) { goto errnokey; }
+      i = mp_unsigned_bin_size(s);
+      if ((err = mp_to_unsigned_bin(s, out + 64 - i)) != CRYPT_OK) { goto errnokey; }
+      out[64] = (unsigned char)(v + 27); /* Recovery ID is 27/28 for Ethereum */
```
karel-m, line 155: before this line we should perhaps check whether `(v + 27) < 256`

src/pk/ecc/ecc_recover_key.c
sjaeckel: Sorry, my question wasn't clear enough :) 
What you described is fine! But is it okay that `recid` is used without any bounds-check? as in 'Would it be okay to call `ecc_recover_key()` with some nonsense `recid=12345` and besides 'it won't work' there's no risk involved!?"

doc/crypt.tex
karel-m: As mentioned elsewhere I am for removing the restriction on secp256k1 and declare `LTC_ECCSIG_ETH27` as ethereum-like style of storing signatures as: zero-padded R value + zero-padded S value + 1 byte with recovery_id + 27.

rmw42:
> Ad `LTC_ECCSIG_ETH27` - I would like to see it slightly more general; not strictly tied to `secp256k1`. Simply an "ethereum-like" format that stores: R (left zero-padded to the curve size), S (left zero-padded to the curve size) and 1 byte with `recovery_id + 27`. @rmw42 could you please change it in this way?

What would use this? This is literally a new signature format that you're defining here, not one which exists anywhere in the world today.

I can see the point in a general-purpose (V,R,S) format, so that applications can get the recovery ID in a single blob of data, but IMHO that shouldn't have the "+27" hack in there.

src/pk/ecc/ecc_recover_key.c
rmw42: OK

karel-m:
> I can see the point in a general-purpose (V,R,S) format, so that applications can get the recovery ID in a single blob of data, but IMHO that shouldn't have the "+27" hack in there.

That is why we have "27" in `LTC_ECCSIG_ETH27` (which should be read as `LTC_ECCSIG_WITH_RECOVERYID_AND_PLUS27_HACK_AS_USED_BY_ETHEREUM`)

I am not aware on any more widely used ecc signature format containing recovery_id therefore I would stick to the ethereum-style format but keep it relaxed as for the underlying curve (providing that we are sure that recovery id always fits to `0..228`)

src/pk/ecc/ecc_sign_hash.c
sjaeckel: I guess that's up to the specification that defines that format and I can also very well imagine that `v` could be okay to overflow in this case as it's "the parity" of the key

sjaeckel:
> I am not aware on any more widely used ecc signature format containing recovery_id therefore I would stick to the ethereum-style format but keep it relaxed as for the underlying curve (providing that we are sure that recovery id always fits to `0..228`)

TBH I wouldn't start to define a new protocol if there's no real use to it and would prefer to stick with this single implementation instead of a generalized imaginative new format which e.g. already has this `+27` special case, overflows and what else comes up then...

src/pk/ecc/ecc_recover_key.c
```diff
@@ -0,0 +1,238 @@
...
+   }
+   else if (sigformat == LTC_ECCSIG_ETH27) {
+      /* Ethereum (v,r,s) format */
+      if (siglen != 65) { /* Only secp256k1 curves use this format, so must be 65 bytes long */
+         err = CRYPT_INVALID_PACKET;
+         goto error;
+      }
+      i = (unsigned long)sig[64];
+      if ((i>=27) && (i<31)) i -= 27; /* Ethereum adds 27 to recovery ID */
+      if (recid >= 0 && ((unsigned long)recid != i)) {
```
sjaeckel, line 106: I just realized that this would fail if someone has the `recid` somehow stored in the format that is used here, i.e. `recid` is `27` or `28`.
I get the impression that this hack makes a proper implementation pretty confusing ;)

src/pk/ecc/ecc_recover_key.c
```diff
@@ -0,0 +1,238 @@
...
+      if ((err = mp_read_unsigned_bin(s, (unsigned char *)sig+i, i)) != CRYPT_OK)                       { goto error; }
+   }
+   else if (sigformat == LTC_ECCSIG_ETH27) {
+      /* Ethereum (v,r,s) format */
+      if (siglen != 65) { /* Only secp256k1 curves use this format, so must be 65 bytes long */
+         err = CRYPT_INVALID_PACKET;
+         goto error;
+      }
+      i = (unsigned long)sig[64];
+      if ((i>=27) && (i<31)) i -= 27; /* Ethereum adds 27 to recovery ID */
```
sjaeckel, line 105: I'm not sure though whether this should better be an error-case when `i < 27 || i >= 31`!?

src/pk/ecc/ecc_recover_key.c
sjaeckel: Also wouldn't it make more sense to check for `i==0||i==1` as that's the only valid parity as of the documentation?
[Edit] plus the check if it equals recid

src/pk/ecc/ecc_recover_key.c
rmw42: Using a crazily large value won't produce any meaningful result, as it'll be larger than the prime modulus. This will fail to verify the signature, so will be rejected - this is tested with the largest two values of `j` in the loop.

For a very large or negative value, the calculations may overflow and fail but I'm not aware of any specific risk. We could limit the range to `0 <= recid < 2*(dp.order+1)` to prevent this?

sjaeckel: > We could limit the range to `0 <= recid < 2*(dp.order+1)` to prevent this?

that sounds good to me :) @karel-m what do you think?

src/pk/ecc/ecc_recover_key.c
rmw42: Some libraries (e.g. ethereum-go) sometimes use 0/1 instead of 27/28, relying on the application to change the value. The "new" (EIP-155) format adds 35/36+2xChainID, so the signature depends on the blockchain that the signature will be used in, and therefore can _only_ be calculated by the application.

2/3 are prohibited by the Ethereum _Yellow Paper_ spec, defined by most of the libraries I've seen (but never expected to be encountered as the probability is below 1-in-2^127), and may or may not be handled correctly by them.

Basically, that line is for Postel's Law - if we might encounter 0/1 or 27/28 as meaning the same thing, we should accept both interchangeably, but only generate the canonical 27/28 form.

sjaeckel: > Basically, that line is for Postel's Law - if we might encounter 0/1 or 27/28 as meaning the same thing, we should accept both interchangeably, but only generate the canonical 27/28 form.

OK!

src/pk/ecc/ecc_sign_hash.c
rmw42: The only way it could overflow is for a curve with a cofactor of at least 114, which will break other things as then the order is not the same (bit-)length as the prime modulus. None of the defined ECC curves have a cofactor >4, which means the largest value for `v` will be 9.

src/pk/ecc/ecc_recover_key.c
rmw42: The `>=0` check was so that `recid` could be passed in as -1 to signal "use the one in the signature", otherwise is must match. This means it's not necessary for an application to extract the recid from the signature blob in order to use `ecc_recover_key`.

src/pk/ecc/ecc_recover_key.c
karel-m: but the `recid` is `int recid` which is IMO always lower than `2*(dp.order+1)` or did I miss something?

rmw42:
For the signature format, how's this for a compromise?

I can add another entry to the enum, corresponding to an extended X9.62 (which seems to be the dominant signature format), containing SEQ{INT(r), INT(s), INT(v)} with the raw recovery value (no +27) and no restrictions on which curves can be used. Since ASN.1 includes length information, this doesn't rely on implicit size calculations based on the size of a curve, unlike the raw formats.

src/pk/ecc/ecc_recover_key.c
rmw42: Sorry, I'm half asleep. I meant `dp.cofactor+1`.

karel-m: OK, `0 <= recid < 2*(dp.cofactor+1)` looks good to me

src/pk/ecc/ecc_recover_key.c
rmw42: This causes some of the old tests to fail for `ecc_verify_hash_rfc7518`. In particular, secp160r1 (20 bytes/160 bits) comes out as a 42-byte signature. That looks to be incorrect from the way the RFC specifies the conversion.

rmw42: Looking at the memory for the signature blob, it's `00 21 1e 28 ... 4d 00 ef ea ...` so it looks like using `pbytes` in `ecc_sign_hash` is causing the signature encoding to do the wrong thing with this curve.

src/pk/ecc/ecc_recover_key.c
rmw42: OK, I'll put that check in.

src/pk/ecc/ecc_recover_key.c
rmw42: Apparently it's the calculation of those values which is the problem: `pbits = 161` and `pbytes = 21`, so this is `mp_count_bits(key->dp.order)` that's returning 161 for a 160-bit curve...

rmw42: I'm going to revert the check to `(siglen % 2)==1` as this potentially affects far too much for a trivial fix.

karel-m: The thing is that the rfc in question (RFC7518) does not talk about `secp160r1`; it mentions only "P-256", "P-384" and "P-521". And `secp160r1` is an example where `curve.order` > `curve.prime`.

we should use rather `order_size = mp_unsigned_bin_size(key->dp.order)` + `if (siglen != (2 * order_size))`

karel-m:
No, that goes too far. I would stay with `LTC_ECCSIG_ETH27` in one of the following variants:

1/ we will accept any curve

2/ we will accept only curves with `cofactor == 1`

3/ we will accept only SECP256K

In any case we should enforce the chosen limitation consistently for `LTC_ECCSIG_ETH27` in `ecc_sign_hash_ex` , `ecc_verify_hash_ex` and `ecc_recovery_key`. Currently you check it only in `ecc_sign_hash_ex`.

src/pk/ecc/ecc_recover_key.c
rmw42: Ugh :(
Yeah, that makes sense... but `r` can never be bigger than `curve.prime`, as it's the x coordinate of a point on the curve, so will never need more than `curve.size` bytes to represent it, even if `s` needs more.

rmw42: Right, that works.

rmw42:
OK, I'll add the curve OID checks to `ecc_verify_hash_ex` and `ecc_recover_key`, then. This format really doesn't exist for anything other than secp256k1. Sound good?

karel-m:
Perhaps consider utilising:
```
if ((err = ecc_get_oid_str(oid, &oidlen, key)) != CRYPT_OK) goto ...;
if (XSTRCMP(oid, "1.3.132.0.10") != 0) goto ...;
```

rmw42:
Yeah, that might be better... but since the same check is done in three places, it might be better as a separate comparison function along side things like `pk_oid_num_to_str`.

Since there's a bunch of curve<->OID stuff needed for the SSH+ECDSA changes, how about we leave it as-is for this PR and I make the changes in that branch?

karel-m:
Ok, leave it as it is

src/pk/ecc/ecc_recover_key.c
rmw42: Range of `recid` is now limited to `0 <= recid < 2*(key->dp.cofactor+1)`

src/pk/ecc/ecc_sign_hash.c
rmw42: `LTC_ECCSIG_ETH27` is now restricted to only the secp256k1 curve, so 0<=v<4, with values of 2/3 being effectively impossible (probability less than 1-in-2^127), meaning no overflow is possible.

rmw42:
OK, I think that's everything that was discussed... have I missed anything? Or is this PR now good to go?

karel-m (approved):

sjaeckel (approved):

sjaeckel:
:+1:

rmw42:
Yay! Cheers :)

PR #437, opened 2018-09-18
Fixed errors with Android by not using asm code in Android.

imays76:
### Checklist

### DIFF
--- a/src/headers/tomcrypt_macros.h
+++ b/src/headers/tomcrypt_macros.h
@@ -353,7 +353,7 @@ static inline ulong32 RORc(ulong32 word, const int i)
 
 
 /* 64-bit Rotates */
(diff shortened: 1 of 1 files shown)


rofl0r:
maybe you post more details and the errors you got ? android can also be compiled on x86, btw...

imays76:
Many of Android devices have ARM CPU. 
The problem I encountered was already several years ago and I don't remember what is the detailed problem. Maybe it is a compile error.

sjaeckel:
> maybe you post more details and the errors you got ? android can also be compiled on x86, btw...

thanks for asking that question

> Many of Android devices have ARM CPU.

but you're aware that this part of the code you've touched would only be used in Android for x86!?

> The problem I encountered was already several years ago and I don't remember what is the detailed problem. Maybe it is a compile error.

Maybe this PR shouldn't be merged until you have more details on how to reproduce a failure and how this patch fixes that issue. Okay?

Issue #306, opened 2017-10-02
RC5 - compiler warnings (Windows x64)


karel-m:
Visual C/C++ 2008
```
1>.\src\math\ltm_desc.c(92) : warning C4244: 'function' : conversion from 'ltc_mp_digit' to 'unsigned long', possible loss of data
1>.\src\math\ltm_desc.c(135) : warning C4244: 'function' : conversion from 'ltc_mp_digit' to 'mp_digit', possible loss of data
1>.\src\math\ltm_desc.c(217) : warning C4244: 'function' : conversion from 'ltc_mp_digit' to 'mp_digit', possible loss of data
1>.\src\math\ltm_desc.c(233) : warning C4244: 'function' : conversion from 'ltc_mp_digit' to 'mp_digit', possible loss of data
1>.\src\math\ltm_desc.c(249) : warning C4244: 'function' : conversion from 'ltc_mp_digit' to 'mp_digit', possible loss of data
1>.\src\math\ltm_desc.c(284) : warning C4244: 'function' : conversion from 'ltc_mp_digit' to 'mp_digit', possible loss of data
```

related to:
```
mp_cmp_d
mp_add_d
mp_sub_d
mp_mul_d
mp_mod_d
mp_set_int
```

@sjaeckel do we want to fix them?

karel-m:
IMO the following stuff is not completely correct:
```C
#ifdef ENDIAN_64BITWORD
typedef ulong64 ltc_mp_digit;
#else
typedef ulong32 ltc_mp_digit;
#endif 
```

Because with MS C/C++ compiler for x86_64:
* ltm's `mp_digit` is 32bit unsigned integer
* ltc's `ltc_mp_digit` is 64bit unsigned integer
* `unsigned long` is 32bit unsigned integer

sjaeckel:
Yeah, I also think we should fix this.

Not 100% sure, but I guess that's the correct question to resolve this: Why is `mp_digit` 32bits with MS C/C++ compiler for x86_64?

karel-m:
> Why is mp_digit 32bits with MS C/C++ compiler for x86_64?

Because ltm is not MSVC+64bit friendly. Even on 64bit Windows you have just:
```C
MP_PROVIDER  = LibTomMath
MP_DIGIT_BIT = 28
```

Frankly, it is so bad that "stock" ltm does not even compile with Visual C/C++ 2008, I have to use my patched version (you perhaps remember my lost battle regarding `stdint.h`).

[...]

PR #438, opened 2018-10-02
LTC_ECCSIG_RFC5656 SSH+ECDSA signature format

rmw42:
Implementation of SSH+ECDSA signature format for ECC signatures as defined in RFC5656.

Includes a SSH data model encode/decode (gated on #define LTC_SSH), and OID string<->Curve comparison helper function (also used for secp256k1 checks for LTC_ECCSIG_ETH27)

### Checklist

### DIFF
--- a/doc/crypt.tex
+++ b/doc/crypt.tex
@@ -5631,6 +5631,7 @@ The following signature formats are suported:
 \hline LTC\_ECCSIG\_ANSIX962 & ASN.1 encoded, ANSI X9.62 \\
 \hline LTC\_ECCSIG\_RFC7518 & raw R, S values as defined in RFC7518 \\
 \hline LTC\_ECCSIG\_ETH27 & raw R, S, V values (V has 27 added) \\
+\hline LTC\_ECCSIG\_RFC5656 & SSH+ECDSA format as defined in RFC5656 \\
 \hline
 \end{tabular}
 \end{center}
@@ -7792,7 +7793,7 @@ Currently LibTomCrypt will detect x86-32, x86-64, MIPS R5900, SPARC and SPARC64 
 There are also options you can specify from the \textit{tomcrypt\_custom.h} header file.
 
 \subsection{X memory routines}
-\index{XMALLOC}\index{XREALLOC}\index{XCALLOC}\index{XFREE}\index{XMEMSET}\index{XMEMCPY}\index{XMEMMOVE}\index{XMEMCMP}\index{XSTRCMP}
+\index{XMALLOC}\index{XREALLOC}\index{XCALLOC}\index{XFREE}\index{XMEMSET}\index{XMEMCPY}\index{XMEMMOVE}\index{XMEMCMP}\index{XSTRCMP}\index{XSTRNCPY}
 At the top of tomcrypt\_custom.h are a series of macros denoted as XMALLOC, XCALLOC, XREALLOC, XFREE, and so on.  They resolve to
 the name of the respective functions from the standard C library by default.  This lets you substitute in your own memory routines.
 If you substitute in your own functions they must behave like the standard C library functions in terms of what they expect as input and
--- a/helper.pl
+++ b/helper.pl
@@ -53,6 +53,8 @@ sub check_source {
       push @{$troubles->{unwanted_memmove}}, $lineno if $file =~ /^src\/.*\.c$/ && $l =~ /\bmemmove\s*\(/;
       push @{$troubles->{unwanted_memcmp}},  $lineno if $file =~ /^src\/.*\.c$/ && $l =~ /\bmemcmp\s*\(/;
       push @{$troubles->{unwanted_strcmp}},  $lineno if $file =~ /^src\/.*\.c$/ && $l =~ /\bstrcmp\s*\(/;
+      push @{$troubles->{unwanted_strcpy}},  $lineno if $file =~ /^src\/.*\.c$/ && $l =~ /\bstrcpy\s*\(/;
+      push @{$troubles->{unwanted_strncpy}}, $lineno if $file =~ /^src\/.*\.c$/ && $l =~ /\bstrncpy\s*\(/;
       push @{$troubles->{unwanted_clock}},   $lineno if $file =~ /^src\/.*\.c$/ && $l =~ /\bclock\s*\(/;
       push @{$troubles->{unwanted_qsort}},   $lineno if $file =~ /^src\/.*\.c$/ && $l =~ /\bqsort\s*\(/;
       push @{$troubles->{sizeof_no_brackets}}, $lineno if $file =~ /^src\/.*\.c$/ && $l =~ /\bsizeof\s*[^\(]/;
--- a/src/headers/tomcrypt_misc.h
+++ b/src/headers/tomcrypt_misc.h
@@ -154,6 +154,23 @@ int padding_pad(unsigned char *data, unsigned long length, unsigned long* padded
 int padding_depad(const unsigned char *data, unsigned long *length, unsigned long mode);
 #endif  /* LTC_PADDING */
 
+#ifdef LTC_SSH
+typedef enum ssh_data_type_ {
+   LTC_SSHDATA_BYTE,
+   LTC_SSHDATA_BOOLEAN,
+   LTC_SSHDATA_UINT32,
+   LTC_SSHDATA_UINT64,
+   LTC_SSHDATA_STRING,
+   LTC_SSHDATA_MPINT,
+   LTC_SSHDATA_NAMELIST,
+   LTC_SSHDATA_EOL
+} ssh_data_type;
+
+/* VA list handy helpers with tuples of <type, data> */
+int ssh_encode_sequence_multi(unsigned char *out, unsigned long *outlen, ...);
+int ssh_decode_sequence_multi(const unsigned char *in, unsigned long inlen, ...);
+#endif /* LTC_SSH */
+
 int compare_testvector(const void* is, const unsigned long is_len, const void* should, const unsigned long should_len, const char* what, int which);
 
 /* ref:         $Format:%D$ */
--- a/src/headers/tomcrypt_pk.h
+++ b/src/headers/tomcrypt_pk.h
@@ -251,7 +251,9 @@ typedef enum ecc_signature_type_ {
    /* raw R, S values */
    LTC_ECCSIG_RFC7518    = 0x1,
    /* raw R, S, V (+27) values */
-   LTC_ECCSIG_ETH27      = 0x2
+   LTC_ECCSIG_ETH27      = 0x2,
+   /* SSH + ECDSA signature format defined by RFC5656 */
+   LTC_ECCSIG_RFC5656    = 0x3,
 } ecc_signature_type;
 
 /** the ECC params provided */
--- a/src/misc/ssh/ssh_decode_sequence_multi.c
+++ b/src/misc/ssh/ssh_decode_sequence_multi.c
@@ -74,6 +74,7 @@ int ssh_decode_sequence_multi(const unsigned char *in, unsigned long inlen, ...)
 
          case LTC_SSHDATA_EOL:
             /* Should never get here */
+            size = (unsigned long)-1;
             err = CRYPT_INVALID_ARG;
             goto error;
       }
@@ -114,11 +115,11 @@ int ssh_decode_sequence_multi(const unsigned char *in, unsigned long inlen, ...)
          case LTC_SSHDATA_NAMELIST:
             sdata = va_arg(args, char*);
             bufsize = va_arg(args, unsigned long);
+            if (size >= bufsize) {
+               err = CRYPT_BUFFER_OVERFLOW;
+               goto error;
+            }
             if (size > 0) {
-               if (size >= bufsize) {
-                  err = CRYPT_BUFFER_OVERFLOW;
-                  goto error;
-               }
                XSTRNCPY(sdata, (const char *)in, size);
                sdata[size] = '\0'; /* strncpy doesn't NUL-terminate */
             } else {
@@ -130,6 +131,10 @@ int ssh_decode_sequence_multi(const unsigned char *in, unsigned long inlen, ...)
             vdata = va_arg(args, void*);
             if (size == 0) {
                if ((err = mp_set(vdata, 0)) != CRYPT_OK)                                                { goto error; }
+            } else if ((in[0] & 0x80) != 0) {
+               /* Negative number - not supported */
+               err = CRYPT_INVALID_PACKET;
+               goto error;
             } else {
                if ((err = mp_read_unsigned_bin(vdata, (unsigned char *)in, size)) != CRYPT_OK)          { goto error; }
             }
--- a/src/misc/ssh/ssh_decode_sequence_multi.c
+++ b/src/misc/ssh/ssh_decode_sequence_multi.c
@@ -53,6 +53,7 @@ int ssh_decode_sequence_multi(const unsigned char *in, unsigned long inlen, ...)
       }
 
       /* Calculate (or read) length of data */
+      size = (unsigned long)-1;
       switch (type) {
          case LTC_SSHDATA_BYTE:
          case LTC_SSHDATA_BOOLEAN:
@@ -74,7 +75,6 @@ int ssh_decode_sequence_multi(const unsigned char *in, unsigned long inlen, ...)
 
          case LTC_SSHDATA_EOL:
             /* Should never get here */
-            size = (unsigned long)-1;
             err = CRYPT_INVALID_ARG;
             goto error;
       }
--- a/src/misc/ssh/ssh_decode_sequence_multi.c
+++ b/src/misc/ssh/ssh_decode_sequence_multi.c
@@ -1,0 +1,156 @@
+/* LibTomCrypt, modular cryptographic library -- Tom St Denis
+ *
+ * LibTomCrypt is a library that provides various cryptographic
+ * algorithms in a highly modular and flexible manner.
+ *
+ * The library is free for all purposes without any express
+ * guarantee it works.
+ */
+#include "tomcrypt_private.h"
+#include <stdarg.h>
+
+/**
+   @file ssh_decode_sequence_multi.c
+   SSH data type representation as per RFC4251, Russ Williams
+*/
+
+#ifdef LTC_SSH
+
+/**
+  Decode a SSH sequence using a VA list
+  @param in     Data to decode
+  @param inlen  Length of buffer to decode
+  @remark <...> is of the form <type, data> (int, void*) except for string <type, data, size>
+  @return CRYPT_OK on success
+*/
+int ssh_decode_sequence_multi(const unsigned char *in, unsigned long inlen, ...)
+{
+   int           err;
+   va_list       args;
+   ssh_data_type type;
+   void          *vdata;
+   unsigned char *cdata;
+   char          *sdata;
+   ulong32       *u32data;
+   ulong64       *u64data;
+   unsigned long size, bufsize;
+
+   LTC_ARGCHK(in    != NULL);
+
+   /* Decode values from buffer */
+   va_start(args, inlen);
+   while ((type = (ssh_data_type)va_arg(args, int)) != LTC_SSHDATA_EOL) {
+      /* Size of length field */
+      if (type == LTC_SSHDATA_STRING ||
+          type == LTC_SSHDATA_NAMELIST ||
+          type == LTC_SSHDATA_MPINT)
+      {
+         /* Check we'll not read too far */
+         if (inlen < 4) {
+            err = CRYPT_BUFFER_OVERFLOW;
+            goto error;
+         }
+      }
+
+      /* Calculate (or read) length of data */
+      switch (type) {
+         case LTC_SSHDATA_BYTE:
+         case LTC_SSHDATA_BOOLEAN:
+            size = 1;
+            break;
+         case LTC_SSHDATA_UINT32:
+            size = 4;
+            break;
+         case LTC_SSHDATA_UINT64:
+            size = 8;
+            break;
+         case LTC_SSHDATA_STRING:
+         case LTC_SSHDATA_NAMELIST:
+         case LTC_SSHDATA_MPINT:
+            LOAD32H(size, in);
+            in += 4;
+            inlen -= 4;
+            break;
+
+         case LTC_SSHDATA_EOL:
+            /* Should never get here */
+            err = CRYPT_INVALID_ARG;
+            goto error;
+      }
+
+      /* Check we'll not read too far */
+      if (inlen < size) {
+         err = CRYPT_BUFFER_OVERFLOW;
+         goto error;
+      } else {
+         inlen -= size;
+      }
+
+      /* Read data */
+      switch (type) {
+         case LTC_SSHDATA_BYTE:
+            cdata = va_arg(args, unsigned char*);
+            *cdata = *in++;
+            break;
+         case LTC_SSHDATA_BOOLEAN:
+            cdata = va_arg(args, unsigned char*);
+            /*
+               The value 0 represents FALSE, and the value 1 represents TRUE.  All non-zero values MUST be
+               interpreted as TRUE; however, applications MUST NOT store values other than 0 and 1.
+            */
+            *cdata = (*in++)?1:0;
+            break;
+         case LTC_SSHDATA_UINT32:
+            u32data = va_arg(args, ulong32*);
+            LOAD32H(*u32data, in);
+            in += 4;
+            break;
+         case LTC_SSHDATA_UINT64:
+            u64data = va_arg(args, ulong64*);
+            LOAD64H(*u64data, in);
+            in += 8;
+            break;
+         case LTC_SSHDATA_STRING:
+         case LTC_SSHDATA_NAMELIST:
+            sdata = va_arg(args, char*);
+            bufsize = va_arg(args, unsigned long);
+            if (size > 0) {
+               if (size >= bufsize) {
+                  err = CRYPT_BUFFER_OVERFLOW;
+                  goto error;
+               }
+               XSTRNCPY(sdata, (const char *)in, size);
+               sdata[size] = '\0'; /* strncpy doesn't NUL-terminate */
+            } else {
+               *sdata = '\0';
+            }
+            in += size;
+            break;
+         case LTC_SSHDATA_MPINT:
+            vdata = va_arg(args, void*);
+            if (size == 0) {
+               if ((err = mp_set(vdata, 0)) != CRYPT_OK)                                                { goto error; }
+            } else {
+               if ((err = mp_read_unsigned_bin(vdata, (unsigned char *)in, size)) != CRYPT_OK)          { goto error; }
+            }
+            in += size;
+            break;
+
+         case LTC_SSHDATA_EOL:
+            /* Should never get here */
+            err = CRYPT_INVALID_ARG;
+            goto error;
+      }
+   }
+   err = CRYPT_OK;
+
+error:
+   va_end(args);
+   return err;
+}
+
+#endif
+
+/* ref:         $Format:%D$ */
+/* git commit:  $Format:%H$ */
+/* commit time: $Format:%ai$ */
--- a/src/misc/ssh/ssh_encode_sequence_multi.c
+++ b/src/misc/ssh/ssh_encode_sequence_multi.c
@@ -1,0 +1,168 @@
+/* LibTomCrypt, modular cryptographic library -- Tom St Denis
+ *
+ * LibTomCrypt is a library that provides various cryptographic
+ * algorithms in a highly modular and flexible manner.
+ *
+ * The library is free for all purposes without any express
+ * guarantee it works.
+ */
+#include "tomcrypt_private.h"
+#include <stdarg.h>
+
+/**
+   @file ssh_encode_sequence_multi.c
+   SSH data type representation as per RFC4251, Russ Williams
+*/
+
+#ifdef LTC_SSH
+
+/**
+  Encode a SSH sequence using a VA list
+  @param out    [out] Destination for data
+  @param outlen [in/out] Length of buffer and resulting length of output
+  @remark <...> is of the form <type, data> (int, void*)
+  @return CRYPT_OK on success
+*/
+int ssh_encode_sequence_multi(unsigned char *out, unsigned long *outlen, ...)
+{
+   int           err;
+   va_list       args;
+   unsigned long size;
+   ssh_data_type type;
+   void         *vdata;
+   const char   *sdata;
+   int           idata;
+   ulong32       u32data;
+   ulong64       u64data;
+
+   LTC_ARGCHK(out    != NULL);
+   LTC_ARGCHK(outlen != NULL);
+
+   /* Check values and calculate output size */
+   size = 0;
+   va_start(args, outlen);
+   while ((type = (ssh_data_type)va_arg(args, int)) != LTC_SSHDATA_EOL) {
+      switch (type) {
+         case LTC_SSHDATA_BYTE:
+         case LTC_SSHDATA_BOOLEAN: /* Both stored as 1 byte */
+            LTC_UNUSED_PARAM( va_arg(args, int) );
+            size++;
+            break;
+         case LTC_SSHDATA_UINT32:
+            LTC_UNUSED_PARAM( va_arg(args, ulong32) );
+            size += 4;
+            break;
+         case LTC_SSHDATA_UINT64:
+            LTC_UNUSED_PARAM( va_arg(args, ulong64) );
+            size += 8;
+            break;
+         case LTC_SSHDATA_STRING:
+         case LTC_SSHDATA_NAMELIST:
+            sdata = va_arg(args, char*);
+            size += 4;
+            size += strlen(sdata);
+            break;
+         case LTC_SSHDATA_MPINT:
+            vdata = va_arg(args, void*);
+            /* Calculate size */
+            size += 4;
+            if (mp_iszero(vdata) != LTC_MP_YES) {
+               size += mp_unsigned_bin_size(vdata);
+               if ((mp_count_bits(vdata) & 7) == 0) size++; /* Zero padding if high bit set */
+            }
+            break;
+
+         case LTC_SSHDATA_EOL: /* Should never get here */
+            err = CRYPT_INVALID_ARG;
+            goto error;
+      }
+   }
+   va_end(args);
+
+   /* Check we have sufficient space */
+   if (*outlen < size) {
+      *outlen = size;
+      err = CRYPT_BUFFER_OVERFLOW;
+      goto errornoargs;
+   }
+   *outlen = size;
+
+   /* Encode values into buffer */
+   va_start(args, outlen);
+   while ((type = (ssh_data_type)va_arg(args, int)) != LTC_SSHDATA_EOL) {
+      switch (type) {
+         case LTC_SSHDATA_BYTE:
+            idata = va_arg(args, int);
+
+            *out++ = (unsigned char)(idata & 255);
+            break;
+         case LTC_SSHDATA_BOOLEAN:
+            idata = va_arg(args, int);
+
+            /*
+               The value 0 represents FALSE, and the value 1 represents TRUE.  All non-zero values MUST be
+               interpreted as TRUE; however, applications MUST NOT store values other than 0 and 1.
+            */
+            *out++ = (idata)?1:0;
+            break;
+         case LTC_SSHDATA_UINT32:
+            u32data = va_arg(args, ulong32);
+            STORE32H(u32data, out);
+            out += 4;
+            break;
+         case LTC_SSHDATA_UINT64:
+            u64data = va_arg(args, ulong64);
+            STORE64H(u64data, out);
+            out += 8;
+            break;
+         case LTC_SSHDATA_STRING:
+         case LTC_SSHDATA_NAMELIST:
+            sdata = va_arg(args, char*);
+            size = strlen(sdata);
+            STORE32H(size, out);
+            out += 4;
+            XSTRNCPY((char *)out, sdata, size);
+            out += size;
+            break;
+         case LTC_SSHDATA_MPINT:
+            vdata = va_arg(args, void*);
+            if (mp_iszero(vdata) == LTC_MP_YES) {
+               STORE32H(0, out);
+               out += 4;
+            } else {
+               size = mp_unsigned_bin_size(vdata);
+               if ((mp_count_bits(vdata) & 7) == 0) {
+                  /* Zero padding if high bit set */
+                  STORE32H(size+1, out);
+                  out += 4;
+                  *out++ = 0;
+               } else {
+                  STORE32H(size, out);
+                  out += 4;
+               }
+               if ((err = mp_to_unsigned_bin(vdata, out)) != CRYPT_OK) {
+                  err = CRYPT_ERROR;
+                  goto error;
+               }
+               out += size;
+            }
+            break;
+
+         case LTC_SSHDATA_EOL: /* Should never get here */
+            err = CRYPT_INVALID_ARG;
+            goto error;
+      }
+   }
+   err = CRYPT_OK;
+
+error:
+   va_end(args);
+errornoargs:
+   return err;
+}
+
+#endif
+
+/* ref:         $Format:%D$ */
+/* git commit:  $Format:%H$ */
+/* commit time: $Format:%ai$ */
--- a/src/pk/ecc/ecc_ssh_ecdsa_encode_name.c
+++ b/src/pk/ecc/ecc_ssh_ecdsa_encode_name.c
@@ -1,0 +1,70 @@
+/* LibTomCrypt, modular cryptographic library -- Tom St Denis
+ *
+ * LibTomCrypt is a library that provides various cryptographic
+ * algorithms in a highly modular and flexible manner.
+ *
+ * The library is free for all purposes without any express
+ * guarantee it works.
+ */
+#include "tomcrypt_private.h"
+
+/**
+   @file ecc_ssh_ecdsa_encode_name.c
+   Curve/OID to SSH+ECDSA name string mapping per RFC5656
+   Russ Williams
+*/
+
+/**
+  Curve/OID to SSH+ECDSA name string mapping
+  @param buffer    [out] The destination for the name
+  @param buflen    [in/out] The max size and resulting size (including terminator) of the name
+  @param key       A public or private ECC key
+  @return CRYPT_OK if successful
+*/
+int ecc_ssh_ecdsa_encode_name(char *buffer, unsigned long *buflen, const ecc_key *key)
+{
+   char oidstr[64];
+   unsigned long oidlen = sizeof(oidstr);
+   unsigned long size = 0;
+   int err;
+
+   LTC_ARGCHK(buffer != NULL);
+   LTC_ARGCHK(buflen != NULL);
+   LTC_ARGCHK(key != NULL);
+
+   /* Get the OID of the curve */
+   if ((err = ecc_get_oid_str(oidstr, &oidlen, key)) != CRYPT_OK) goto error;
+
+   /* Check for three named curves: nistp256, nistp384, nistp521 */
+   if (XSTRCMP("1.2.840.10045.3.1.7", oidstr) == 0) {
+      /* nistp256 - secp256r1 - OID 1.2.840.10045.3.1.7 */
+      size = snprintf(buffer, *buflen, "ecdsa-sha2-nistp256");
+   }
+   else if (XSTRCMP("1.3.132.0.34", oidstr) == 0) {
+      /* nistp384 - secp384r1 - OID 1.3.132.0.34 */
+      size = snprintf(buffer, *buflen, "ecdsa-sha2-nistp384");
+   }
+   else if (XSTRCMP("1.3.132.0.35", oidstr) == 0) {
+      /* nistp521 - secp521r1 - OID 1.3.132.0.35 */
+      size = snprintf(buffer, *buflen, "ecdsa-sha2-nistp521");
+   } else {
+      /* Otherwise we use the OID... */
+      size = snprintf(buffer, *buflen, "ecdsa-sha2-%s", oidstr);
+   }
+
+   /* snprintf returns size that would have been written, but limits to buflen-1 chars plus terminator */
+   if (size >= *buflen) {
+      err = CRYPT_BUFFER_OVERFLOW;
+   } else {
+      err = CRYPT_OK;
+   }
+   *buflen = size + 1; /* the string length + NUL byte */
+
+error:
+   return err;
+}
+
+
+/* ref:         $Format:%D$ */
+/* git commit:  $Format:%H$ */
+/* commit time: $Format:%ai$ */
--- a/tests/common.h
+++ b/tests/common.h
@@ -23,6 +23,7 @@ extern prng_state yarrow_prng;
 #define SHOULD_FAIL(x) do { run_cmd((x) != CRYPT_OK ? CRYPT_OK : CRYPT_FAIL_TESTVECTOR, __LINE__, __FILE__, #x, NULL); } while (0)
 #endif
 
+#define COMPARE_TESTVECTOR(i, il, s, sl, wa, wi) do { DO(do_compare_testvector((i), (il), (s), (sl), (wa), (wi))); } while(0)
 
 #if !((defined(_WIN32) || defined(_WIN32_WCE)) && !defined(__GNUC__))
 #define LTC_TEST_READDIR
--- a/tests/ecc_test.c
+++ b/tests/ecc_test.c
@@ -37,6 +37,93 @@ static unsigned int sizes[] = {
 #endif
 };
 
+static const char* curvenames[] = {
+#ifdef LTC_ECC_SECP112R1
+   "SECP112R1", "ECC-112",
+   "secp112r1",              /* name is case-insensitive */
+   "S E C-P-1_1_2r1",        /* should pass fuzzy matching */
+#endif
+#ifdef LTC_ECC_SECP112R2
+   "SECP112R2",
+#endif
+#ifdef LTC_ECC_SECP128R1
+   "SECP128R1", "ECC-128",
+#endif
+#ifdef LTC_ECC_SECP128R2
+   "SECP128R2",
+#endif
+#ifdef LTC_ECC_SECP160R1
+   "SECP160R1", "ECC-160",
+#endif
+#ifdef LTC_ECC_SECP160R2
+   "SECP160R2",
+#endif
+#ifdef LTC_ECC_SECP160K1
+   "SECP160K1",
+#endif
+#ifdef LTC_ECC_BRAINPOOLP160R1
+   "BRAINPOOLP160R1",
+#endif
+#ifdef LTC_ECC_SECP192R1
+   "SECP192R1", "NISTP192", "PRIME192V1", "ECC-192", "P-192",
+#endif
+#ifdef LTC_ECC_PRIME192V2
+   "PRIME192V2",
+#endif
+#ifdef LTC_ECC_PRIME192V3
+   "PRIME192V3",
+#endif
+#ifdef LTC_ECC_SECP192K1
+   "SECP192K1",
+#endif
+#ifdef LTC_ECC_BRAINPOOLP192R1
+   "BRAINPOOLP192R1",
+#endif
+#ifdef LTC_ECC_SECP224R1
+   "SECP224R1", "NISTP224", "ECC-224", "P-224",
+#endif
+#ifdef LTC_ECC_SECP224K1
+   "SECP224K1",
+#endif
+#ifdef LTC_ECC_BRAINPOOLP224R1
+   "BRAINPOOLP224R1",
+#endif
+#ifdef LTC_ECC_PRIME239V1
+   "PRIME239V1",
+#endif
+#ifdef LTC_ECC_PRIME239V2
+   "PRIME239V2",
+#endif
+#ifdef LTC_ECC_PRIME239V3
+   "PRIME239V3",
+#endif
+#ifdef LTC_ECC_SECP256R1
+   "SECP256R1", "NISTP256", "PRIME256V1", "ECC-256", "P-256",
+#endif
+#ifdef LTC_ECC_SECP256K1
+   "SECP256K1",
+#endif
+#ifdef LTC_ECC_BRAINPOOLP256R1
+   "BRAINPOOLP256R1",
+#endif
+#ifdef LTC_ECC_BRAINPOOLP320R1
+   "BRAINPOOLP320R1",
+#endif
+#ifdef LTC_ECC_SECP384R1
+   "SECP384R1", "NISTP384", "ECC-384", "P-384",
+#endif
+#ifdef LTC_ECC_BRAINPOOLP384R1
+   "BRAINPOOLP384R1",
+#endif
+#ifdef LTC_ECC_BRAINPOOLP512R1
+   "BRAINPOOLP512R1",
+#endif
+#ifdef LTC_ECC_SECP521R1
+   "SECP521R1", "NISTP521", "ECC-521", "P-521",
+#endif
+};
+
+
 #ifdef LTC_ECC_SHAMIR
 static int _ecc_test_shamir(void)
 {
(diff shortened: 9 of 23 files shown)


rmw42:
@karel-m As promised :)

src/headers/tomcrypt_pk.h
```diff
@@ -321,6 +324,10 @@ int  ecc_recover_key(const unsigned char *sig,  unsigned long siglen,
                      const unsigned char *hash, unsigned long hashlen,
                      int recid, ecc_signature_type sigformat, ecc_key *key);
 
+#ifdef LTC_SSH
+int ecc_ssh_ecdsa_encode_name(char *buffer, unsigned long buflen, const ecc_key *key);
```
karel-m: Please move this part to `tomcrypt_private.h`

src/misc/ssh/ssh_test.c
```diff
@@ -0,0 +1,243 @@
+/* LibTomCrypt, modular cryptographic library -- Tom St Denis
```
karel-m: Please move `src/misc/ssh/ssh_test.c` to `tests/ssh_test.c` (+ do not forget to run `./updatemakes.sh`)

src/misc/ssh/ssh_test.c
```diff
@@ -0,0 +1,243 @@
...
+   return CRYPT_OK;
+}
+
+/**
+  LTC_SSH self-test
+  @return CRYPT_OK if successful, CRYPT_NOP if tests have been disabled.
+*/
+int ssh_test(void)
+{
+#ifndef LTC_TEST
```
karel-m: Once we move this stuff to `tests/ssh_test.c` there is no need for `#ifndef LTC_TEST`

src/pk/ecc/ecc_cmp_oid_str.c
```diff
@@ -0,0 +1,38 @@
...
+
+/** Compare curve's OID to constant string
+  @param oidstr [in] the desired OID in string format
+  @param key    The ECC key
+  Return        CRYPT_OK if the curve has the correct OID
+*/
+
+int ecc_cmp_oid_str(const char *oidstr, const ecc_key *key)
+{
+   char oid[32];
```
karel-m: Isn't it too small? I am for `oid[64]`.

src/pk/ecc/ecc_cmp_oid_str.c
```diff
@@ -0,0 +1,38 @@
...
+   unsigned long oidlen = 32;
+   int err;
+
+   LTC_ARGCHK(key != NULL);
+
+   err = pk_oid_num_to_str(key->dp.oid, key->dp.oidlen, oid, &oidlen);
+   if (err != CRYPT_OK) return err;
+   else if (!XSTRCMP(oidstr, oid)) return CRYPT_OK;
```
karel-m: I prefer to use explicit: `XSTRCMP(x, y) != 0` or `XSTRCMP(x, y) == 0` (applies also to other places)

src/pk/ecc/ecc_ssh_ecdsa_encode_name.c
```diff
@@ -0,0 +1,57 @@
...
+
+int ecc_ssh_ecdsa_encode_name(char *buffer, unsigned long buflen, const ecc_key *key)
+{
+   int err;
+
+   /* Check for three named curves: nistp256, nistp384, nistp521 */
+   if (ecc_cmp_oid_str("1.2.840.10045.3.1.7", key) == CRYPT_OK) {
+      /* nistp256 - secp256r1 - OID 1.2.840.10045.3.1.7 */
+
+      if (buflen <= 19) { err = CRYPT_BUFFER_OVERFLOW; goto error; }
```
karel-m: Maybe rather `buflen <= 20` - according to `man 3 strcpy`: The `strcpy()` function copies the string pointed to by src, **including the terminating null byte**.

src/pk/ecc/ecc_ssh_ecdsa_encode_name.c
```diff
@@ -0,0 +1,57 @@
...
+      strcpy(buffer, "ecdsa-sha2-nistp256");
+   }
+   else if (ecc_cmp_oid_str("1.3.132.0.34", key) == CRYPT_OK) {
+      /* nistp384 - secp384r1 - OID 1.3.132.0.34 */
+
+      if (buflen <= 19) { err = CRYPT_BUFFER_OVERFLOW; goto error; }
```
karel-m: `buflen <= 20`

src/pk/ecc/ecc_ssh_ecdsa_encode_name.c
```diff
@@ -0,0 +1,57 @@
...
+      strcpy(buffer, "ecdsa-sha2-nistp384");
+   }
+   else if (ecc_cmp_oid_str("1.3.132.0.35", key) == CRYPT_OK) {
+      /* nistp521 - secp521r1 - OID 1.3.132.0.35 */
+
+      if (buflen <= 19) { err = CRYPT_BUFFER_OVERFLOW; goto error; }
```
karel-m: `buflen <= 20`

src/pk/ecc/ecc_ssh_ecdsa_encode_name.c
```diff
@@ -0,0 +1,57 @@
...
+      strcpy(buffer, "ecdsa-sha2-nistp512");
```
karel-m: I thing it should be `nistp521`

src/pk/ecc/ecc_ssh_ecdsa_encode_name.c
```diff
@@ -0,0 +1,57 @@
...
+   }
+   else { /* Otherwise, we use the OID of the curve */
+      char oidstr[32];
+      unsigned long oidlen = 32;
+      if ((err = ecc_get_oid_str(oidstr, &oidlen, key)) != CRYPT_OK) goto error;
+
+      if (buflen <= strlen(oidstr)+11) { err = CRYPT_BUFFER_OVERFLOW; goto error; }
```
karel-m: Perhaps `if (buflen <= strlen(oidstr)+12)` or `if (buflen <= oidlen + 12)` - sprintf AFAIK writes terminating nul byte

src/pk/ecc/ecc_ssh_ecdsa_encode_name.c
```diff
@@ -0,0 +1,57 @@
...
+      strcpy(buffer, "ecdsa-sha2-nistp384");
+   }
+   else if (ecc_cmp_oid_str("1.3.132.0.35", key) == CRYPT_OK) {
+      /* nistp521 - secp521r1 - OID 1.3.132.0.35 */
+
+      if (buflen <= 19) { err = CRYPT_BUFFER_OVERFLOW; goto error; }
+      strcpy(buffer, "ecdsa-sha2-nistp512");
+   }
+   else { /* Otherwise, we use the OID of the curve */
+      char oidstr[32];
```
karel-m: Isn't it too small? I am for oidstr[64].

src/pk/ecc/ecc_ssh_ecdsa_encode_name.c
```diff
@@ -0,0 +1,57 @@
...
+ */
+#include "tomcrypt_private.h"
+
+/**
+   @file ecc_ssh_ecdsa_encode_name.c
+   Curve/OID to SSH+ECDSA name string mapping per RFC5656
+   Russ Williams
+*/
+
+int ecc_ssh_ecdsa_encode_name(char *buffer, unsigned long buflen, const ecc_key *key)
```
karel-m: To keep the style used on other places in libtomcrypt it should be `unsigned long *buflen` (but as it is a private function maybe we can leave it as it is, @sjaeckel ?)

karel-m (changes requested):

karel-m:
Great job!

karel-m:
And 2 warnings:
```
src/misc/ssh/ssh_test.c:59:5: error: no previous prototype for ‘_ssh_encoding_test’ [-Werror=missing-prototypes]
 int _ssh_encoding_test(void)
     ^~~~~~~~~~~~~~~~~~
src/misc/ssh/ssh_test.c:154:5: error: no previous prototype for ‘_ssh_decoding_test’ [-Werror=missing-prototypes]
 int _ssh_decoding_test(void)
     ^~~~~~~~~~~~~~~~~~
```

src/misc/ssh/ssh_test.c
```diff
@@ -0,0 +1,243 @@
...
+static const unsigned char nlist1[] = {0x00, 0x00, 0x00, 0x00};
+static const unsigned char nlist2[] = {0x00, 0x00, 0x00, 0x04, 0x7a, 0x6c, 0x69, 0x62};
+static const unsigned char nlist3[] = {0x00, 0x00, 0x00, 0x09, 0x7a, 0x6c, 0x69, 0x62, 0x2c, 0x6e, 0x6f, 0x6e, 0x65};
+
+
+/**
+  LTC_SSH encoding test
+  @return CRYPT_OK if successful
+*/
+int _ssh_encoding_test(void)
```
karel-m: `static int _ssh_encoding_test(void)`

src/misc/ssh/ssh_test.c
```diff
@@ -0,0 +1,243 @@
...
+   if (XMEMCMP(buffer, nlist3, buflen)) return CRYPT_FAIL_TESTVECTOR;
+
+   return CRYPT_OK;
+}
+
+/**
+  LTC_SSH decoding test
+  @return CRYPT_OK if successful
+*/
+int _ssh_decoding_test(void)
```
karel-m: `static int _ssh_decoding_test(void)`

src/misc/ssh/ssh_test.c
```diff
@@ -0,0 +1,243 @@
...
+   if (err != CRYPT_BUFFER_OVERFLOW) return CRYPT_FAIL_TESTVECTOR;
+
+
+   /* uint32 */
+   buflen = BUFSIZE;
+   zeromem(buffer, BUFSIZE);
+   DO(ssh_encode_sequence_multi(buffer, &buflen,
+                                LTC_SSHDATA_UINT32, 0x29b7f4aa,
+                                LTC_SSHDATA_EOL,    NULL));
+   if (XMEMCMP(buffer, uint32, buflen)) return CRYPT_FAIL_TESTVECTOR;
```
karel-m: `if (XMEMCMP(buffer, uint32, buflen) != 0)` + all similar places below

src/pk/ecc/ecc_cmp_oid_str.c
karel-m: and clang-tidy complains here:
```
src/pk/ecc/ecc_cmp_oid_str.c:30:4: error: do not use 'else' after 'return' [readability-else-after-return,-warnings-as-errors]
   else if (!XSTRCMP(oidstr, oid)) return CRYPT_OK;
   ^~~~~~~~
```

So maybe change it to: `if (XSTRCMP(oidstr, oid) == 0) return CRYPT_OK;`

src/pk/ecc/ecc_cmp_oid_str.c
```diff
@@ -0,0 +1,38 @@
...
+   else return CRYPT_INVALID_ARG;
```
karel-m: without `else`

src/pk/ecc/ecc_ssh_ecdsa_encode_name.c
```diff
@@ -0,0 +1,57 @@
...
+{
+   int err;
+
+   /* Check for three named curves: nistp256, nistp384, nistp521 */
+   if (ecc_cmp_oid_str("1.2.840.10045.3.1.7", key) == CRYPT_OK) {
+      /* nistp256 - secp256r1 - OID 1.2.840.10045.3.1.7 */
+
+      if (buflen <= 19) { err = CRYPT_BUFFER_OVERFLOW; goto error; }
+      strcpy(buffer, "ecdsa-sha2-nistp256");
```
karel-m: clang-tidy complains here (+ few other places in this file):
```
src/pk/ecc/ecc_ssh_ecdsa_encode_name.c:32:7: error: Call to function 'strcpy' is insecure as it does not provide bounding of the memory buffer. Replace unbounded copy functions with analogous functions that support length arguments such as 'strlcpy'. CWE-119 [clang-analyzer-security.insecureAPI.strcpy,-warnings-as-errors]
      strcpy(buffer, "ecdsa-sha2-nistp384");
      ^
```
So perhaps use something else than `strcpy` e.g. `strncpy` / `XMEMCPY`

src/headers/tomcrypt_pk.h
rmw42: OK

src/misc/ssh/ssh_test.c
rmw42: OK

src/pk/ecc/ecc_cmp_oid_str.c
rmw42: If an OID can be 64 chars then the SSH 64-char limit on the "ecdsa-sha2-XXXX" string would need to be higher. Happy to change it if there are super-long OIDs out there.

src/pk/ecc/ecc_cmp_oid_str.c
rmw42: Will do. And I'll go with the "goto error / single return" pattern so clang-tidy is happy.

src/pk/ecc/ecc_ssh_ecdsa_encode_name.c
rmw42: Yeah... my fingers are too familiar with powers of 2 ^_^;;

rmw42:
> Great job!

Cheers! :)

src/misc/ssh/ssh_test.c
rmw42: OK

src/pk/ecc/ecc_ssh_ecdsa_encode_name.c
rmw42: Yeah, the string is 19 chars long, so we need 19+1 to store it... if the buffer is 19 or fewer, it's too small. I figured "<= 19" was slightly clearer than "< 20" or "< 19 + 1", but none are particularly great.

What about:
```
   const char *name = "ecdsa-sha2-nistp256";
   if (buflen <= strlen(name)) { err = ... }
   XSTRCPY(buffer, name);
```

src/misc/ssh/ssh_test.c
rmw42: Done.

src/pk/ecc/ecc_ssh_ecdsa_encode_name.c
rmw42: Completely rewritten to use snprintf() and not need any hand-counted string lengths

src/pk/ecc/ecc_ssh_ecdsa_encode_name.c
rmw42: Replaced with snprintf() and combined with buffer length checking, so no more magic numbers.

src/pk/ecc/ecc_ssh_ecdsa_encode_name.c
rmw42: Done, and it writes out the length into *buflen.

src/pk/ecc/ecc_cmp_oid_str.c
rmw42: Changed to 64

src/pk/ecc/ecc_ssh_ecdsa_encode_name.c
rmw42: Changed to 64

rmw42:
Fixes in, and building now... hopefully that's all acceptable. Let me know if I missed anything! :)

karel-m:
gcc-7.3 with `-Wformat=2`
```
src/pk/ecc/ecc_ssh_ecdsa_encode_name.c: In function ‘ecc_ssh_ecdsa_encode_name’:
src/pk/ecc/ecc_ssh_ecdsa_encode_name.c:47:4: error: format not a string literal, argument types not checked [-Werror=format-nonliteral]
    size = snprintf(buffer, *buflen, pattern, oidstr);
    ^~~~
```
Can we somehow avoid this warning?

karel-m:
clang-tidy is without warnings 👍

rmw42:
> gcc-7.3 with `-Wformat=2`
> 
> ```
> src/pk/ecc/ecc_ssh_ecdsa_encode_name.c: In function ‘ecc_ssh_ecdsa_encode_name’:
> src/pk/ecc/ecc_ssh_ecdsa_encode_name.c:47:4: error: format not a string literal, argument types not checked [-Werror=format-nonliteral]
>     size = snprintf(buffer, *buflen, pattern, oidstr);
>     ^~~~
> ```
> Can we somehow avoid this warning?

Ah, damn, an I thought I was being so clever changing the format string... well, if it has to be a literal, I can get rid of `pattern` entirely, use four `snprintf()` calls in the different branches of the `if` but leave the buffer length check as-is at the end...

Give me 5 minutes ;-)

doc/crypt.tex
```diff
@@ -5254,6 +5254,17 @@ \subsection{Key Helper Functions}
 \textit{out} is where the ASCII output is placed, and the \textit{outlen}
 parameter is updated to hold the output OID size (including NUL byte).
 
+To check that the curve OID matches a NUL--terminated string the following function is provided:
+
+\index{ecc\_cmp\_oid\_str()}
+\begin{verbatim}
+int  ecc_cmp_oid_str(const char *oidstr,
```
sjaeckel: please don't make this public API or do you see a future requirement from a user PoV that this is needed?

If you agree, please remove this part and move the function declaration from `tomcrypt_pk.h` to `tomcrypt_private.h`

src/headers/tomcrypt_misc.h
```diff
@@ -154,6 +154,25 @@ int padding_pad(unsigned char *data, unsigned long length, unsigned long* padded
 int padding_depad(const unsigned char *data, unsigned long *length, unsigned long mode);
 #endif  /* LTC_PADDING */
 
+#ifdef LTC_SSH
+typedef enum ssh_data_type_ {
```
sjaeckel, line 158: ~~This should go to `tomcrypt_private.h`~~

Oh no, it shouldn't.

src/headers/tomcrypt_misc.h
```diff
@@ -154,6 +154,25 @@ int padding_pad(unsigned char *data, unsigned long length, unsigned long* padded
...
+   LTC_SSHDATA_STRING,
+   LTC_SSHDATA_MPINT,
+   LTC_SSHDATA_NAMELIST,
+   LTC_SSHDATA_EOL
+} ssh_data_type;
+
+int ssh_test(void);
+
+/* VA list handy helpers with tuples of <type, data> */
+int ssh_encode_sequence_multi(unsigned char *out, unsigned long *outlen, ...);
```
sjaeckel, line 170: instead this should probably be as follows?

`int ssh_encode_sequence_multi(unsigned char *out, unsigned long *outlen, int type, void* data, ...);`

not sure though if it would clutter the implementation too much!?

(same applies to `ssh_decode_sequence_multi()`)

src/misc/ssh/ssh_decode_sequence_multi.c
```diff
@@ -0,0 +1,136 @@
...
+                   ((unsigned long)in[2] << 8) |
+                   ((unsigned long)in[3]);
+            in += 4;
+            if (size > 0) {
+               if (size >= bufsize) {
+                  va_end(args);
+                  return CRYPT_BUFFER_OVERFLOW;
+               }
+               strncpy((char *)data, (const char *)in, size);
+               ((char *)data)[size] = 0; /* strncpy doesn't null terminate */
```
sjaeckel: string terminator is `'\0'`

src/misc/ssh/ssh_decode_sequence_multi.c
```diff
@@ -0,0 +1,136 @@
...
+            } else {
+               *(char *)data = 0;
```
sjaeckel: same

src/misc/ssh/ssh_decode_sequence_multi.c
```diff
@@ -0,0 +1,136 @@
...
+         case LTC_SSHDATA_UINT32:
+            size = ((unsigned long)in[0] << 24) |
+                   ((unsigned long)in[1] << 16) |
+                   ((unsigned long)in[2] << 8) |
+                   ((unsigned long)in[3]);
+            in += 4;
+            *(unsigned long *)data = size;
+            break;
+         case LTC_SSHDATA_UINT64:
+            size = ((unsigned long)in[0] << 56) |
```
sjaeckel: I doubt this works on all architectures [\*cough\*](https://github.com/libtom/libtomcrypt/issues/306#issuecomment-333765095) ;)

src/misc/ssh/ssh_encode_sequence_multi.c
```diff
@@ -0,0 +1,174 @@
...
+            *out++ = ((unsigned long)data)?1:0;
+            break;
+         case LTC_SSHDATA_UINT32:
+            size = (unsigned long)data;
+            *out++ = (unsigned char)((size >> 24) & 255);
+            *out++ = (unsigned char)((size >> 16) & 255);
+            *out++ = (unsigned char)((size >> 8) & 255);
+            *out++ = (unsigned char)(size & 255);
+            break;
+         case LTC_SSHDATA_UINT64:
```
sjaeckel, line 113: c.f. comment regarding `LTC_SSHDATA_UINT64` in `ssh_decode_sequence_multi()`

src/misc/ssh/ssh_decode_sequence_multi.c
```diff
@@ -0,0 +1,136 @@
...
+                   ((unsigned long)in[1] << 16) |
+                   ((unsigned long)in[2] << 8) |
+                   ((unsigned long)in[3]);
+            in += 4;
+            if (size > 0) {
+               if (size >= bufsize) {
+                  va_end(args);
+                  return CRYPT_BUFFER_OVERFLOW;
+               }
+               strncpy((char *)data, (const char *)in, size);
```
sjaeckel: First I thought why we can't use `strlcpy()`... then I remembered this standardization thingy... as we stay with this please introduce a macro `XSTRNCPY` in `tomcrypt_custom.h` (and also add it to the list in `crypt.tex` and `helper.pl` please)

src/misc/ssh/ssh_decode_sequence_multi.c
```diff
@@ -0,0 +1,136 @@
...
+            size = ((unsigned long)in[0]);
+            in++;
+            /*
+               The value 0 represents FALSE, and the value 1 represents TRUE.  All non-zero values MUST be
+               interpreted as TRUE; however, applications MUST NOT store values other than 0 and 1.
+            */
+            *(unsigned long *)data = (size)?1:0;
+            break;
+         case LTC_SSHDATA_UINT32:
+            size = ((unsigned long)in[0] << 24) |
```
sjaeckel: can't we use `LOAD32H()` here?

The same applies to all the other cases where we load some `unsigned long` (use `LOAD64H` where appropriate) resp. we could use `STORE{32,64}H()` when storing those values.

src/misc/ssh/ssh_decode_sequence_multi.c
sjaeckel: Btw. we also have types for exactly that purpose, `ulong32` and `ulong64` (see `tomcrypt_cfg.h` for details)

tests/ecc_test.c
```diff
@@ -37,6 +37,93 @@ static unsigned int sizes[] = {
 #endif
 };
 
+const char* names[] = {
```
sjaeckel: `static` and a more meaningful name please

tests/ssh_test.c
```diff
@@ -0,0 +1,237 @@
...
+   if (err != CRYPT_BUFFER_OVERFLOW) return CRYPT_FAIL_TESTVECTOR;
+
+
+   /* uint32 */
+   buflen = BUFSIZE;
+   zeromem(buffer, BUFSIZE);
+   DO(ssh_encode_sequence_multi(buffer, &buflen,
+                                LTC_SSHDATA_UINT32, 0x29b7f4aa,
+                                LTC_SSHDATA_EOL,    NULL));
+   if (XMEMCMP(buffer, uint32, buflen) != 0) return CRYPT_FAIL_TESTVECTOR;
```
sjaeckel: You can also use `DO(do_compare_testvector(...`, this prints some more information on why it failed.

... we could probably also add a macro ...

```
#define COMPARE_TESTVECTOR(i, il, s, sl, wa, wi) do { DO(do_compare_testvector((i), (il), (s), (sl), (wa), (wi))); } while(0)
```

src/misc/ssh/ssh_decode_sequence_multi.c
```diff
@@ -0,0 +1,136 @@
...
+
+      if (type == LTC_SSHDATA_EOL) {
+         break;
+      }
+
+      switch (type) {
+         case LTC_SSHDATA_BYTE:
+            size = ((unsigned long)in[0]);
+            in++;
+            *(unsigned long *)data = size;
```
sjaeckel: why is a byte stored as `unsigned long`?

src/misc/ssh/ssh_decode_sequence_multi.c
```diff
@@ -0,0 +1,136 @@
...
+                   ((unsigned long)in[2] << 8) |
+                   ((unsigned long)in[3]);
+            in += 4;
+            if (size == 0) {
+               if ((err = mp_set(data, 0)) != CRYPT_OK) {
+                  va_end(args);
+                  return CRYPT_ERROR;
+               }
+            } else {
+               if ((err = mp_read_unsigned_bin(data, (unsigned char *)in, size)) != CRYPT_OK) {
```
sjaeckel: ~~I just realized that the `src` argument in `unsigned_read()` in `tomcrypt_math.h` could be changed to `const unsigned char *src` as both ltm and tfm nowadays have `const*` arguments where applicable~~ oops ... ltm isn't released yet with those changes... \*cough\* https://github.com/libtom/libtommath/issues/122 \*cough\*

sjaeckel (changes requested):
Looks good!

:+1:

doc/crypt.tex
rmw42: I put it in the same places as `ecc_get_oid_str()`, but this probably shouldn't be public - it's unlikely that a user application will have created a key but not know what type it is, so this only really makes sense for internal checks.

Will move it and remove the docs.

src/misc/ssh/ssh_decode_sequence_multi.c
rmw42: OK

tests/ecc_test.c
rmw42: That's what it was called when it was function-local ;-)
I'll rename it to `curvenames` and make it static.

src/misc/ssh/ssh_decode_sequence_multi.c
rmw42: Managed to outthink myself considering register passing conventions/what this would be on the 'stack'/that it's a cast of `void *` when encoding... basically being silly. I'll make it something sane - uint8?

rmw42:
> in case of `CRYPT_BUFFER_OVERFLOW` we should return required buffer length

Done. I've refactored so it always returns the buffer length (i.e. string length + 1).

src/misc/ssh/ssh_decode_sequence_multi.c
rmw42: Ah, those were exactly what I was looking for. Cheers! Fixing now :)

src/misc/ssh/ssh_decode_sequence_multi.c
rmw42: Awesome. That's much tidier now :)

src/misc/ssh/ssh_decode_sequence_multi.c
rmw42: Done. I've also added a check for `strcpy` in `helper.pl` - not used anywhere in the code as far as I can tell, but presumably it should be alerted if someone introduces it.

src/misc/ssh/ssh_decode_sequence_multi.c
rmw42: `unsigned char` then

src/headers/tomcrypt_misc.h
rmw42: Not sure how well va_args lists handle restarting in the middle of the fixed parameters?

I'm going to alter the implementation so we don't have `void *data` being cast all over the place but explicitly collect the correct type from the stack.

sjaeckel: how about something like:

```
diff --git a/src/misc/ssh/ssh_decode_sequence_multi.c b/src/misc/ssh/ssh_decode_sequence_multi.c
index 2507609..3ace026 100644
--- a/src/misc/ssh/ssh_decode_sequence_multi.c
+++ b/src/misc/ssh/ssh_decode_sequence_multi.c
@@ -25,7 +25,7 @@
 */
-int ssh_decode_sequence_multi(const unsigned char *in, unsigned long inlen, ...)
+int ssh_decode_sequence_multi(const unsigned char *in, unsigned long inlen, int type, void *data, ...)
 {
    int           err;
-   ssh_data_type type;
-   void          *data;
+   ssh_data_type t;
+   void          *d;
    va_list       args;
@@ -36,7 +36,6 @@ int ssh_decode_sequence_multi(const unsigned char *in, unsigned long inlen, ...)
    /* Decode values from buffer */
-   va_start(args, inlen);
+   va_start(args, data);
+   t = type;
+   d = data;
    for (;;) {
-      type = (ssh_data_type)va_arg(args, int);
-      data = va_arg(args, void*);
-      LTC_UNUSED_PARAM(data);
 
@@ -46,3 +45,3 @@ int ssh_decode_sequence_multi(const unsigned char *in, unsigned long inlen, ...)
 
-      switch (type) {
+      switch (t) {
          case LTC_SSHDATA_BYTE:
@@ -122,9 +121,12 @@ int ssh_decode_sequence_multi(const unsigned char *in, unsigned long inlen, ...)
          case LTC_SSHDATA_EOL: /* Should never get here */
-            va_end(args);
-            return CRYPT_INVALID_ARG;
+            err = CRYPT_INVALID_ARG;
+            goto out;
        }
+      t = (ssh_data_type)va_arg(args, int);
+      d = va_arg(args, void*);
    }
+out:
    va_end(args);
 
-   return CRYPT_OK;
+   return err;
 }
```

[edit]: FYI yes, that's an incomplete patch ;)

rmw42: I've gone for something a bit more radical... just updating the tests to cover byte/boolean/uint64 now. Let me know what you think, I can always change it if you're not happy.

tests/ssh_test.c
rmw42: Done

helper.pl
```diff
@@ -53,6 +53,8 @@ sub check_source {
       push @{$troubles->{unwanted_memmove}}, $lineno if $file =~ /^src\/.*\.c$/ && $l =~ /\bmemmove\s*\(/;
       push @{$troubles->{unwanted_memcmp}},  $lineno if $file =~ /^src\/.*\.c$/ && $l =~ /\bmemcmp\s*\(/;
       push @{$troubles->{unwanted_strcmp}},  $lineno if $file =~ /^src\/.*\.c$/ && $l =~ /\bstrcmp\s*\(/;
+      push @{$troubles->{unwanted_strcpy}},  $lineno if $file =~ /^src\/.*\.c$/ && $l =~ /\bstrcpy\s*\(/;
```
sjaeckel, line 56: :+1:

src/headers/tomcrypt_misc.h
rmw42: Changed to a single return path and re-pushed.

src/misc/ssh/ssh_decode_sequence_multi.c
```diff
@@ -0,0 +1,116 @@
...
+   va_list       args;
+
+   LTC_ARGCHK(in    != NULL);
+
+   /* Decode values from buffer */
+   va_start(args, inlen);
+   for (;;) {
+      type = (ssh_data_type)va_arg(args, int);
+
+      if (type == LTC_SSHDATA_EOL) {
```
sjaeckel: TBH I prefer a switch-case over a long else-if chain for two reasons: 
1. it's cleaner to read IMO
2. if you do the switch on an enum and don't add a default case, you get immediate feedback from the compiler that the switch has unhandled cases of the enum

rmw42: I was mostly concerned about the need for a bunch of variable definitions at the top of the function which would each only be used by a single case. What about switch-case with braced code blocks? e.g.:
```
case LTC_SSHDATA_BYTE:
{
   unsigned char *data = va_arg(args, unsigned char *);
...
}
break;
```

sjaeckel: > I was mostly concerned about the need for a bunch of variable definitions at the top of the function which would each only be used by a single case. What about switch-case with braced code blocks?

I don't have a problem with both, but it's true - the braced code blocks worsen readability again

rmw42: I guess there's not going to be any way to make it pretty :(

Switch-case with variable definitions at the top of the loop, then? And I can put `type` into the formal arg list, change the `for` to `while (type != LTC_SSHDATA_EOL)` and put the reading of `type` at the end of the block. Sound good?

Any other changes I should make after that?

src/headers/tomcrypt_misc.h
rmw42: Ah... `warning: second argument to 'va_start' is not the last named parameter [-Wvarargs]`. Since it has to go through the loop twice, `type` can't be in the formal parameters for `ssh_encode_sequence_multi`...

src/misc/ssh/ssh_decode_sequence_multi.c
sjaeckel: > I guess there's not going to be any way to make it pretty :(

unfortunately not :/



> Switch-case with variable definitions at the top of the loop, then?

I'd say at the top of the function, not the loop



> And I can put `type` into the formal arg list, change the `for` to `while (type != LTC_SSHDATA_EOL)` and put the reading of `type` at the end of the block. Sound good?

why not change the signature to the proposed `int ssh_decode_sequence_multi(const unsigned char *in, unsigned long inlen, int type, void *data, ...)`? 

Should it really be a valid case to call `ssh_decode_sequence_multi(foo, bar, LTC_SSHDATA_EOL)`?

src/headers/tomcrypt_misc.h
sjaeckel: I guess you forgot to update the `va_start()` argument...

```
diff --git a/src/misc/ssh/ssh_decode_sequence_multi.c b/src/misc/ssh/ssh_decode_sequence_multi.c
index ca088f9..e44162a 100644
--- a/src/misc/ssh/ssh_decode_sequence_multi.c
+++ b/src/misc/ssh/ssh_decode_sequence_multi.c
@@ -25,6 +25,6 @@
 */
-int ssh_decode_sequence_multi(const unsigned char *in, unsigned long inlen, ...)
+int ssh_decode_sequence_multi(const unsigned char *in, unsigned long inlen, int type_, ...)
 {
    int           err;
-   ssh_data_type type;
+   ssh_data_type type = type_;
    va_list       args;
@@ -34,5 +34,4 @@ int ssh_decode_sequence_multi(const unsigned char *in, unsigned long inlen, ...)
    /* Decode values from buffer */
-   va_start(args, inlen);
+   va_start(args, type_);
    for (;;) {
-      type = (ssh_data_type)va_arg(args, int);
 
@@ -103,2 +102,3 @@ int ssh_decode_sequence_multi(const unsigned char *in, unsigned long inlen, ...)
       }
+      type = (ssh_data_type)va_arg(args, int);
    }
```
this compiles w/o warning and tests are OK
(please don't call it `type_` ;) )

src/misc/ssh/ssh_decode_sequence_multi.c
rmw42: I can't put `type` into the formal parameters because the encoding needs to go through the list twice, and can't be `va_start`ed from the middle of the formal parameters, so it would make the signatures of the two methods different :(
And `data` shouldn't really go in there because it then forces things to be `void*` pointers even when they should be integers :(

src/headers/tomcrypt_misc.h
rmw42: That's different to how `der_encode_sequence_multi` etc handle varargs, but I can do it that way if you want?

sjaeckel: OK, leave the signatures as they are, sorry for the noise :)

src/misc/ssh/ssh_decode_sequence_multi.c
sjaeckel: Both valid points I didn't see!

Again, sorry for the noise!

sjaeckel: But please, still change back to switch-case :)

src/headers/tomcrypt_misc.h
rmw42: No worries... it's annoying that there's no elegant solution, it's like an itch in the brain! :)

src/misc/ssh/ssh_decode_sequence_multi.c
rmw42: Done, and I've moved the params to the top of the function.

rmw42:
I added a new check in `ssh_decode_sequence_multi` to ensure we don't read past the end of the buffer, which caught a bug in the name-list decoding tests (I'd used `sizeof(string)` instead of `sizeof(nlist3)` - copy/paste bug).

rmw42:
OK, I've rebased to current develop branch (might need to do so again with the `ecc_verify_hash_ex` reversion), and I think that's all the things we've discussed... if there are any more issues, please let me know 😄

karel-m (approved):

src/headers/tomcrypt_misc.h
```diff
@@ -154,6 +154,25 @@ int padding_pad(unsigned char *data, unsigned long length, unsigned long* padded
...
+   LTC_SSHDATA_BOOLEAN,
+   LTC_SSHDATA_UINT32,
+   LTC_SSHDATA_UINT64,
+   LTC_SSHDATA_STRING,
+   LTC_SSHDATA_MPINT,
+   LTC_SSHDATA_NAMELIST,
+   LTC_SSHDATA_EOL
+} ssh_data_type;
+
+int ssh_test(void);
```
sjaeckel: oops, just found that this doesn't belong here anymore but in `tests/tomcrypt_test.h`

src/headers/tomcrypt_pk.h
```diff
@@ -251,7 +251,9 @@ typedef enum ecc_signature_type_ {
    /* raw R, S values */
    LTC_ECCSIG_RFC7518    = 0x1,
    /* raw R, S, V (+27) values */
-   LTC_ECCSIG_ETH27      = 0x2
+   LTC_ECCSIG_ETH27      = 0x2,
+   /* SSH + ECDSA signature format defined by RFC5656 */
+   LTC_ECCSIG_RFC5656    = 0x3
```
sjaeckel: I'm not 100% sure if this shouldn't be within `#ifdef LTC_SSH`?

if we're already at it, can you please directly add a `,` at the end :)

src/misc/ssh/ssh_decode_sequence_multi.c
```diff
@@ -0,0 +1,120 @@
...
+            break;
+
+         case LTC_SSHDATA_EOL:
+            /* Should never get here */
+            err = CRYPT_INVALID_ARG;
+            goto error;
+      }
+
+      /* Check we've not read too far */
+      if (in > sentinel) {
```
sjaeckel: :heart:

sjaeckel: shouldn't this kind of check be in each case where we read more than 1 byte?

src/misc/ssh/ssh_encode_sequence_multi.c
```diff
@@ -0,0 +1,167 @@
...
+         case LTC_SSHDATA_EOL: /* Should never get here */
+            err = CRYPT_INVALID_ARG;
+            goto error;
+      }
+   }
+   va_end(args);
+
+   /* Check we have sufficient space */
+   if (*outlen < size) {
+      err = CRYPT_BUFFER_OVERFLOW;
```
sjaeckel, line 85: didn't we say we should return the required size in this error-case?

src/pk/ecc/ecc_cmp_oid_str.c
```diff
@@ -0,0 +1,40 @@
...
+
+int ecc_cmp_oid_str(const char *oidstr, const ecc_key *key)
+{
+   char oid[64];
+   unsigned long oidlen = sizeof(oid);
+   int err;
+
+   LTC_ARGCHK(key != NULL);
+
+   if ((err = pk_oid_num_to_str(key->dp.oid, key->dp.oidlen, oid, &oidlen)) != CRYPT_OK)                { goto error; }
```
sjaeckel: if we merge this after #404 there's going to be `int pk_oid_cmp_with_ulong(const char *o1, const unsigned long *o2, unsigned long o2size);` ;)

tests/common.h
```diff
@@ -17,10 +17,12 @@ extern prng_state yarrow_prng;
 #define DO(x) do { fprintf(stderr, "%s:\n", #x); run_cmd((x), __LINE__, __FILE__, #x, NULL); } while (0)
 #define DOX(x, str) do { fprintf(stderr, "%s - %s:\n", #x, (str)); run_cmd((x), __LINE__, __FILE__, #x, (str)); } while (0)
 #define SHOULD_FAIL(x) do { fprintf(stderr, "%s:\n", #x); run_cmd((x) != CRYPT_OK ? CRYPT_OK : CRYPT_FAIL_TESTVECTOR, __LINE__, __FILE__, #x, NULL); } while (0)
+#define COMPARE_TESTVECTOR(i, il, s, sl, wa, wi) do { DO(do_compare_testvector((i), (il), (s), (sl), (wa), (wi))); } while(0)
 #else
 #define DO(x) do { run_cmd((x), __LINE__, __FILE__, #x, NULL); } while (0)
 #define DOX(x, str) do { run_cmd((x), __LINE__, __FILE__, #x, (str)); } while (0)
 #define SHOULD_FAIL(x) do { run_cmd((x) != CRYPT_OK ? CRYPT_OK : CRYPT_FAIL_TESTVECTOR, __LINE__, __FILE__, #x, NULL); } while (0)
+#define COMPARE_TESTVECTOR(i, il, s, sl, wa, wi) do { DO(do_compare_testvector((i), (il), (s), (sl), (wa), (wi))); } while(0)
```
sjaeckel: that define is only required once after the `#endif`

sjaeckel (commented):
Sorry for more nitpicks, minor comments, but there's also one important thing I only spotted now (regarding the length-checks when decoding)

src/headers/tomcrypt_misc.h
rmw42: Moved.

src/headers/tomcrypt_pk.h
rmw42: I think it makes sense to _define_ the signature format in all cases, whether or not the implementation is built into the library. If `LTC_SSH` isn't defined, then it'll fall through to the "unknown format" errors in the functions.

I think a trailing comma in an enum results in a compiler warning. Not sure, though?

src/misc/ssh/ssh_decode_sequence_multi.c
rmw42: Probably, but it would make the implementation a lot uglier. It'd mean we couldn't use the `LOAD32H` macros as they don't do checks after each byte.

The main reason I could see for doing that is that there might be information leaked if it can read past the end of the buffer... but only by someone with the ability to change the code, who could read from that memory anyway.

It's in the loop, so it will be checked after each object read, whether that's 4 or 8 or however many bytes. Personally, I think that's OK, but I'm happy to move it if you disagree.

src/misc/ssh/ssh_encode_sequence_multi.c
rmw42: Oops. Changed another case (`ecc_ssh_decode_name`) but missed this one. Fixing it now.

src/pk/ecc/ecc_cmp_oid_str.c
rmw42: I can drop `ecc_cmp_oid_str` entirely, then. Turned out that it was only needed for the secp256k1 checks for `LTC_ECCSIG_ETH27`. Let me know if we want to do that.

tests/common.h
rmw42: OK

src/headers/tomcrypt_pk.h
sjaeckel: > I think it makes sense to _define_ the signature format in all cases

agreed!



> I think a trailing comma in an enum results in a compiler warning. Not sure, though?

shouldn't on modern compilers :) (e.g. the `ltc_asn1_XX` enums have it as well)

src/misc/ssh/ssh_decode_sequence_multi.c
sjaeckel: > it would make the implementation a lot uglier

sure, but it's required IMO as the following isn't the only bad thing that could happen IMO

> The main reason I could see for doing that is that there might be information leaked if it can read past the end of the buffer...

... but if you put the buffer at the end of a page it could also segfault or not?

And isn't it even UB to read past an allocated buffer?

src/pk/ecc/ecc_cmp_oid_str.c
sjaeckel: We keep it for now as is I'd say as the other PR seems to take some more iterations as it's not entirely ready yet.

sjaeckel: Even better, simply cherry-pick those commits and whoever gets merged first will have them and the second will be rebased and the commits removed :)  Git FTW

c0407783ac383f9509a62dde99de4b6606c4d4f7
c81e1cb1dc8b36286233605c3a10c47584ec0e79
6abfab4d03389ccfe2c31f7ead84c706befc3e60

src/headers/tomcrypt_pk.h
rmw42: > shouldn't on modern compilers

Yup. I thought I'd seen this somewhere recently, but apparently not - it compiles cleanly.

src/pk/ecc/ecc_cmp_oid_str.c
rmw42: Took a while to figure out the syntax - thought I'd have to name the branch I was pulling from - but that was remarkably painless! 🍒 

I'll get rid of `ecc_cmp_oid_str` now.

src/misc/ssh/ssh_decode_sequence_multi.c
rmw42: Well, the same would apply to _anything_ that reads from a buffer if the specified length is greater than the size of the buffer... it shouldn't segfault unless someone has done something weird with the virtual memory page protection. Even if it does, if an application asks us to read 1,000,000 bytes from a 1,000 byte buffer, or passes us a pointer which cannot be read from at all, is a segfault not an appropriate response?

Instead of a sentinel at the end, I could keep track of the bytes available to read, and put a second switch-case in the loop to calculate the size which will be read? That's still going to be somewhat delicate around the variable-length string/mpint cases...

rmw42: OK, I've ended up with three switch-cases:
- the length field of the object (if needed)
- read/calculate the length of the object
- read the object itself

Should now be impossible for `ssh_decode_sequence_multi` to read beyond the end of the input.

sjaeckel: > is a segfault not an appropriate response?

sure it is and if we can prevent it by properly checking indicators like the required vs available length we should do it IMO.



> Instead of a sentinel at the end, I could keep track of the bytes available to read, and put a second switch-case in the loop to calculate the size which will be read? That's still going to be somewhat delicate around the variable-length string/mpint cases...

how about

```
diff --git a/src/misc/ssh/ssh_decode_sequence_multi.c b/src/misc/ssh/ssh_decode_sequence_multi.c
index 2375df8..aa171b3 100644
--- a/src/misc/ssh/ssh_decode_sequence_multi.c
+++ b/src/misc/ssh/ssh_decode_sequence_multi.c
@@ -37,14 +37,21 @@ int ssh_decode_sequence_multi(const unsigned char *in, unsigned long inlen, ...)
    unsigned long size, bufsize;
+   long len;
 
    LTC_ARGCHK(in    != NULL);
+   LTC_ARGCHK(inlen <= LONG_MAX);
 
    /* Decode values from buffer */
+   len = inlen;
    va_start(args, inlen);
    while ((type = (ssh_data_type)va_arg(args, int)) != LTC_SSHDATA_EOL) {
+      err = CRYPT_BUFFER_OVERFLOW;
       switch (type) {
          case LTC_SSHDATA_BYTE:
+            if (len < 1) goto error;
             cdata = va_arg(args, unsigned char*);
             *cdata = *in++;
+            len--;
             break;
          case LTC_SSHDATA_BOOLEAN:
+            if (len < 1) goto error;
             cdata = va_arg(args, unsigned char*);
@@ -55,4 +62,6 @@ int ssh_decode_sequence_multi(const unsigned char *in, unsigned long inlen, ...)
             *cdata = (*in++)?1:0;
+            len--;
             break;
          case LTC_SSHDATA_UINT32:
+            if (len < 4) goto error;
             u32data = va_arg(args, ulong32*);
@@ -60,4 +69,6 @@ int ssh_decode_sequence_multi(const unsigned char *in, unsigned long inlen, ...)
             in += 4;
+            len -= 4;
             break;
          case LTC_SSHDATA_UINT64:
+            if (len < 8) goto error;
             u64data = va_arg(args, ulong64*);
@@ -65,2 +76,3 @@ int ssh_decode_sequence_multi(const unsigned char *in, unsigned long inlen, ...)
             in += 8;
+            len -= 8;
             break;
@@ -73,6 +85,3 @@ int ssh_decode_sequence_multi(const unsigned char *in, unsigned long inlen, ...)
             if (size > 0) {
-               if (size >= bufsize) {
-                  err = CRYPT_BUFFER_OVERFLOW;
-                  goto error;
-               }
+               if (size >= bufsize) goto error;
                XSTRNCPY(sdata, (const char *)in, size);
@@ -83,4 +92,6 @@ int ssh_decode_sequence_multi(const unsigned char *in, unsigned long inlen, ...)
             in += size;
+            len -= size;
             break;
          case LTC_SSHDATA_MPINT:
+            if (len < 4) goto error;
             vdata = va_arg(args, void*);
@@ -91,2 +102,3 @@ int ssh_decode_sequence_multi(const unsigned char *in, unsigned long inlen, ...)
             } else {
+               if (len < size) goto error;
                if ((err = mp_read_unsigned_bin(vdata, (unsigned char *)in, size)) != CRYPT_OK)          { goto error; }
@@ -94,2 +106,3 @@ int ssh_decode_sequence_multi(const unsigned char *in, unsigned long inlen, ...)
             in += size;
+            len -= size;
             break;
```

rmw42: Yeah, that'd work too, but some of the variable-length ones would need extra checks.

How's the triple-switch version? It's a bit longer but maybe clearer intent...

sjaeckel: > How's the triple-switch version?

I'm not too much a fan of it, @karel-m your thoughts?

rmw42:
I've updated the length-check to remove one of the switches and keep track of the length remaining, rather than the end pointer, and rebased to the `develop` branch to fix the merge issues around cherry-picking `pk_oid_cmp_with_ulong`.

sjaeckel:
@mkj did you have a look at this?

src/misc/ssh/ssh_decode_sequence_multi.c
```diff
@@ -0,0 +1,156 @@
...
+            bufsize = va_arg(args, unsigned long);
+            if (size > 0) {
+               if (size >= bufsize) {
+                  err = CRYPT_BUFFER_OVERFLOW;
+                  goto error;
+               }
+               XSTRNCPY(sdata, (const char *)in, size);
+               sdata[size] = '\0'; /* strncpy doesn't NUL-terminate */
+            } else {
+               *sdata = '\0';
```
mkj, line 126: Invalid write if size==0 && bufsize==0

src/misc/ssh/ssh_decode_sequence_multi.c
```diff
@@ -0,0 +1,156 @@
...
+            }
+            in += size;
+            break;
+         case LTC_SSHDATA_MPINT:
+            vdata = va_arg(args, void*);
+            if (size == 0) {
+               if ((err = mp_set(vdata, 0)) != CRYPT_OK)                                                { goto error; }
+            } else {
+               if ((err = mp_read_unsigned_bin(vdata, (unsigned char *)in, size)) != CRYPT_OK)          { goto error; }
```
mkj, line 139: Check that the top bit isn't set, that implies a negative value
https://tools.ietf.org/html/rfc4251#page-8

mkj (commented):
Looks OK, couple of small things I've noted.

Unrelated to the patch, I'm curious where SSH signature encoding is being used externally to the SSH protocol?

src/misc/ssh/ssh_decode_sequence_multi.c
rmw42: Good catch. Done.

src/misc/ssh/ssh_decode_sequence_multi.c
rmw42: Moved the `size >= bufsize` check up, out of the `if (size > 0)` case. That way if the string is 0 bytes long - an empty string - we still need a buffer of at least 1 byte to output a C-style NUL-terminated string.

rmw42:
> Unrelated to the patch, I'm curious where SSH signature encoding is being used externally to the SSH protocol?

I don't think it is used anywhere else... @karel-m asked for this so you could avoid having to transform another signature format in Dropbear.

karel-m:
@mkj the idea was mine. During my (unfinished) work - https://github.com/karel-m/dropbear/commit/4530ff68975932680d674a33ea477fa7afc79ade - on a PR adopting Dropbear to the new libtomcrypt ecc_* stuff I saw the dark magic in dropbear's `ecc.c` related to ECDSA signatures which lead me to the idea that libtomcrypt should do a better job supporting our "prominent customer" (= dropbear).

sjaeckel (approved):

mkj:
Ah right. I'm not sure it really makes sense for libtomcrypt to be parsing SSH packets when it's alongside another parser anyway. 

I think it would work fine if there were just a 
```
int ecc_verify_hash_raw(void *r, void *s,
    const unsigned char *hash, unsigned long hashlen,
    int *stat, const ecc_key *key)
```

etc which could be called? Dropbear needs to do some work outside anyway to know which hash_desc to use.

Currently Dropbear's bundled libtomcrypt skips LTC_DER since it isn't necessary (I just treat them as opaque blobs in a few places) - it avoids some code size. It would be nice to avoid having to pull that in too, maybe the `dropbear_ecc_curve.oid` in the new branch should just be a long array to memcmp()?

sjaeckel:
> maybe the `dropbear_ecc_curve.oid` in the new branch should just be a long array to memcmp()?

The plan is to make all OID appearances per default string-based (providing still the possibility to use the ulong-array approach to de-/encode via the DER API). Would that be fine for you as well?

rmw42:
Rebased to current develop branch

karel-m:
Ad @mkj 's comments

> I think it would work fine if there were just a `int ecc_verify_hash_raw(void *r, void *s, ...)`

I do not like this. This will expose internal libtomcrypt guts (pointers to bignum structures) which we do not have yet in any public API function. Plus there is of course a completely new bunch of troubles related to the fact that math provider is not always libtommath.

> Currently Dropbear's bundled libtomcrypt skips LTC_DER since it isn't necessary...

This is a valid point. The current all-in-one `ecc_sign_hash_ex` and `ecc_verify_hash_ex` forces anyone using ECDSA to link the whole ASN.1 stuff to their binary even if they are using just RFC 7518 or RFC 5656 (ssh2) signature format. IMO it is worth rethinking in a separate PR.

@sjaeckel can we merge this PR? (to me it looks good)

rmw42:
> The current all-in-one `ecc_sign_hash_ex` and `ecc_verify_hash_ex` forces anyone using ECDSA to link the whole ASN.1 stuff to their binary

We could put that branch of the conditional on `#ifdef LTC_DER` so `LTC_ECCSIG_ANSIX962` is only available as a signature format if the library is built with DER support and returns an error otherwise.

karel-m:
This will not handle a static linking scenario when you have a static library built with all ASN.1/DER support but you want to just use RFC 7518 ECDSA signatures. I think we should somehow split `ecc_sign_hash_ex.c` and `ecc_verify_hash_ex.c` into more `.c` files.

rmw42:
Cheers! 👍

PR #443, opened 2018-10-04
Less strict ecc_verify_hash_ex (as it was before ecc_recover_key)

karel-m:
The recent ECDSA related changes (ecc_recover_key & co.) made `ecc_verify_hash_ex` a little bit more stricter (I know, it was my idea).

However, it turned out that 2 tests from wycheproof test suite started to fail.

Both cases were in category "acceptable" (so basically it is fine to reject them but also to accept them).

The 100% correct signature looks like this (hexa):
```
00000000000000000000000000000000000000000000000000000000000000050000000000000000000000000000000000000000000000000000000000000001
```

The "acceptable" signature looks like this (hexa):
```
0501
```

So with this PR we will accept the "acceptable" signature as we did before ecc_recover_key.

I have added both failing test cases to `ecc_test.c`


karel-m:
Cc: @rmw42

tests/ecc_test.c
```diff
@@ -117,7 +117,7 @@ static int _ecc_test_shamir(void)
 }
 #endif
 
-static int _ecc_issue108(void)
+static int _ecc_issue108(void) /* https://github.com/libtom/libtomcrypt/issues/108 */
```
sjaeckel: can you please put this comment above the function

tests/ecc_test.c
```diff
@@ -151,6 +151,56 @@ static int _ecc_issue108(void)
    return err;
 }
 
+static int _ecc_issue443(void) /* https://github.com/libtom/libtomcrypt/issues/443 */
```
sjaeckel: same

src/pk/ecc/ecc_verify_hash.c
```diff
@@ -76,11 +76,11 @@ int ecc_verify_hash_ex(const unsigned char *sig,  unsigned long siglen,
    }
    else if (sigformat == LTC_ECCSIG_RFC7518) {
       /* RFC7518 format - raw (r,s) */
-      i = mp_unsigned_bin_size(key->dp.order);
-      if (siglen != (2*i)) {
```
sjaeckel, line 80: I also preferred this version of the check TBH

src/pk/ecc/ecc_verify_hash.c
```diff
@@ -76,11 +76,11 @@ int ecc_verify_hash_ex(const unsigned char *sig,  unsigned long siglen,
...
+      if (siglen % 2) {
```
sjaeckel: probably revert again to the original version `if ((siglen % 2) == 1)`? I'm not entirely sure why, but I don't like the implicit to-boolean conversion of C anymore...

rmw42:
Ah... this is the checking the length of the r/s values, rather than just dividing the buffer length in 2? Fair enough. There's no great solution in a format which doesn't mark the lengths :(

sjaeckel (approved):

tests/ecc_test.c
karel-m: done

src/pk/ecc/ecc_verify_hash.c
karel-m: Done. I believe clang-tidy will like `if ((siglen % 2) == 1)` more as well.

sjaeckel:
> The 100% correct signature looks like this (hexa):
> 
> ```
> 00000000000000000000000000000000000000000000000000000000000000050000000000000000000000000000000000000000000000000000000000000001
> ```
> The "acceptable" signature looks like this (hexa):
> 
> ```
> 0501
> ```

I just thought that this check could then still be done after the `mp_read_unsigned_bin()` or would this be redundant with the "check for zero" tests after init of `r` and `s`?

src/pk/ecc/ecc_verify_hash.c
rmw42: Yeah, I'm not convinced that {0x05, 0x01} should be a valid signature blob for ECDSA, let alone that it should be in a security/proofing library.

rmw42:
A rather more radical solution would be to throw the test out entirely, and consider every possible partition of the input string, one by one. That way we make no assumptions at all about the padded size of `r` and `s`, and could support `0501` or `0000050001` or `05000000000000000001`...

Would probably want to be a flags parameter to `ecc_verify_hash_ex_ex`, for strict/normal/ludicrously_flexible... and I can't see any reason why anyone would want that functionality...

rmw42 (approved):

src/pk/ecc/ecc_verify_hash.c
sjaeckel: I also don't like that it could be a valid signature, but it's a valid edge-case... :/ 

... and history has proven that there are always implementations who do worse than your implementation so this (`0x000..05...`) is ~~a good~~ another example of Postel's Law IMO :)

karel-m:
I do not have a strong opinion on this. Let's give it a day or two to think it over.

sjaeckel:
> I do not have a strong opinion on this. Let's give it a day or two to think it over.

IMO it's good as it is now

rmw42:
Yeah, agreed. I'd say there's a problem with wycheproof but, until/unless they fix it, the test needs to be the way it is now.

karel-m:
In fact there is newer wycheproof test suite than I am using - https://github.com/google/wycheproof

rmw42:
> In fact there is newer wycheproof test suite than I am using - https://github.com/google/wycheproof

It's still got that test in there, though :(

karel-m:
BTW my quick check - there is at least 20 new ECDSA edge-cases that we do not implement correctly 😞

PR #445, opened 2018-10-07
pbes: some linting

fperrad:
another point : the definition of `oid_to_pbes` is duplicated
https://github.com/libtom/libtomcrypt/blob/develop/src/misc/pbes/pbes1.c#L59-L62
https://github.com/libtom/libtomcrypt/blob/develop/src/misc/pbes/pbes2.c#L40-L43

### DIFF
--- a/src/misc/pbes/pbes.c
+++ b/src/misc/pbes/pbes.c
@@ -20,7 +20,7 @@
 */
 int pbes_decrypt(const pbes_arg  *arg, unsigned char *dec_data, unsigned long *dec_size)
 {
-   int err, hid = -1, cid = -1;
+   int err, hid, cid;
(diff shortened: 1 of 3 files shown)


sjaeckel:
> another point : the definition of `oid_to_pbes` is duplicated

right, it's nowhere used besides the two files w/o any usage possibility outside, so I thought it doesn't really belong in a header file... (besides probably `src/misc/pbes.h` which would be against our currently unwritten rules :) )

sjaeckel:
I understand that most likely @eduardsui could have some problems when he re-creates his amalgamation... and otherwise it's perfectly fine to re-define a typedef if the second is exactly the same :P

PR #446, opened 2018-10-07
Wycheproof failing ECC tests

### LINKED ISSUE #116: ecc_verify_hash_raw fails when not using Shamir method and provided hash is all zeros
The math for handling the ECC verify with a hash with null characters, such as {0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0} and hashlen = 20. This results in an invalid signature verification (v != r). This only happens if the traditional (non Shamir) method is used to calculate "u1*mG + u2*mQ = mG".

This happens when calling "ecc_verify_hash_raw" and not using Shamir method (ecc_mul2add == NULL) and provided hash is all zeros.

Has anyone seen this or have any idea how to resolve the math?

karel-m:
Let's have a valid signature like:

      0x30,0x45,0x02,0x20,0x2b,0xa3,0xa8,0xbe,0x6b,0x94,0xd5,0xec,0x80,0xa6,0xd9,0xd1,0x19,0x0a,
      0x43,0x6e,0xff,0xe5,0x0d,0x85,0xa1,0xee,0xe8,0x59,0xb8,0xcc,0x6a,0xf9,0xbd,0x5c,0x2e,0x18,
      0x02,0x21,0x00,0xb3,0x29,0xf4,0x79,0xa2,0xbb,0xd0,0xa5,0xc3,0x84,0xee,0x14,0x93,0xb1,0xf5,
      0x18,0x6a,0x87,0x13,0x9c,0xac,0x5d,0xf4,0x08,0x7c,0x13,0x4b,0x49,0x15,0x68,0x47,0xdb,

Now let's patch the first `0x30` to `0x31` like:

      0x31,0x45,0x02,0x20,0x2b,0xa3,0xa8,0xbe,0x6b,0x94,0xd5,0xec,0x80,0xa6,0xd9,0xd1,0x19,0x0a,
      0x43,0x6e,0xff,0xe5,0x0d,0x85,0xa1,0xee,0xe8,0x59,0xb8,0xcc,0x6a,0xf9,0xbd,0x5c,0x2e,0x18,
      0x02,0x21,0x00,0xb3,0x29,0xf4,0x79,0xa2,0xbb,0xd0,0xa5,0xc3,0x84,0xee,0x14,0x93,0xb1,0xf5,
      0x18,0x6a,0x87,0x13,0x9c,0xac,0x5d,0xf4,0x08,0x7c,0x13,0x4b,0x49,0x15,0x68,0x47,0xdb,

We still verify the patched signature whereas wycheproof says that the patched signature should be considered invalid.

This PR contains just a failing test not the fix.


karel-m:
I have added one more wycheproof failing test labeled "Edge case for Shamir multiplication" - in this case we reject a signature that is valid (which is less dangerous).

karel-m:
Added to more cases (both ASN.1 length encoding) when we accept signature that wycheproof guys consider invalid.

tests/ecc_test.c
```diff
@@ -291,6 +291,177 @@ static int _ecc_issue443_447(void)
...
+      0x77,0x72,0xa2,0xf9,0x1d,0x73,0x28,0x6f,0xfd,0x68,0x3b,0x9b,0xb2,0xcf,0x4f,0x1b
+   };
+   /* msg1+pub1+sig4 test vector is from wycheproof/ecdsa_test INVALID (long form encoding of length) */
+   const unsigned char sig4[] = {
+      0x30,0x81,0x45,0x02,0x20,0x2b,0xa3,0xa8,0xbe,0x6b,0x94,0xd5,0xec,0x80,0xa6,0xd9,0xd1,0x19,
+      0x0a,0x43,0x6e,0xff,0xe5,0x0d,0x85,0xa1,0xee,0xe8,0x59,0xb8,0xcc,0x6a,0xf9,0xbd,0x5c,0x2e,
+      0x18,0x02,0x21,0x00,0xb3,0x29,0xf4,0x79,0xa2,0xbb,0xd0,0xa5,0xc3,0x84,0xee,0x14,0x93,0xb1,
+      0xf5,0x18,0x6a,0x87,0x13,0x9c,0xac,0x5d,0xf4,0x08,0x7c,0x13,0x4b,0x49,0x15,0x68,0x47,0xdb
+   };
+   /* msg1+pub1+sig5 test vector is from wycheproof/ecdsa_test INVALID (length contains leading 0) */
```
sjaeckel, line 403: I don't like the fact that wycheproof decides against the standard: X.690 states explicitly in 8.1.3.5 - Note 2: "In the long form, it is a sender's option whether to use more length octets than the minimum necessary."

_But_ IMO it's true that Postel's law shouldn't be applied in cryptography.

Btw. this comment also applies to the test above `msg1+pub1+sig4 test vector is from wycheproof/ecdsa_test INVALID (long form encoding of length)` where AFAIU X.690 is ambiguous on whether this form is valid or not

sjaeckel:
I tried to look into these failing cases, but haven't been able to find any documentation on what these test cases exactly do. Does anyone else have a hint?

levitte:
Gods, this has been ages, hasn't it?  I can try to have a look in the coming days...

sjaeckel:
> Gods, this has been ages, hasn't it?

Absolutely!

> I can try to have a look in the coming days...

Absolutely 2 :)

tests/ecc_test.c
levitte: 8.1.3.5 is about BER...  DER has extra conditions to be deterministic, which should mean that notes like Note 2 are not valid for DER.  I actually expected this to be remarked upon in 10.1 (Length forms), but no.  Maybe that was an omission that no one noticed.

Still, to have DER deterministic, it would be logical that the minimum necessary length octets be a requirement for DER.

levitte:
> I tried to look into these failing cases, but haven't been able to find any documentation on what these test cases exactly do. Does anyone else have a hint?

I had a look...  and first of all, when I ran the tests of this PR, they all succeed, so not sure what's supposedly failing.

As for the INVALID test vectors that are added in this PR, they seem to be correct (i.e. they are, in fact, invalid, and the test does detect that they fail as they should).

So er, what was the question?

BTW, for anyone that needs a link to wycheproof data, here it is: https://github.com/C2SP/wycheproof/tree/master/testvectors
This PR only adds a select few test vectors, not the whole lot.  The wycheproof/ecdsa_test reference in the comments is this file: https://github.com/C2SP/wycheproof/blob/master/testvectors/ecdsa_test.json

levitte:
As far as I'm concerned, this PR can be merged as is.

For the future, it might be a good thing to consider generating a separate C source file with just data, using the wycheproof JSON files as input.  That's gonna be a biiiig source file, though.

levitte (approved):

sjaeckel:
@karel-m please correct me if I'm wrong

FTR I've just force-pushed again, because I accidentally wiped my last changes by pushing an old version from my second machine ...

> I had a look... and first of all, when I ran the tests of this PR, they all succeed, so not sure what's supposedly failing.
> 
> As for the INVALID test vectors that are added in this PR, they seem to be correct (i.e. they are, in fact, invalid, and the test does detect that they fail as they should).
> 
> So er, what was the question?

If I run the tests they pass, that's true, but I get this:

```
$ make CFLAGS="-DLTM_DESC -DUSE_LTM -I../libtommath" EXTRALIBS="../libtommath/libtommath.a" test -j65
[...]
$ ./test ecc
[...]
ecc_test............XXX-TODO should be valid - wycheproof / Edge case for Shamir multiplication
XXX-TODO should be valid - wycheproof / extreme value for k and edgecase s
XXX-TODO should be valid - wycheproof / extreme value for k
XXX-TODO should be valid - wycheproof / extreme value for k and s^-1
passed   8090.779ms
[...]
```

So these tests should succeed, but we don't accept those signatures because something's wrong in `ltc_ecc_mul2add()`.

> For the future, it might be a good thing to consider generating a separate C source file with just data, using the wycheproof JSON files as input. That's gonna be a biiiig source file, though.

IIRC perl-CryptX already does all those tests. I'm not sure whether it'd really make sense to add all those tests here, or whether it'd be better to contribute to perl-CryptX and use it as kind of "test harness".

levitte:
Argh!  I build with CMake, and I just discovered that I have to set `MPI_PROVIDER` right to get the tests that use MPI to run at all.  When I did `ctest --test-dir _build --verbose` (I did without `--verbose` before), I got to see that `ecc_test` was a "nop".

So here I go again, with better build setup.  Let's see what I get to see this time around

levitte:
I built against GMP, and I only get these:

```
1: ecc_test............XXX-TODO should be valid - wycheproof / extreme value for k and edgecase s
1: XXX-TODO should be valid - wycheproof / extreme value for k
1: passed   7298.617ms
```

So it could be that at least *some* of the issues are bound to the math library.  I'll dive into that this evening.

sjaeckel:
FYI if you want to you can also build against all 3 MPI's [^0] and run the tests.

For me it looks as follows:
```sh
$ cmake -DWITH_LTM=On -DWITH_TFM=On -DWITH_GMP=On -DBUILD_TESTING=On ..
[...]
$ make -j$(($(nproc)*2+1))
[...]
$ for mpi in ltm gmp tfm; do ./tests/test-ltc ecc $mpi | tail -n7; done
XXX-TODO should be valid - wycheproof / Edge case for Shamir multiplication
XXX-TODO should be valid - wycheproof / extreme value for k and edgecase s
XXX-TODO should be valid - wycheproof / extreme value for k
XXX-TODO should be valid - wycheproof / extreme value for k and s^-1
MP_PROVIDER  = LibTomMath
MP_DIGIT_BIT = 60
sizeof(ltc_mp_digit) = 8

ecc_test............passed   3171.687ms

SUCCESS: passed=1 failed=0 nop=0 duration=3.2sec real=3.2sec
XXX-TODO should be valid - wycheproof / extreme value for k and edgecase s
XXX-TODO should be valid - wycheproof / extreme value for k
MP_PROVIDER  = GNU MP
MP_DIGIT_BIT = 64
sizeof(ltc_mp_digit) = 8

ecc_test............passed   2633.805ms

SUCCESS: passed=1 failed=0 nop=0 duration=2.6sec real=2.6sec
Non-fatal 'no-operation' requested. (2)
/path/to/libtomcrypt/tests/ecc_test.c:1792:s_ecc_import_export()
XXX-TODO should be valid - wycheproof / Edge case for Shamir multiplication
XXX-TODO should be valid - wycheproof / extreme value for k and edgecase s
XXX-TODO should be valid - wycheproof / extreme value for k
XXX-TODO should be valid - wycheproof / extreme value for k and s^-1
Non-fatal 'no-operation' requested. (2)
/path/to/libtomcrypt/tests/ecc_test.c:1800:s_ecc_test_recovery()
MP_PROVIDER  = TomsFastMath
MP_DIGIT_BIT = 64
sizeof(ltc_mp_digit) = 8

ecc_test............passed   1795.282ms

SUCCESS: passed=1 failed=0 nop=0 duration=1.8sec real=1.8sec
```

So this indeed looks like the first and last failing tests are caused by something inside ltm (CC @czurnieden) resp. tfm, and the two other failing tests are universal ...

[^0]: both ltm and tfm provide CMake projects which are exported to the CMake registry, so you can simply clone the repos and build them in CMake, and from then on the ltc CMake project will find those builds automagically.

levitte:
I did a deep dive for the "wycheproof / extreme value for k and edgecase s" case, and found that something happens in `ltc_ecc_mul2add` such that at this point, all elements in `C` are zero:

https://github.com/libtom/libtomcrypt/blob/2e9f2b5e44dd498dba60e5175c6c7effa5ff3728/src/pk/ecc/ltc_ecc_mul2add.c#L169

This leads back to this area in `ecc_verify_hash_ex()`:

https://github.com/libtom/libtomcrypt/blob/2e9f2b5e44dd498dba60e5175c6c7effa5ff3728/src/pk/ecc/ecc_verify_hash.c#L181-L192

Since `mG->x` is zero following the call to `ltc_mp.ecc_mul2add()`, `v` also becomes zero after the call of `mp_mod()`, and of course, that won't compare with `r`, which is not zero.

I'm stopping here for the moment, 'cause trying to figure out what goes wrong in `ltc_ecc_mul2add` is more than I have time for today.

sjaeckel:
We're down to

```
$ for mpi in ltm gmp tfm; do ./tests/test-ltc ecc $mpi | tail -n7; echo ----------; done
XXX-TODO should be valid - wycheproof / Edge case for Shamir multiplication
XXX-TODO should be valid - wycheproof / extreme value for k and s^-1
MP_PROVIDER  = LibTomMath
MP_DIGIT_BIT = 60
sizeof(ltc_mp_digit) = 8

ecc_test............passed   4758.523ms

SUCCESS: passed=1 failed=0 nop=0 duration=4.8sec real=4.8sec
----------
MP_PROVIDER  = GNU MP
MP_DIGIT_BIT = 64
sizeof(ltc_mp_digit) = 8

ecc_test............passed   3624.249ms

SUCCESS: passed=1 failed=0 nop=0 duration=3.6sec real=3.6sec
----------
XXX-TODO should be valid - wycheproof / Edge case for Shamir multiplication
XXX-TODO should be valid - wycheproof / extreme value for k and s^-1
Non-fatal 'no-operation' requested. (2)
/home/jaeckel/devel/open_source/libtomcrypt/tests/ecc_test.c:2254:s_ecc_import_export()
Non-fatal 'no-operation' requested. (2)
/home/jaeckel/devel/open_source/libtomcrypt/tests/ecc_test.c:2261:s_ecc_test_recovery()
MP_PROVIDER  = TomsFastMath
MP_DIGIT_BIT = 64
sizeof(ltc_mp_digit) = 8

ecc_test............passed   1929.827ms

SUCCESS: passed=1 failed=0 nop=0 duration=1.9sec real=1.9sec
----------

```

sjaeckel:
OK, I'm so much out of my comfort zone here ... @czurnieden maybe you can help?

The only thing I can say is: if I apply the following patch, which replaces the montgomery reduction of ltm by the regular modulo operation, the Wycheproof tests are OK (and the execution time skyrockets).

```
diff --git a/src/math/ltm_desc.c b/src/math/ltm_desc.c
index 99c9f14b..65c72929 100644
--- a/src/math/ltm_desc.c
+++ b/src/math/ltm_desc.c
@@ -403,10 +403,8 @@ static int montgomery_setup(const void *a, void **b)
    if (*b == NULL) {
       return CRYPT_MEM;
    }
-   if ((err = mpi_to_ltc_error(mp_montgomery_setup(a, *b))) != CRYPT_OK) {
-      XFREE(*b);
-   }
-   return err;
+   **(mp_digit**)b = 1;
+   return CRYPT_OK;
 }
 
 /* get normalization value */
@@ -414,7 +412,7 @@ static int montgomery_normalization(void *a, const void *b)
 {
    LTC_ARGCHK(a != NULL);
    LTC_ARGCHK(b != NULL);
-   return mpi_to_ltc_error(mp_montgomery_calc_normalization(a, b));
+   return set_int(a, 1);
 }
 
 /* reduce */
@@ -423,7 +421,7 @@ static int montgomery_reduce(void *a, const void *b, void *c)
    LTC_ARGCHK(a != NULL);
    LTC_ARGCHK(b != NULL);
    LTC_ARGCHK(c != NULL);
-   return mpi_to_ltc_error(mp_montgomery_reduce(a, b, *((mp_digit *)c)));
+   return mpi_to_ltc_error(mp_mod(a, b, a));
 }
 
 /* clean up */
```

```
ecc_test............passed   3965.769ms
vs.
ecc_test............passed  17124.081ms
```

At one point, while looping in `ltc_ecc_mul2add()`, a value of 0 is calculated for `C->z`, c.f. below for details. This then results in `ltc_ecc_is_point_at_infinity()` calculating `(C->x)^3` resp. `(C->y)^2`. There lies the problem, since the two should be equal for those two failing Wycheproof cases, leading to setting the point to infinity, but they are different when using the montgomery reduction.

**Debug details**

gdb breakpoint conditional: `(x == 31 && nibble == 3) || x > 31` on the `if (first == 0) {` statement inside the `for (x = 0;; ) {` loop.
1. The following call to `ltc_mp.ecc_ptadd()` calculates `C->z` as 0.
2. The loop continues and the next call to `ltc_mp.ecc_ptdbl()` should set `C` to infinity.

sjaeckel:
I just pushed the [debug-wycheproof-failures](https://github.com/libtom/libtomcrypt/tree/debug-wycheproof-failures) branch, in order to simplify this if someone else wants to have a look.

karel-m:
My investigation revealed that the culprit is Shamir's trick/optimisation.

However, removing this trick and switching to the unoptimised implementation (see my last commits) means a significant performance penalty.

before patch:
```
ecc_test............passed   8393.784ms
```

after patch:
```
ecc_test............passed  10803.603ms
```

29% slow down

not sure whether this patch is good for merging

UPDATE: maybe the problem is not the optimisation itself, but a faulty ltc_ecc_mul2add implementation

karel-m:
there is a bug in ltc_ecc_projective_add_point (I've just commited a failing test)

karel-m:
looks like the similar improvements/changes now need to be added for TFM

karel-m:
now squashed all my mess into one commit

karel-m:
@sjaeckel this PR is IMO ready for review/merge

sjaeckel (approved):

PR #447, opened 2018-10-07
Make LTC_ECCSIG_RFC7518 strict (again) 

karel-m:
This is basically a revert of my PR - Less strict ecc_verify_hash_ex (as it was before ecc_recover_key) #443

It enforces the strict verification mode for `LTC_ECCSIG_RFC7518` but allow relaxed variant via `LTC_ECCSIG_RFC7518_RELAXED`.

I have changed my mind after investigation the latest wycheproof test suit. it seems that since approx. a yar ago wycheproof guys (=google) did change the status of some "acceptable" cases to "invalid". In other words what is declared today as "acceptable" is a candidate for tomorrow's "invalid".

So my new proposal is either enforce the strict length validation (the size of the signature should always be twice the number of bytes of the size of the order) or this PR.

Cc: @sjaeckel @rmw42

### DIFF
--- a/tests/ecc_test.c
+++ b/tests/ecc_test.c
@@ -240,11 +240,12 @@ done:
 }
 
 /* https://github.com/libtom/libtomcrypt/issues/443 */
-static int _ecc_issue443(void)
+/* https://github.com/libtom/libtomcrypt/issues/447 */
+static int _ecc_issue443_447(void)
 {
    const ltc_ecc_curve* cu;
    ecc_key key;
-   int stat = 0;
+   int err, stat = 0;
    unsigned char hash[64];
    unsigned long hashlen;
    const unsigned char msg[] = { 0x54,0x65,0x73,0x74 };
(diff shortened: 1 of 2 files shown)


src/headers/tomcrypt_pk.h
```diff
@@ -248,10 +248,12 @@ typedef struct {
 typedef enum ecc_signature_type_ {
    /* ASN.1 encoded, ANSI X9.62 */
    LTC_ECCSIG_ANSIX962   = 0x0,
-   /* raw R, S values */
+   /* raw R, S values - both values strictly padded to the size of the curve order*/
    LTC_ECCSIG_RFC7518    = 0x1,
+   /* raw R, S values - without strict padding check (used just for verification) */
+   LTC_ECCSIG_RFC7518_RELAXED = 0x2,
```
sjaeckel: if we really change it, would it probably make more sense to make this enum (at least partially) a bitfield and relaxed is a flag (e.g. `0x80`) where there's currently only a relaxed implementation of RFC 7518?

tests/ecc_test.c
```diff
@@ -153,52 +153,35 @@ static int _ecc_issue108(void)
...
-   const unsigned char pub1[] = {
+   const unsigned char pub[] = {
       0x04,
       0x4a,0x03,0xef,0x9f,0x92,0xeb,0x26,0x8c,0xaf,0xa6,0x01,0x07,0x24,0x89,0xa5,0x63,
       0x80,0xfa,0x0d,0xc4,0x31,0x71,0xd7,0x71,0x28,0x13,0xb3,0xa1,0x9a,0x1e,0xb5,0xe5,
       0x3e,0x21,0x3e,0x28,0xa6,0x08,0xce,0x9a,0x2f,0x4a,0x17,0xfd,0x83,0x0c,0x66,0x54,
       0x01,0x8a,0x79,0xb3,0xe0,0x26,0x3d,0x91,0xa8,0xba,0x90,0x62,0x2d,0xf6,0xf2,0xf0
    };
-   const unsigned char sig1[] = { 0x05, 0x01 };
-   /* msg+pub2+sig2 test vector is from wycheproof  - ecdsa_webcrypto_test (incorrect size of signature) */
```
sjaeckel: shouldn't this testcase stay in there for the relaxed api where it was okay before?

sjaeckel (commented):
My personal preference would be to not have relaxed API's for security critical stuff!

karel-m:
changed to the strict approach

sjaeckel (approved):

PR #450, opened 2018-10-14
Fix #449 (ECDSA forcing DER/ASN.1)

### LINKED ISSUE #449: ecc_sign_hash_ex + ecc_verify_hash_ex forces linking DER/ASN.1 code
this is an issue related to the discussion in #438 

> The current all-in-one ecc_sign_hash_ex and ecc_verify_hash_ex forces anyone using ECDSA to link the whole ASN.1 stuff to their binary even if they are using just RFC 7518 or RFC 5656 (ssh2) signature format. 

One possible approach is to split `ecc_verify_hash_ex.c` into:

* `ecc_verify_hash_internal.c` (only in `tomcrypt_private.h`)
* `ecc_verify_hash.c` (this is the troublemaker depending on ASN.1 stuff)
* `ecc_sign_hash_rfc7518.c`
* `ecc_sign_hash_rfc5656.c`
* `ecc_sign_hash_eth27.c`

Or another option:
* `ecc_verify_hash_internal.c` (only in `tomcrypt_private.h`)
* `ecc_verify_hash.c` (this is the troublemaker depending on ASN.1 stuff)
* `ecc_verify_hash_ex.c` (rfc7518 + rfc5656 + eth27)

karel-m:
Fix ECDSA forcing DER.

TODO:
- [x] update doc
- [x] add a test case with eth27 signature format
- [x] refactor `ecc_recover_key()`

This fixes #449

### DIFF
--- a/doc/crypt.tex
+++ b/doc/crypt.tex
@@ -1,4 +1,5 @@
 \documentclass[synpaper]{book}
+\usepackage[T1]{fontenc}
 \usepackage{geometry}
 \usepackage{hyperref}
 \usepackage{makeidx}
@@ -44,6 +45,7 @@
 \def\C{{\mathbb C}}
 \def\Q{{\mathbb Q}}
 \definecolor{DGray}{gray}{0.5}
+\newcommand{\code}[1]{\texttt{\textit{#1}}}
 \newcommand{\emailaddr}[1]{\mbox{$<${#1}$>$}}
 \def\twiddle{\raisebox{0.3ex}{\mbox{\tiny $\sim$}}}
 \def\gap{\vspace{0.5ex}}
--- a/doc/crypt.tex
+++ b/doc/crypt.tex
@@ -5953,18 +5953,18 @@ int ecc_recover_key(const unsigned char *sig,
                                 ecc_key *key);
 \end{verbatim}
 
-This function will recover (a) public key from the ECDSA signature in the array pointed to by \textit{sig} of length \textit{siglen} octets, the message digest
-pointed to by the array \textit{hash} of length \textit{hashlen}, and the recovery id \textit{recid}. It will store the recovered
-key into \textit{key} and return CRYPT\_OK if recovery succeeds, or an error if recovery fails.
+This function will recover (a) public key from the \textit{ECDSA} signature in the array pointed to by \code{sig} of length \code{siglen} octets, the message digest
+pointed to by the array \code{hash} of length \code{hashlen}, and the recovery id \code{recid}. It will store the recovered
+key into \code{key} and return \code{CRYPT\_OK} if recovery succeeds, or an error if recovery fails.
 This is for compatibility with the (v,r,s) signatures used in Ethereum, where public keys are not explicitly shared,
-only the parity of the public key. For curves like secp256k1, recid will take values of 0 or 1, corresponding to the
-parity of the public key's y coordinate. For curves like secp112r2, with a cofactor of 4, values 0..7 are possible,
+only the parity of the public key. For curves like \textit{secp256k1}, \code{recid} will take values of 0 or 1, corresponding to the
+parity of the public key's y coordinate. For curves like \textit{secp112r2}, with a cofactor of 4, values 0..7 are possible,
 with the low bit corresponding to the parity and the higher bits specifying the public key's x coordinate's multiple
 of the curve's order.
-If the signature format contains the recovery id (currently only \textit{LTC\_ECCSIG\_ETH27}), \textit{recid} can be -1
+If the signature format contains the recovery id (currently only \code{LTC\_ECCSIG\_ETH27}), \code{recid} can be -1
 which signals that the recovery id from the signature blob should be used. This means an application does not need to
 extract the recovery id from such a signature in order to use this function.
-The function \textit{ecc\_recover\_key} implements multiple signature formats, and the output is compliant for GF(p) curves.
+The function \code{ecc\_recover\_key()} implements multiple signature formats, and the output is compliant for GF(p) curves.
 
 \subsection{Signature Formats}
 The following signature formats are suported:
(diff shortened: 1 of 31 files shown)


src/pk/ecc/ecc_sign_hash_rfc7518.c
```diff
@@ -0,0 +1,83 @@
...
+  @param prng      An active PRNG state
+  @param wprng     The index of the PRNG you wish to use
+  @param key       A private ECC key
+  @return CRYPT_OK if successful
+*/
+int ecc_sign_hash_rfc7518(const unsigned char *in,  unsigned long inlen,
+                          unsigned char *out, unsigned long *outlen,
+                          prng_state *prng, int wprng, const ecc_key *key)
+{
+   return ecc_sign_hash_rfc7518_ex(in, inlen, out, outlen, prng, wprng, NULL, key);
```
sjaeckel, line 70: why not make `ecc_sign_hash_rfc7518_ex()` the new `ecc_sign_hash_rfc7518`?
we could simply tell the user to pass NULL if he doesn't want the recid... I'd prefer that as the difference is now really not that huge.

sjaeckel (commented):
LGTM

src/pk/ecc/ecc_sign_hash_rfc7518.c
karel-m: The `ecc_sign_hash_rfc7518` was already released or not?

sjaeckel: let me check

sjaeckel: true, damn, okay let's keep it like that!

sjaeckel: OTOH will the next release break API anyways... so what!? :)

karel-m: I do not know.

Just a note that there is also #477 which may also need some API changes.

sjaeckel: ok, let's keep it for now as is

sjaeckel:
so we wait now until the two missing points are done?

karel-m:
If you agree with the proposed split, which I assume you do, I (or any other volunteer) have to update doc and tests.

sjaeckel:
:+1:

levitte (commented):
This looks good so far.  Do you need help with the last bits (docs / test)?

sjaeckel:
> Do you need help with the last bits (docs / test)?

Sure! Thanks :)

levitte:
> > Do you need help with the last bits (docs / test)?
> 
> Sure! Thanks :)

Sorry that I never came back on this...  [sigh] ambition doesn't always align with time

doc/crypt.tex
```diff
@@ -5809,91 +5809,97 @@ \subsection{Signature Generation}
...
-int ecc_sign_hash(const unsigned char *in,
-                        unsigned long  inlen,
-                        unsigned char *out,
-                        unsigned long *outlen,
-                           prng_state *prng,
-                                  int  wprng,
-                              ecc_key *key);
+int ecc_sign_hash(const unsigned char *in,  unsigned long inlen,
+                  unsigned char *out, unsigned long *outlen,
+                  prng_state *prng, int wprng, const ecc_key *key);
```
levitte: I dunno, I actually found the column of parameters easier to read

doc/crypt.tex
```diff
@@ -5809,91 +5809,97 @@ \subsection{Signature Generation}
...
 \end{verbatim}
 
-This function will ECDSA sign the message digest stored in the array pointed to by \textit{in} of length \textit{inlen} octets.  The signature
-will be stored in the array pointed to by \textit{out} of length \textit{outlen} octets.  The function requires a properly seeded PRNG, and
-the ECC \textit{key} provided must be a private key.
+This function will ECDSA sign the message digest stored in the array pointed to by \texttt{in} of length \texttt{inlen} octets.  The signature
+will be stored in the array pointed to by \texttt{out} of length \texttt{outlen} octets.  The function requires a properly seeded PRNG, and
+the ECC \texttt{key} provided must be a private key.
```
levitte: I think that the *italics* style was meant to align with [man-pages(7)](https://www.man7.org/linux/man-pages/man7/man-pages.7.html), see "Formatting conventions for manual pages describing functions".

...  but can't really say why `monospace` wouldn't be just as good.  If we decide on the latter, should there be a future PR with s/textit/texttt/ applied (of course just for variable names)?

sjaeckel:
Thanks for the review!

> I actually found the column of parameters easier to read

Yeah, I was lazy and c&p'ed from the sources, so it's the same style.

I'll have another look.

> the italics style was meant to align with man-pages(7)

Aahh alright, that makes sense.

I believe that `monospace` is better suited for that purpose to highlight code/variable names (and that's also what is used in most modern formatting styles). Additionally this allows to use _italics_ to reference to proper nouns/standards/etc.

> should there be a future PR with s/textit/texttt/ applied

Absolutely - for all "code-related" terms.

sjaeckel:
> the italics style was meant to align with man-pages(7)

Aahh alright, that makes sense.

I believe that `monospace` is better suited for that purpose to highlight code/variable names (and that's also what is used in most modern formatting styles). Additionally this allows to use _italics_ to reference to proper nouns/standards/etc.

> should there be a future PR with s/textit/texttt/ applied

Absolutely - for all "code-related" terms.

levitte:
So, just for kicks, I tried a change from \\textit to \\texttt, and well, I dunno...  typographically, it's a question what's more distinct in a text flow.  Compare these two images:

Monotone:
monotone

-----

Italics:
italic

sjaeckel:
I liked the idea, but I found it hard to decide on two completely different parts of the text, so I took the same text and applied the two different styles.

### `Monospace` vs. _Italics_ vs. _`Monospace-Italics`_
monospace
------------
italics
------------
monospace-italic


... and I agree that the text flow is different in the `monospace` version. It somehow feels kind of "better readable" with _italics_, but don't ask me why, I have the impression the meaning is clearer in the `monospace` version.

[Edit] and _`Monospace-Italics`_ kind of combines both. It's a bit strange at first but it's very well readable and stands out from the regular text.

What should we do? :shrug:

I'm ~slightly~ strongly leaning towards ~`monospace`~ _`monospace-italics`_ ...

[Edit 2] FTR: I created the third with the following addition on the top

```
diff --git a/doc/crypt.tex b/doc/crypt.tex
index 81995116..4b4274c8 100644
--- a/doc/crypt.tex
+++ b/doc/crypt.tex
@@ -1,2 +1,3 @@
 \documentclass[synpaper]{book}
+\usepackage[T1]{fontenc}
 \usepackage{geometry}
@@ -46,2 +47,3 @@
 \definecolor{DGray}{gray}{0.5}
+\newcommand{\code}[1]{\texttt{\textit{#1}}}
 \newcommand{\emailaddr}[1]{\mbox{$<${#1}$>$}}
```

The first change I'd keep for sure, since otherwise one cant copy&paste from the PDF.

The second change declares the `\code{}` macro to format as  _`monospace-italics`_.

levitte:
I think we can agree on *`monospace-italics`*.

The `\code{}` macro should be added either way, so whatever we end up choosing (or changing again in the future, 'cause one never knows), it makes for consistency, makes life easier, so just add it 😉

levitte (approved):
Did a visual check, looks great :-)

Issue #307, opened 2017-10-05
Fix gcm_memory() API


sjaeckel:
The pointers for in- and output of `gcm_memory()` are  `pt` and `ct` whereas in the entire library the pattern `in` and `out` is used.
This should be fixed in `gcm_memory()`.

sjaeckel:
I'm thinking of keeping this for now and deprecating the API in the 2.0 cycle and introduce a generic concept for the AEAD class, similar to what we already have.
Any opinions?

PR #451, opened 2018-10-14
Wycheproof failing GCM test - invalid/modified tag

karel-m:
The wycheproof GCM test in question:
```
{
  "tcId" : 16,
  "comment" : "Flipped bit 0 in tag",
  "key" : "000102030405060708090a0b0c0d0e0f",
  "iv" : "505152535455565758595a5b",
  "aad" : "",
  "msg" : "202122232425262728292a2b2c2d2e2f",
  "ct" : "eb156d081ed6b6b55f4612f021d87b39",
  "tag" : "d9847dbc326a06e988c77ad3863e6083",
  "result" : "invalid",
  "flags" : []
},
```
The trouble is that we do not reject invalid tag `d9847dbc326a06e988c77ad3863e6083`, corresponding valid tag has the first byte `d8`.


karel-m:
Exactly the same trouble with ChaCha20Poly1305 - failing test (one of them):
```
        {
          "tcId" : 62,
          "comment" : "Flipped bit 0 in tag expected tag:a3e3fdf9fba6861b5ad2607f40b7f447",
          "key" : "00112233445566778899aabbccddeeff00112233445566778899aabbccddeeff",
          "iv" : "000102030405060708090a0b",
          "aad" : "616164",
          "msg" : "",
          "ct" : "",
          "tag" : "a2e3fdf9fba6861b5ad2607f40b7f447",
          "result" : "invalid",
          "flags" : []
        },
```

karel-m:
@sjaeckel a question for you (pls answer without looking into doc and/or code)

Let us have the following code:
```C
      err = gcm_memory(idx, key, sizeof(key), iv, sizeof(iv), NULL, 0,
                       pt, sizeof(ct), ct, tag, &taglen, GCM_DECRYPT);
```
Do you expect that the caller should fill the tag + taglen before calling `gcm_memory` (which will handle the tag validation and return CRYPT_OK only if the tag is correct) OR do you expect that `gcm_memory` will only calculate the tag and store it into tag + taglen (so that the tag validation is caller's job)?

sjaeckel:
ah the good ol' `gcm_memory()` API

TBH I've no clue, I'd have to have a look in the code or docs... Btw. there's still #307 ;)

sjaeckel:
did I already mention that this is super inconsistent with e.g. `ccm_memory()` :)

(or the other AEAD algos which don't have a `xx_memory()` function...)

karel-m:
Yes, it is a mess. And BTW I have a security hole in my perl bindings https://github.com/DCIT/perl-CryptX/issues/47 as I was expecting that it validates the tag (which it doesn't).

What is the proper fix?

I tend to like more two separate functions: `gcm_decrypt_verify_memory` + `gcm_encrypt_authenticate_memory` as we have with EAX and completely DEPRECATE `gcm_memory`

karel-m:
BTW the doc for `gcm_memory` does not help at all. 

A possible "fix" may be just updating doc and declare the current behaviour, IMO slightly unexpected, as a feature. Or can we afford an API breakage here?

sjaeckel:
> Yes, it is a mess. And BTW I have a security hole in my perl bindings [DCIT/perl-CryptX#47](https://github.com/DCIT/perl-CryptX/issues/47) as I as expecting that it validates the tag (which it doesn't).

ooops

> What is the proper fix?

very good question!

we did put some effort into cleaning-up the API of `ccm_memory()` in #73 / #76 

we could re-use that pattern?

Otherwise I'd be fine with two separate functions, but should we then change that in CCM as well?!

karel-m:
I have the code aligning `gcm_memory` and `chacha20poly1305_memory` to the `ccm_memory` behaviour (== adding tag validation to decrypt mode) nearly ready. I'll finish it & push it and after that we can decide what next.

sjaeckel (approved):

PR #452, opened 2018-10-14
Wycheproof failing CCM test - invalid tag len

karel-m:
The wycheproof CCM test in question:

```
{
  "ivSize" : 96,
  "keySize" : 128,
  "tagSize" : 40,
  "type" : "AeadTest",
  "tests" : [
    {
      "tcId" : 480,
      "comment" : "Invalid tag size",
      "key" : "000102030405060708090a0b0c0d0e0f",
      "iv" : "",
      "aad" : "",
      "msg" : "",
      "ct" : "d3dab1ee494cc229099d6cac7df14add",
      "tag" : "231a2d8f6a",
      "result" : "invalid",
      "flags" : []
    }
  ]
},
```
The trouble is that taglen should be from `( 4, 6, 8, 10, 12, 14, 16 )` - if I understand correctly https://nvlpubs.nist.gov/nistpubs/Legacy/SP/nistspecialpublication800-38c.pdf

Which means that any attempt to decrypt+validate using a tag with invalid taglen should be rejected


karel-m:
Thanks for additional polishing. Can we merge this one?

sjaeckel (approved):

PR #459, opened 2018-10-26
Avoid LTC_ARGCHK in void functions

karel-m:
By default LTC_ARGCHK never returns (it aborts). 

However, with `ARGTYPE == 4` it calls `return CRYPT_INVALID_ARG` which causes compiler warnings.


sjaeckel:
that's why there's also `LTC_ARGCHKVD()` :)

karel-m:
Fixed

sjaeckel:
Approved

PR #483, opened 2019-05-25
Improve/x509

sjaeckel:
This factors-out the core part of `rsa_import_x509()` so it can be reused later, e.g. #364


karel-m:
Can it be reused in already existing `ecc_import_x509` as well?

sjaeckel:
> Can it be reused in already existing `ecc_import_x509` as well?

fine like that? :)

sjaeckel:
ping @karel-m

karel-m (approved):

PR #497, opened 2019-09-23
Bcrypt

sjaeckel:
~Add the bcrypt password-based hashing algorithm.~

Add the bcrypt PBKDF algorithm as used when encrypting OpenSSH key files.

### Checklist

### DIFF
--- a/src/headers/tomcrypt_misc.h
+++ b/src/headers/tomcrypt_misc.h
@@ -60,10 +60,10 @@ int base16_decode(const          char *in,  unsigned long  inlen,
 #endif
 
 #ifdef LTC_BCRYPT
-int bcrypt_pbkdf_openbsd(const          char *password, unsigned long password_len,
-                         const unsigned char *salt,     unsigned long salt_len,
-                               unsigned int  rounds,              int hash_idx,
-                               unsigned char *out,      unsigned long *outlen);
+int bcrypt_pbkdf_openbsd(const          void *secret, unsigned long secret_len,
+                         const unsigned char *salt,   unsigned long salt_len,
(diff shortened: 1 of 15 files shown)


src/headers/tomcrypt_misc.h
```diff
@@ -59,6 +59,13 @@ int base16_decode(const          char *in,  unsigned long  inlen,
                         unsigned char *out, unsigned long *outlen);
 #endif
 
+#ifdef LTC_BCRYPT
+int bcrypt(const          char *password, unsigned long password_len,
```
karel-m: Providing the fact that we have test cases like `"pass\0word"` I would rather use `const unsigned char *password` or even `const unsigned char *secret`

sjaeckel: IIUC "pass\0word" is no valid password for most password-based crypto applications (e.g. SCRAM) but it's no problem for us to handle it, so how about changing it to `const void*`?

Another point I realized in the last days... would it probably make sense to rename this function to `bcrypt_openssh()` as it's a specific version which differs from the original brcypt proposal?

karel-m: The `const void*` is definitely better than `const char*` and AFAIK we already use it like this somewhere (pkcs8?).

Did you mean `bcrypt_openssh` or `bcrypt_openbsd`?

sjaeckel: > `bcrypt_openssh` or `bcrypt_openbsd`

I don't have an opinion :-) maybe `bcrypt_openbsd` is better as it was first used there.

PR #499, opened 2019-09-25
Minor improvements

sjaeckel:
* add more rotate intrinsics
* speed up the build of PR's
* minor changes of makefiles

### DIFF
--- a/.ci/build.sh
+++ b/.ci/build.sh
@@ -23,6 +23,22 @@ if [[ $mret -ne 0 ]] || [[ $cnt -gt 1 ]]; then
    exit 1
 fi
 
+# remove the standard arguments from the make options
+opts=${3//makefile.shared/}
+opts=${opts//makefile/}
+opts=${opts//V=1/}
+opts=${opts//COVERAGE=1/}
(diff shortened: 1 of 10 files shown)


karel-m (approved):
In general I am fine with the changes. Any idea why `coverage/coveralls` failed?

sjaeckel:
> Any idea why `coverage/coveralls` failed?

because my patch wasn't complete... :-) 

I cherry-picked some more commits from #497 as they didn't belong there in the first place and that we can see if the difference in coverage is only by the change of the gcc version used!

sjaeckel:
I would say that the results of travis build [1845.1](https://travis-ci.org/libtom/libtomcrypt/jobs/591409342) and its [coveralls results](https://coveralls.io/jobs/53830686) which were still on Trusty/gcc 4.8.5 vs. build [1846.1](https://travis-ci.org/libtom/libtomcrypt/jobs/591417395) and its [coveralls results](https://coveralls.io/jobs/53831353) which are now on Bionic/gcc 7.4.0 indicate that the drop of coverage by 0.4% is caused by the new version of gcc, or am I mistaken?

Issue #175, opened 2017-03-23
 Polish API `unsigned long` vs. `size_t` for sizes/lengths


karel-m:
As mentioned in #174 and #173 we should consider switching from `unsigned long` to `size_t` when passing the length or size of something.

@rofl0r said:
> also: in C the proper type for sizes is size_t, not unsigned long. (usually it's the same, but some odd platforms like windows have a 32bit long on 64bit).

and @rofl0r a bit later:
> using size_t does not necessarily mean including stdint.h. since you're a fan of ifdef hacks over configure tests you could hardcode a list of CPP defines which define HAVE_STDINT_H and only use it there, for the other platforms fall back to typedefing it to unsigned long, or unsigned long long on windows64 bit.

rofl0r:
i thought about this some more and i came to the conclusion that in order to keep future problems to a mininum it's probably not a good idea to make a manual typedef to size_t on platforms that currently do not provide stdint.h, because future version may actually have stdint.h and pull in the header implicitly through inclusion of another header, so the typedef will be problematical. so if you want to go that route it's preferable to have a custom type, for example ltc_size, which is typedef'd to unsigned long or size_t, depending on the platform.

karel-m:
To my surprise even pretty old MS compilers are able to handle `size_t`.

sjaeckel:
TBH I'm not aware of any platform that I used lately which doesn't have `stdint.h` so I'm against introducing something like `ltc_size` or even some "macro-own-typedef-magic"^TM. I'd rather make it up to the integrator of the library to provide a `stdint.h` himself with the appropriate typedefs if there's none available ootb.

rofl0r:
btw according to `man 0p stddef.h`, stddef.h is the proper header to include to get this type.

sjaeckel:
> stddef.h is the proper header to include to get this type

mea culpa, you're right! no more typing for me today

PR #500, opened 2019-09-26
Fix SSH API

sjaeckel:
While playing around with the SSH API I realized that as it is now it is incomplete resp. broken.

When reading the RFC4251 Ch. 5 it says

```
   string

      Arbitrary length binary string.  Strings are allowed to contain
      arbitrary binary data, including null characters and 8-bit
      characters.  ...
```
which is clearly not possible as it is implemented now.

I think there are two ways this could be handled in the order of preference

1. fix the `LTC_SSHDATA_STRING` as proposed in the API description of the two c files
2. add a new `LTC_SSHDATA_BUFFER` as proposed in the header file

The disadvantage of 1. would be that it would break existing applications, but I think we can live with that for the sake of a better API.

Your thoughts?

CC @rmw42 @mkj

### DIFF
--- a/doc/crypt.tex
+++ b/doc/crypt.tex
@@ -7233,6 +7233,71 @@ int hkdf(           int  hash_idx,
 
 Parameters are as in \textit{hkdf\_extract()} and \textit{hkdf\_expand()}.
 
+
+\mysection{SSH}
+
+The library provides functions to encode and decode SSH data as specified in RFC4251 Ch. 5.
+
+\subsection{Data types}
+
+The following enum is used to indicate a specific SSH data type
+(besides EOL which is an internal one that indicates the end of a sequence).
+
+\begin{figure}[h]
+\begin{center}
+\begin{small}
+\begin{tabular}{|l|l|l|}
+\hline \textbf{Definition}    & \textbf{arg data Type}      & \textbf{SSH Type} \\
+\hline LTC\_SSHDATA\_EOL      & -                      & End of SSH data sequence. \\
+\hline LTC\_SSHDATA\_BYTE     & \texttt{unsigned char} & \texttt{byte} type \\
+\hline LTC\_SSHDATA\_BOOLEAN  & \texttt{unsigned char} & \texttt{boolean} type \\
+\hline LTC\_SSHDATA\_UINT32   & \texttt{ulong32}       & \texttt{uint32} \\
+\hline LTC\_SSHDATA\_UINT64   & \texttt{ulong64}       & \texttt{uint64} \\
+\hline LTC\_SSHDATA\_STRING   & \texttt{char*}         & \texttt{string} (one octet per char) \\
+\hline LTC\_SSHDATA\_MPINT    & \texttt{mp\_int}       & \texttt{mpint} \\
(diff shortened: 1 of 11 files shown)


karel-m:
I am for `fixing the LTC_SSHDATA`

sjaeckel:
> I am for `fixing the LTC_SSHDATA`

currently writing the documentation on it, should be ready by tonight

sjaeckel:
Short question, while using the API I thought it'd be useful if the signature of the decoding had a pointer of the length as argument, so we could return the processed length. What do you think?

```
-int ssh_decode_sequence_multi(const unsigned char *in, unsigned long inlen, ...)
+int ssh_decode_sequence_multi(const unsigned char *in, unsigned long *inlen, ...)
```

Yes, this would differ from the way its inspiration (DER multi) works, but it already differs by the order by which the arguments get passed...

karel-m:
Following our DER multi pattern is not a good idea, couple of times I was missing the information about the length of actually decoded data.

It is IMO necessary to somehow return how much input data was "consumed" during decoding - either by `unsigned long *inlen` or via an extra argument.

sjaeckel:
This would be the behavior I thought about... okay?

```
-  @param inlen  Length of buffer to decode
+  @param inlen  [in/out] Length of buffer to decode/Remaining bytes after decoding the given sequence
```

karel-m:
Well, I would slightly prefer reverse logic: _"Actual length of in buffer used for decoding the given sequence"_ but yours: _"Remaining bytes after decoding the given sequence"_ will work as well.

sjaeckel:
> Well, I would slightly prefer reverse logic: _"Actual length of in buffer used for decoding the given sequence"_ but yours: _"Remaining bytes after decoding the given sequence"_ will work as well.

I thought from a usability context if we'd reverse the logic one has to do

```
decode(p, &len, multi combination 1);
tot_len -= len;
len = tot_len;
if (some recently decoded thing)
  decode(p, &len, multi variation 1);
else
  decode(p, &len, multi variation 2);
tot_len -= len;
len = tot_len;
decode(p, &len, multi combination 2);
```
whereas with my proposal one could do

```
decode(p, &len, multi combination 1);
if (some recently decoded thing)
  decode(p, &len, multi variation 1);
else
  decode(p, &len, multi variation 2);
decode(p, &len, multi combination 2);
```

karel-m:
We already have:
```c
/**
   ASN.1 DER Flexi(ble) decoder will decode arbitrary DER packets and create a linked list of the decoded elements.
   @param in      The input buffer
   @param inlen   [in/out] The length of the input buffer and on output the amount of decoded data
   @param out     [out] A pointer to the linked list
   @return CRYPT_OK on success.
*/
int der_decode_sequence_flexi(const unsigned char *in, unsigned long *inlen, ltc_asn1_list **out)
```

sjaeckel:
I went with the old way, because I realized that to do what I wanted, it'd have required too many (paradigm) changes.

sjaeckel:
Does someone have a reason why we shouldn't use `size_t` where appropriate _now_? I'd change that then in this PR.

sjaeckel:
> Does someone have a reason why we shouldn't use `size_t` where appropriate _now_? I'd change that then in this PR.

okay, one step back, let's just stay with `unsigned long` for now and change to `size_t` all at once

karel-m:
ad `size_t` we already have #175

PR #504, opened 2019-10-02
Minor fixes and improvements

sjaeckel:
Title says all

e.g. update to latest ltm API

@jbech-linaro do you probably have the possibility to check/time to review that the "check ranges of more ciphers" commit does indeed what you did in #493 for AES?


src/ciphers/rc5.c
```diff
@@ -136,6 +136,10 @@ int rc5_ecb_encrypt(const unsigned char *pt, unsigned char *ct, const symmetric_
    LTC_ARGCHK(pt   != NULL);
    LTC_ARGCHK(ct   != NULL);
 
+   if (skey->rc5.rounds < 12 || skey->rc5.rounds > 24) {
```
jockebech, line 139: I'm not too familiar with the inner workings of RC5, but the lower limit seems fine, however for the upper limit I'm not sure what the correct value would be (max value according to spec is 255 IIRC). But it's probably unlikely that one would use a greater value than 24. Related to this, `rfc2040` seems to define `MAX_ROUNDS` to 20.

jockebech (commented):
For commit b96e137b8f644f6642322f0c63c829a6f0a6e4cf 
`Reviewed-by: Joakim Bech <joakim.bech@linaro.org>`

src/ciphers/rc5.c
sjaeckel: thx!

> Related to this, `rfc2040` seems to define `MAX_ROUNDS` to 20

that value is implementation defined as is 24 for ltc.

sjaeckel:
I just asked myself if
1. it'd make sense to review the default rounds of all the ciphers? as e.g. as of [1] our default value of 12 rounds for RC5 can be attacked
2. we can even change those values? as this will break users who rely on 0 being that then-old default number of rounds...

[1] https://en.wikipedia.org/w/index.php?title=RC5&oldid=917621181#cite_note-Biryukov-1

PR #509, opened 2019-10-08
fix and cleanup des and 3des test cases

armcc:
- Drop unused 'num' field from struct des_test_case.
 - Fix the order of arguments passed to compare_testvector() for the
   "DES" tests (actual -vs- expected buffers were swapped, leading to
   misleading error messages for failing tests).
 - Move the 3des test vectors which were previously included in
   des_test() into des3_test(). Note that these 3des test vectors
   were previously guarded by LTC_TEST_EXT, so even though they
   always failed when run as des tests, the failure was not seen by
   default.
 - Minor indent fixes.

Signed-off-by: Andre McCurdy <armccurdy@gmail.com>

### Checklist


sjaeckel:
Thanks for the PR!

> * Drop unused 'num' field from struct des_test_case.
> * Fix the order of arguments passed to compare_testvector() for the
>   "DES" tests (actual -vs- expected buffers were swapped, leading to
>   misleading error messages for failing tests).
> * Minor indent fixes.

Those 3 points are fine IMO

> * Move the 3des test vectors which were previously included in
>   des_test() into des3_test(). Note that these 3des test vectors
>   were previously guarded by LTC_TEST_EXT, so even though they
>   always failed when run as des tests, the failure was not seen by
>   default.

This part is misleading as they were never explicitly 3DES TV's but just "more" DES TV's and now you're (ab)using the fact that the same key used thrice results in the same result as a single DES round...

I think this part should be reverted. @karel-m your thoughts?

sjaeckel:
btw if you're already at it, feel free to re-factor the TV's in `des_test()` and de-duplicate them :)

armcc:
Thanks for the feedback! I started looking at the extra test vectors because they failed when I enabled them. Googling the test vector values found a golang test suite that uses them for 3DES, so I just copied that approach. However, I see now that the real bug in the current LTC code is that these vectors are marked as DES decryption tests when they should be DES encryption tests... just fixing that makes the tests pass. I'll send an updated PR.

src/ciphers/des.c
```diff
@@ -1963,28 +1929,31 @@ int des_test(void)
...
-        if (compare_testvector(cases[i].out, sizeof(tmp), tmp, sizeof(tmp), "DES", i) != 0) {
-           return CRYPT_FAIL_TESTVECTOR;
+        des_ecb_decrypt(ct, tmp, &des);
+
+        if (compare_testvector(tmp, sizeof(tmp), cases[i].txt, 8, "DES Decrypt", i) != 0) {
+            return CRYPT_FAIL_TESTVECTOR;
         }
 
-        /* now see if we can encrypt all zero bytes 1000 times, decrypt and come back where we started */
+        /* now see if we can encrypt all zero bytes 100 times, decrypt and come back where we started */
```
sjaeckel: Wouldn't it be enough if we do this with one key with 1000 rounds?

Now I also realize why I excluded those additional testvectors, I only remember that it took forever to run when I enabled them, but that makes sense now :-)

sjaeckel (commented):
Looks good!

src/ciphers/des.c
armcc: Each iteration only encrypts and decrypts 8 bytes, so there's not a huge amount of actual work being done even with 1000 iterations for every test vector. Running the 1000 iterations test outside the test vector loop does seem cleaner and more logical though, so probably worth doing for that reason.

sjaeckel (approved):

sjaeckel:
Thanks! Looks very good now and thx for the TDES TV's!

Now you only need to fix this travis error and we're good to go :-) 

https://travis-ci.org/libtom/libtomcrypt/jobs/595904038

btw. you can install the pre-commit hooks via `make install_hooks` which will then run the same basic tests which failed in said build job.

armcc:
The indent issue should now be fixed but it looks like the Travis build then failed for some other reason? Should I force push the latest version again to try to re-trigger Travis?

sjaeckel:
> The indent issue should now be fixed

yes, that's fixed

> it looks like the Travis build then failed for some other reason

yeah, there's an issue with their infrastructure, c.f. [1] resp. [2]

> Should I force push the latest version again to try to re-trigger Travis?

no need, I've just restarted the failed builds

[1] https://travis-ci.community/t/sometimes-build-fails-when-apt-is-updating-postgresql-apt-repository/4872
[2] https://github.com/travis-ci/travis-build/pull/1771

PR #517, opened 2019-10-25
Visualstudio solution

FredeJ:
Migrate the solution file to Visual Studio 15.

The goal of this is to allow libtomcrypt to be made available through vcpkg (https://github.com/microsoft/vcpkg/)


karel-m:
👎  While you are able to import the old format (VS2008) into newer Visual Studio (e.g. VS2015) it is not possible the other way around. I am for keeping VS2008 compatibility.

sjaeckel:
@FredeJ is it an option to use the `makefile.msvc` for your purpose?

karel-m:
No, makefile.msvc is not what I want.

The other thing is that it is not sufficient to just save the original `.vcproj` as `.vcxproj`. Somebody has also update `helper.pl` so that the `.vcxproj` file is updated when a new source file is added, deleted or renamed.

sjaeckel:
> No, makefile.msvc is not what I want.

I asked @FredeJ ;-)

> The other thing is that it is not sufficient to just save the original `.vcproj` as `.vcxproj`. Somebody has also update `helper.pl` so that the `.vcxproj` file is updated when a new source file is added, deleted or renamed.

Right, I thought that was done, but apparently it wasn't.

karel-m:
I will support this PR only when it keeps support for both cases:
* `libtomcrypt_VS2008.sln` + `libtomcrypt_VS2008.vcproj`
* `libtomcrypt_VS2015.sln` + `libtomcrypt_VS2015.vcxproj` (not sure about having .vcxproj.filters in git)

But again, somebody has to fix `helper.pl` so that it is able to update .vcxproj.

FredeJ:
@sjaeckel `makefile.msvc` seems to be able to get me basically 95% there. I was not aware of `nmake` at all but was just using `msbuild`, which failed with the older project and solution files.

Now all I'm missing is figuring out how to build both debug and release, and both libtomcrypt and libtommath should be buildable with vcpkg :)

sjaeckel:
> @sjaeckel `makefile.msvc` seems to be able to get me basically 95% there

cool :)

> Now all I'm missing is figuring out how to build both debug and release, and both libtomcrypt and libtommath should be buildable with vcpkg :)

uhm, pass in some CFLAGS and change the libname for the debug build? something like `nmake -f makefile.msvc CFLAGS="-MTd" LIBMAIN_S="tomcrypt-debug.lib"`

I don't run windows, so I can't tell if it's exactly like that, but that should give you a direction

TheWerewolf:
Two issues.

First, it's bad practice to refuse to update to newer versions of VS as until Microsoft started making full community editions, you had to BUY a copy of VS and those versions are EOL. VS2019 and later have Community Editions that are complete releases and were intended exactly for open source and hobyy use. It also suggests the code is compiler dependent which is also a bad practice.0

Second, I just tried to compile it using the 2008 SLN and it can't compile because "tommath.h" is missing, so maybe check builds on that version at the very least? Tried chabnging it to "tomcrypt_math.h" which can be found, but doesn't resolve the 72 errors generated during compile.

When (and if) I can get this to work with VS2022, I'll post a pull request (and yes, I'll leave the VS2008 sln untouched).

enginelesscc:
> 👎 While you are able to import the old format (VS2008) into newer Visual Studio (e.g. VS2015) it is not possible the other way around. I am for keeping VS2008 compatibility.

Fyi, as of **at least** VS2022 (possibly earlier), upgrading does no longer work and its impossible to use these relict-vs solution files.

Upgrading these solution files to at least VS2010 will fix that.

PR #520, opened 2019-11-07
Timing benchmark

sjaeckel:
This modifies how the timing demo of RSA and ECC work.

1. it updates the timing output to be easier machine-readable.
1. it introduces reference-keys to measure the timing.


karel-m (changes requested):
Could you please move changes related to `src/math/ltm_desc.c` into a separate PR?

The rest is IMO fine for merging.

karel-m (approved):

PR #524, opened 2019-11-19
Make all modes use the ECB mode API

sjaeckel:
This PR implements usage of the ECB API for all other modes, MAC and AEAD algorithms that use block ciphers.

This will make all above mentioned algorithms use an eventually set ECB accelerator function.
Also it cleans up duplicated parts like cipher ID and block-length.


levitte:
Nice cleanup.  This PR could use some love, though...

PR #535, opened 2020-05-07
Add "memory" to clobber list of `STORE/LOAD32`

### LINKED ISSUE #532: [x86 inline assembly] Macros read or write in memory without it being stated
### Description
Macros STORE32H and LOAD32H are dereferencing a pointer given in argument, but without the keyword ```"memory"```, the compiler assumes they do not: missing dataflow information may lead to undue optimization.
Note that the -64H  version already list "memory" in their clobbers.

### Patches

The simplest one would be to add ```"memory"``` in the clobber list but I found more elegant to transform the constraint from register ```'r'``` to memory cell ```'m'```.
Here is the patch for 32H version:

```diff
diff --git a/libtomcrypt/src/headers/tomcrypt_macros.h b/libtomcrypt/src/headers/tomcrypt_macros.h
index ae37450..e80d46e 100644
--- a/libtomcrypt/src/headers/tomcrypt_macros.h
+++ b/libtomcrypt/src/headers/tomcrypt_macros.h
@@ -10,6 +10,8 @@
 /* ---- HELPER MACROS ---- */
 #ifdef ENDIAN_NEUTRAL
 
+#include <stdint.h>
+
 #define STORE32L(x, y)                                                                     \
   do { (y)[3] = (unsigned char)(((x)>>24)&255); (y)[2] = (unsigned char)(((x)>>16)&255);   \
        (y)[1] = (unsigned char)(((x)>>8)&255); (y)[0] = (unsigned char)((x)&255); } while(0)
@@ -70,17 +72,19 @@ do { XMEMCPY (&(x), (y), 4);                    \
 #elif !defined(LTC_NO_BSWAP) && (defined(INTEL_CC) || (defined(__GNUC__) && (defined(__DJGPP__) || defined(__CYGWIN__) || defined(__MINGW32__) || defined(__i386__) || defined(__x86_64__))))
 
 #define STORE32H(x, y)           \
-asm __volatile__ (               \
+asm  (                           \
    "bswapl %1     \n\t"          \
    "movl   %1, %0 \n\t"          \
    "bswapl %1     \n\t"          \
-   : "m"(*(y)) : "r"(x));
+   : "=m"(*((uint32_t*)(y)))     \
+   : "r"(x));
 
 #define LOAD32H(x, y)          \
-asm __volatile__ (             \
+asm (                          \
    "movl %1, %0\n\t"           \
    "bswapl %0\n\t"             \
-   :"=r"(x): "r"(*(y)));
+   : "=r"(x)                   \
+   : "m"(*((uint32_t*)(y))));
 
 #else
 

```

Note (1): ```volatile``` keyword is no longer required since the compiler knows what is the output.
Note (2): we are using the standard integer header for the fixed width ```uint32_t``` type. It looks like the ```ulong32``` may be wider that 32bit.

Note (3): in the STORE macro, the second bswapl restore the value of the read only entry ```x```. It is correct, but it is also possible to inform the compiler that the entry has changed by linking it to a dummy output such that the compiler will save its value only if it need it again. No solution is perfect:
- the actual one may perform useless ```bswapl``` instructions
- the following one may force the compiler to save it when reused

```diff
diff --git a/libtomcrypt/src/headers/tomcrypt_macros.h b/libtomcrypt/src/headers/tomcrypt_macros.h
index 94e368f..099675a 100644
--- a/libtomcrypt/src/headers/tomcrypt_macros.h
+++ b/libtomcrypt/src/headers/tomcrypt_macros.h
@@ -10,6 +10,8 @@
 /* ---- HELPER MACROS ---- */
 #ifdef ENDIAN_NEUTRAL
 
+#include <stdint.h>
+
 #define STORE32L(x, y)                                                                     \
   do { (y)[3] = (unsigned char)(((x)>>24)&255); (y)[2] = (unsigned char)(((x)>>16)&255);   \
        (y)[1] = (unsigned char)(((x)>>8)&255); (y)[0] = (unsigned char)((x)&255); } while(0)
@@ -69,18 +71,21 @@ do { XMEMCPY (&(x), (y), 4);                    \
 
 #elif !defined(LTC_NO_BSWAP) && (defined(INTEL_CC) || (defined(__GNUC__) && (defined(__DJGPP__) || defined(__CYGWIN__) || defined(__MINGW32__) || defined(__i386__) || defined(__x86_64__))))
 
-#define STORE32H(x, y)           \
-asm __volatile__ (               \
-   "bswapl %0     \n\t"          \
-   "movl   %0,(%1)\n\t"          \
-   "bswapl %0     \n\t"          \
-      ::"r"(x), "r"(y));
+#define STORE32H(x, y) do {         \
+   uint32_t dummy;                  \
+   asm ("bswapl %0 \n\t"            \
+        "movl %0, %1 \n\t"          \
+        : "=r" (dummy),             \
+          "=m" (*((uint32_t *)(y))) \
+        : "0" (x));                 \
+   } while (0);
 
 #define LOAD32H(x, y)          \
-asm __volatile__ (             \
-   "movl (%1),%0\n\t"          \
+asm (                          \
+   "movl %1, %0\n\t"           \
    "bswapl %0\n\t"             \
-   :"=r"(x): "r"(y));
+   : "=r"(x)                   \
+   : "m"(*((uint32_t*)(y))));
 
 #else
 

```

### Steps to Reproduce
No bug was observed yet, but it cost nothing to fix it once for all.
Actually, we demonstrate "in lab" problems with these categories of bugs in case of inlining (especially with LTO enabler) or code refactoring.

### Additional Information

We are academic researchers working in automated program analysis.
We are currently interested in checking compliance of inline asm chunks as found in C programs.
We would be very interested to hear your opinion on these matters.
Are you interested in such errors and patches?
Also, besides the patches, we are currently working on a code analyzer prototype designed to check asm compliance and to propose patches when the chunk is not compliant. This is still work in progress and we are finalizing it.
The error and patch I reported to you came from my prototype.
In case such a prototype would be made available, would you consider using it?

Regards,
Frédéric Recoules

sjaeckel:
c.f. cefff85550786ec869b39c0cb4a5904e88c84319

This fixes #532

### DIFF
--- a/src/headers/tomcrypt_macros.h
+++ b/src/headers/tomcrypt_macros.h
@@ -74,13 +74,13 @@ asm __volatile__ (               \
    "bswapl %0     \n\t"          \
    "movl   %0,(%1)\n\t"          \
    "bswapl %0     \n\t"          \
-      ::"r"(x), "r"(y));
+      ::"r"(x), "r"(y): "memory");
 
 #define LOAD32H(x, y)          \
 asm __volatile__ (             \
    "movl (%1),%0\n\t"          \
    "bswapl %0\n\t"             \
-   :"=r"(x): "r"(y));
+   :"=r"(x): "r"(y): "memory");
 
 #else
 


sjaeckel:
@recoules fine like that?

recoules:
Yes, it solves the issue.

PR #537, opened 2020-05-29
Cast STORE64H/STORE64L argument to ulong64

mkj:
Avoids undefined behaviour with right shift greater than 32 bits. (c99 6.5.7 "If the value of the right operand is negative or is *greater than or equal to the width* of the promoted left operand, the behavior is undefined."

I haven't reproduced it here but it was hit using STORE64H outside of libtomcrypt on BCM4706 MIPS32r2 with GCC 4.2.4  https://github.com/mkj/dropbear/pull/99

The patch doesn't seem to affect code generation with -DLTC_NO_ASM on x64.

### DIFF
--- a/src/headers/tomcrypt_macros.h
+++ b/src/headers/tomcrypt_macros.h
@@ -21,10 +21,10 @@
            ((ulong32)((y)[0] & 255)); } while(0)
 
 #define STORE64L(x, y)                                                                     \
-  do { (y)[7] = (unsigned char)(((x)>>56)&255); (y)[6] = (unsigned char)(((x)>>48)&255);   \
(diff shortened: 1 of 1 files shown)


sjaeckel:
IMO this cast belongs in the calling code and not in the STORE macro.

Those macro's clearly say that they store a 64bit Variable and if you pass a 32bit in there you have to cast it.

@karel-m your opinion?

karel-m:
I do not have strong opinion on this. But I slightly lean towards leaving the cast to the calling code as by casting in these macros we may hide some potential issues on the calling side (which now throws at least some warning).

sjaeckel:
Closing this, since @karel-m and me basically have the same opinion on this.

Please re-open if you have further questions/comments.

PR #538, opened 2020-06-10
Avoid zero length array with LTC_BASE64_URL unset

mkj:
If LTC_BASE64_URL is unset it creates a zero length array, which isn't allowed by some compilers

### DIFF
--- a/src/misc/base64/base64_decode.c
+++ b/src/misc/base64/base64_decode.c
@@ -48,8 +48,8 @@ static const unsigned char map_base64[256] = {
 255, 255, 255, 255 };
 #endif /* LTC_BASE64 */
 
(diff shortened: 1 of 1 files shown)


sjaeckel:
I've implemented a different version of this patch. Can you please have a look whether this would also be acceptable for you?

https://github.com/libtom/libtomcrypt/tree/fix-b64-zero-length-array

karel-m:
I think @sjaeckel 's patch https://github.com/libtom/libtomcrypt/commit/d20fdcc76d6b78059af4b847b428bd34972883ce is slightly better

karel-m:
@sjaeckel I suggest merging this PR do develop.

PR #539, opened 2020-06-17
Relicense

sjaeckel:
After so many complaints about the evil wording of the WTFPL and the non-applicability of the PD in most jurisdictions and a personal request by future contributors, we decided to change the license of LibTomCrypt to "The Unlicense" (as already done for libtommath).

Before this license change we made even sure that The Unlicense gets approved by the OSI, which is official since yesterday [1] ... just in case your legal department asks ;-)

[1] http://lists.opensource.org/pipermail/license-review_lists.opensource.org/2020-June/004890.html


karel-m (approved):
👍

buggywhip:
+1

Congratulations!!!

Sent fm iTouch

>

czurnieden (approved):
Oh, that went faster than expected!

dlegaultbbry:
Any hope for a new version release after this happens so we can get the new goodies officially? (25519 stuff for example)

sjaeckel:
@dlegaultbbry let's see if I can find the time

PR #554, opened 2020-11-17
Add Xcode support

cntrump:

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "tommath"]
+	path = tommath
+	url = https://github.com/libtom/libtommath.git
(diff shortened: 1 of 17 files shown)


sjaeckel (changes requested):
Thanks for this PR.

FYI: I've canceled the travis build because this won't be accepted as is.

1. the submodule of libtommath must go. it's okay to have it as "default dependency" as for the msvc projects, but not like that.
2. the project files you committed will become out-of-date as soon as a new file is added or removed. Either integrate it in helper.pl as done for the msvc project files or it isn't acceptable.

cntrump:
I am trying add `Cocoapods` support instead

sjaeckel:
Doesn't conan.io already have ltc support?

Issue #551, opened 2020-10-27
AES implementation is not side-channel resistant


ereshetova:
I guess everyone is aware that the current AES implementation is not resistant to traditional non-speculative side channels due to the direct usage of T-tables, which can reveal the information about the key being used (usage of T-tables basically creates a data access pattern that depends on the key). The key question is what would be the right approach for libtomcrypt to address this. 

I see a number of options (each has its pros and cons):

1. **Usage of dedicated CPU-specific instructions**. Instructions like AES-NI [1] and their alternatives can be used to solve the above problem in a pretty straightforward way. Many open source libs have went this way, i.e. openssl, bearssl, etc. 
**Pros**: Easy to implement, the end implementation is fast and efficient, applicable to all modes
**Cons**: Not a generic solution, CPU-specific. 

2. **Bitslicing**. This is a technique [2] that allows to achieve good side-channel resistance for the modes that can be parallelized. 
It is implemented in some crypto libraries, for example in bearSSL [3].
**Pros**: no CPU dependency, relatively efficient implementation 
**Cons**: Not useful for non-parallelizable modes.

3.  **Algebraic implementation.** This approach would involve computing the result of S-box on the fly using algebraic structure of S-box, i.e. the fact that an Sbox is a combination of finding a byte-wise inverse in GF(2^8) + bitwise affine transformation. The computation itself should be very carefully implemented not to contain any side-channels as well. I am not aware of any major crypto library that follows this approach. 
**Pros**: no CPU dependency, applicable for all modes
**Cons**: Likely to be slow and not efficient (albeit can try to use some speedup tricks given that we always work in the same GF)

[1] https://en.wikipedia.org/wiki/AES_instruction_set 
[2] https://link.springer.com/chapter/10.1007/11935070_14
[3] https://bearssl.org/constanttime.html#bitslicing

What are people's opinions on this for libtomcrypt? Which way is seen as a preferred direction?

sjaeckel:
IMO it'd make sense to have at least a bitsliced implementation in addition to the current one.

To go for CPU-specific instructions would be nice as well but something that should be provided additional to bitslicing... AES-NI for x86 or AES instructions for aarch64 could be a starter.

I don't think an algebraic implementation makes sense... I've played with htable [1] some years ago and it was really slow ... :) Maybe it could be made faster but additional to its slowness it was also big IIRC.

[1] https://github.com/coron/htable

stevensam:
[...]

Issue #480, opened 2019-04-10
Hardware acceleration


fuchsia36:
Would it be within the scope of this project to include support for architecture-specific intrinsics?

For instance, all Intel CPUs after Westmere (2010) and all AMD CPUs after Jaguar (2013) support the AES-NI instruction set, all ARM CPUs after ARMv8-A (2013) support hardware acceleration of AES, SHA-1, and SHA-256; and modern Intel (post-Goldmont, 2016) and AMD (post-Ryzen, 2017) CPUs support hardware acceleration of SHA-1 and SHA-256, in addition to even faster AES acceleration.

Other intrinsics (SSE, AVX) could also be used to speed up other algorithms.

The downside is that there is more code to maintain, although it becomes very easy to compare it to known-good implementations.

sjaeckel:
> Would it be within the scope of this project to include support for architecture-specific intrinsics?

The concept of cipher and hash descriptors lets you easily replace the standard implementation by a customized one w/o touching the standard implementation.

So definitely yes, I would happily accept PR's regarding this topic. We should discuss beforehand e.g. how this should be integrated and enabled...

> The downside is that there is more code to maintain...

...especially in that context.

> Other intrinsics (SSE, AVX) could also be used to speed up other algorithms.

There's gcm which has SSE2 improvements and the last time I looked at the asm, gcc was generating the exact same, so I'm not sure if that's really worth the effort... sometimes yes, sometimes no... :-)

fuchsia36:
You could also use it for random number generation.  
At least for fortuna anyway, where you can put what RDRAND generates as one source, so even if it's heavily backdoored it still won't hurt.

stevensam:
Just for reference, there is a related open issue below, to which I've added some reference links regarding adding AES-NI. This issue thread might have been the more appropriate one to do so on however:

https://github.com/libtom/libtomcrypt/issues/551

PR #557, opened 2020-12-21
Add AES-NI

sjaeckel:

This PR adds support for the AES-NI instructions as discussed in e.g. #480 or #551

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -38,6 +38,7 @@ jobs:
           - { BUILDNAME: 'VALGRIND',             BUILDOPTIONS: '',                                                    BUILDSCRIPT: '.ci/valgrind.sh' }
           - { BUILDNAME: 'STOCK',                BUILDOPTIONS: '',                                                    BUILDSCRIPT: '.ci/run.sh' }
           - { BUILDNAME: 'STOCK-MPI',            BUILDOPTIONS: '-ULTM_DESC -UTFM_DESC -UUSE_LTM -UUSE_TFM',           BUILDSCRIPT: '.ci/run.sh' }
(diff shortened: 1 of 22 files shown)


karel-m:
Basically fine. 

I am just thinking about having:
```
aes_autodetect_desc
```
on top of:

```
aesni_desc
aes_desc
```

I have to think it over.

karel-m:
Or even:
```
aesni_desc      (new AES-NI)
aessw_desc      (old SW implementation aes_desc renamed to aessw_desc)
aes_desc        (aes with runtime autodetection of aessw/aesni variant)
```

With this all calls like `find_cipher("aes")` (we have it at many places) will choose autodetected aessw/aesni variant.

sjaeckel:
> ```
> aesni_desc      (new AES-NI)
> aessw_desc      (old SW implementation aes_desc renamed to aessw_desc)
> aes_desc        (aes with runtime autodetection of aessw/aesni variant)
> ```
> 
> With this all calls like `find_cipher("aes")` (we have it at many places) will choose autodetected aessw/aesni variant.

I like the idea but what's the purpose of the `aesni_desc`? It should still check whether the CPU has support as otherwise bad things will happen.

My proposal would therefor be to make aesni the new default:

```
aes_desc        (aes with runtime autodetection of aessw/aesni variant)
aessw_desc      (old SW implementation aes_desc renamed to aessw_desc)
```

What do you think?

I'm having an early WIP about adding support for the AES instructions on ARMv8, there we could then basically use the same pattern and add a `aes_arm.c` which provides `aes_desc` and also does autodetection!?

sjaeckel:
Hah, having a second look I understand your reasoning ... I have to think about it a bit more!

isarrider:
@sjaeckel will there be an update?

PR #558, opened 2020-12-24
Add new utf8 test-vector, update comments explaining utf8 decoding

jamuir:
**Description:**
Minor changes to help test and clarify the way utf8 strings are
decoded.  This originated from my misunderstanding of the fix for
issue #507.  The new test-vector uses two bytes to encode each
wide-char.

The utf8 format is described here:

https://tools.ietf.org/html/rfc3629#section-3

**Testing:**
```
  $ make clean
  $ make CFLAGS="-DUSE_LTM -DLTM_DESC -I../libtommath" EXTRALIBS="../libtommath/libtommath.a" test
  $ ./test
```
You can confirm that the new utf8 test data is correct using python:
```
  >>> s="\xD7\xA9\xD7\x9C\xD7\x95\xD7\x9D"
  >>> s.decode("utf-8")
  u'\u05e9\u05dc\u05d5\u05dd'
```

### Checklist

### DIFF
--- a/src/pk/asn1/der/utf8/der_decode_utf8_string.c
+++ b/src/pk/asn1/der/utf8/der_decode_utf8_string.c
@@ -11,11 +11,11 @@
 #ifdef LTC_DER
 
 /**
-  Store a UTF8 STRING
+  Decode a UTF8 STRING and recover an array of unicode characters.
   @param in      The DER encoded UTF8 STRING
   @param inlen   The size of the DER UTF8 STRING
-  @param out     [out] The array of utf8s stored (one per char)
-  @param outlen  [in/out] The number of utf8s stored
+  @param out     [out] The array of unicode characters (wchar_t*)
+  @param outlen  [in/out] The number of unicode characters in the array
   @return CRYPT_OK if successful
 */
 int der_decode_utf8_string(const unsigned char *in,  unsigned long inlen,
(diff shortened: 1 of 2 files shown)


src/pk/asn1/der/utf8/der_decode_utf8_string.c
```diff
@@ -53,21 +53,23 @@ int der_decode_utf8_string(const unsigned char *in,  unsigned long inlen,
 
    /* proceed to decode */
    for (y = 0; x < inlen; ) {
-      /* get first byte */
+      /* read first byte */
       tmp = in[x++];
 
-      /* count number of bytes */
+      /* count number of left-shifts needed to get zero in most-sig bit */
```
sjaeckel: IMO this comment makes it now even less clear, but this utf8 thing is also less than trivial... not to speak of well written documentation of source code ... ;)

What you described now is what is technically done, not what the purpose is.

The purpose of the for-loop is to count the number of most-significant bits which are set to 1.
This has to be done as the number of msb-1s tells us the number of bytes that are used to encode the utf8 symbol.

```
0 -> 1 byte
{2,3,4} -> {2,3,4} bytes
```

src/pk/asn1/der/utf8/der_decode_utf8_string.c
```diff
@@ -53,21 +53,23 @@ int der_decode_utf8_string(const unsigned char *in,  unsigned long inlen,
...
       for (z = 0; (tmp & 0x80) && (z <= 4); z++, tmp = (tmp << 1) & 0xFF);
 
+      /* for valid utf8, z is in {0,2,3,4}.
+         if z>0, then z-1 equals the number of additional bytes to read */
       if (z == 1 || z > 4 || (x + (z - 1) > inlen)) {
          return CRYPT_INVALID_PACKET;
       }
 
-      /* decode, grab upper bits */
+      /* right-shift tmp to restore least-sig bits */
```
sjaeckel, line 84: :+1:

src/pk/asn1/der/utf8/der_decode_utf8_string.c
```diff
@@ -53,21 +53,23 @@ int der_decode_utf8_string(const unsigned char *in,  unsigned long inlen,
...
-      /* get first byte */
+      /* read first byte */
       tmp = in[x++];
 
-      /* count number of bytes */
+      /* count number of left-shifts needed to get zero in most-sig bit */
       for (z = 0; (tmp & 0x80) && (z <= 4); z++, tmp = (tmp << 1) & 0xFF);
 
+      /* for valid utf8, z is in {0,2,3,4}.
+         if z>0, then z-1 equals the number of additional bytes to read */
```
sjaeckel: I'd move that second line below to the `--z` operation.

sjaeckel (changes requested):
Thx for the PR!

src/pk/asn1/der/utf8/der_decode_utf8_string.c
jamuir: I added the second comment line to make the length check `x+(z-1) > inlen` more comprehensible.  If I move that comment below, then there is no longer a hint to explain `x+(z-1) > inlen`.

Maybe the thing to do here is move `x+(z-1) > inlen` into a separate if-statement.

src/pk/asn1/der/utf8/der_decode_utf8_string.c
jamuir: I didn't want to try and explain too much in this first comment.  I just wanted to explain the for-loop.  You're right that we are counting the number of ms-bits equal to one, but the way we count that is by left-shifting until the ms-bit is zero.

Note that there is a comment below that explains how `z` gives the number of additional bytes to read.

I can add an additional comment above the line `tmp = in[x++]` explaining the utf-8 decoding strategy, if that would help.

jamuir:
@sjaeckel : please take another look and let me know what you think.

sjaeckel (approved):
Perfect, thanks!

PR #560, opened 2021-02-10
Cleanup and fixes

sjaeckel:
Some housekeeping and fixes

### DIFF
--- a/tests/mpi_test.c
+++ b/tests/mpi_test.c
@@ -6,16 +6,16 @@
 static int s_radix_to_bin_test(void)
 {
    /* RADIX 16 */
-   const char *ghex = "2";
-   const char *phex = "FFFFFFFFFFFFFFFFC90FDAA22168C234C4C6628B80DC1CD129024E088A67CC74020BBEA63B139B22"
+   const char ghex[] = "2";
+   const char phex[] = "FFFFFFFFFFFFFFFFC90FDAA22168C234C4C6628B80DC1CD129024E088A67CC74020BBEA63B139B22"
                       "514A08798E3404DDEF9519B3CD3A431B302B0A6DF25F14374FE1356D6D51C245E485B576625E7EC6"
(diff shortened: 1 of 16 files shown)


tests/mpi_test.c
```diff
@@ -6,51 +6,51 @@
 static int s_radix_to_bin_test(void)
 {
    /* RADIX 16 */
-   const char *ghex = "2";
-   const char *phex = "FFFFFFFFFFFFFFFFC90FDAA22168C234C4C6628B80DC1CD129024E088A67CC74020BBEA63B139B22"
+   const char ghex[] = "2";
+   const char phex[] = "FFFFFFFFFFFFFFFFC90FDAA22168C234C4C6628B80DC1CD129024E088A67CC74020BBEA63B139B22"
```
jamuir, line 10: nice. I also prefer `buf[]` rather than `buf*` since then you can use `sizeof(buf)-1` for the buffer length.

sjaeckel: you meant the string length? :)

btw. one can also use `strlen()` for const strings, this is usually evaluated at compile time and not causing a call to the function

karel-m:
The branch `cleanup-and-fixes` fails to build with cygwin + gcc-10.2.0
```
  * cc src/ciphers/camellia.o
In file included from ./src/headers/tomcrypt.h:78,
                 from ./src/headers/tomcrypt_private.h:4,
                 from src/ciphers/aes/aes.c:25:
./src/headers/tomcrypt_cfg.h:10: error: unterminated #ifndef
   10 | #ifndef TOMCRYPT_CFG_H
      |
```

sjaeckel:
> The branch `cleanup-and-fixes` fails to build with cygwin + gcc-10.2.0

what the ... 5bfa8a3 apparently uncovered that `tomcrypt_cfg.h` was "broken" for how long exactly? :D 

... not really broken, but the include guards didn't do what they promised ...

karel-m (approved):

sjaeckel:
@mkj do you maybe want to have a look as well?

PR #569, opened 2021-05-26
rsa: add rsa key generate with bignumber public exponent

cneveux:
Function rsa_make_key() limits the RSA key generates to a public
exponent of type long (32 bits or 64 bits).
RSA standard specify that public exponent e can be between 65537 (included)
and 2^256 (excluded).

Add function rsa_make_key_bn_e to use a bignumber public exponent.

Signed-off-by: Cedric Neveux <cedric.neveux@nxp.com>

### Checklist

### DIFF
--- a/src/headers/tomcrypt_pk.h
+++ b/src/headers/tomcrypt_pk.h
@@ -45,7 +45,8 @@ typedef struct Rsa_key {
 } rsa_key;
 
 int rsa_make_key(prng_state *prng, int wprng, int size, long e, rsa_key *key);
-
+int rsa_make_key_ubin_e(prng_state *prng, int wprng, int size,
+                        const unsigned char *e, unsigned long elen, rsa_key *key);
 int rsa_get_size(const rsa_key *key);
 
 int rsa_exptmod(const unsigned char *in,   unsigned long inlen,
--- a/src/pk/rsa/rsa_make_key.c
+++ b/src/pk/rsa/rsa_make_key.c
@@ -9,51 +9,37 @@
 
 #ifdef LTC_MRSA
 
-/**
-   Create an RSA key
-   @param prng     An active PRNG state
-   @param wprng    The index of the PRNG desired
-   @param size     The size of the modulus (key size) desired (octets)
-   @param e        The "e" value (public key).  e==65537 is a good choice
-   @param key      [out] Destination of a newly created private key pair
-   @return CRYPT_OK if successful, upon error all allocated ram is freed
-*/
-int rsa_make_key(prng_state *prng, int wprng, int size, long e, rsa_key *key)
+static int s_rsa_make_key(prng_state *prng, int wprng, int size, void *e, rsa_key *key)
 {
-   void *p, *q, *tmp1, *tmp2, *tmp3;
+   void *p, *q, *tmp1, *tmp2;
    int    err;
 
    LTC_ARGCHK(ltc_mp.name != NULL);
    LTC_ARGCHK(key         != NULL);
    LTC_ARGCHK(size        > 0);
 
-   if ((e < 3) || ((e & 1) == 0)) {
-      return CRYPT_INVALID_ARG;
-   }
-
    if ((err = prng_is_valid(wprng)) != CRYPT_OK) {
       return err;
    }
 
-   if ((err = mp_init_multi(&p, &q, &tmp1, &tmp2, &tmp3, NULL)) != CRYPT_OK) {
+   if ((err = mp_init_multi(&p, &q, &tmp1, &tmp2, NULL)) != CRYPT_OK) {
       return err;
    }
 
    /* make primes p and q (optimization provided by Wayne Scott) */
-   if ((err = mp_set_int(tmp3, e)) != CRYPT_OK)                      { goto cleanup; }  /* tmp3 = e */
 
    /* make prime "p" */
    do {
        if ((err = rand_prime( p, size/2, prng, wprng)) != CRYPT_OK)  { goto cleanup; }
        if ((err = mp_sub_d( p, 1,  tmp1)) != CRYPT_OK)               { goto cleanup; }  /* tmp1 = p-1 */
-       if ((err = mp_gcd( tmp1,  tmp3,  tmp2)) != CRYPT_OK)          { goto cleanup; }  /* tmp2 = gcd(p-1, e) */
+       if ((err = mp_gcd( tmp1,  e,  tmp2)) != CRYPT_OK)             { goto cleanup; }  /* tmp2 = gcd(p-1, e) */
    } while (mp_cmp_d( tmp2, 1) != 0);                                                  /* while e divides p-1 */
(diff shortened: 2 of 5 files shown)


src/headers/tomcrypt_pk.h
```diff
@@ -45,6 +45,7 @@ typedef struct Rsa_key {
 } rsa_key;
 
 int rsa_make_key(prng_state *prng, int wprng, int size, long e, rsa_key *key);
+int rsa_make_key_bn_e(prng_state *prng, int wprng, int size, void *e, rsa_key *key);
```
sjaeckel: I like the idea, but we don't accept API's anymore which have MPI-specific arguments.

In the spirit of the `rsa_set_xyz()` functions could you please change it to:

```suggestion
int rsa_make_key_bn_e(prng_state *prng, int wprng, int size, const unsigned char *e, unsigned long elen, rsa_key *key);
```

I'm also not too much a fan of the name `rsa_make_key_bn_e()` but I don't have a better idea right now. Maybe you @karel-m ?

I didn't review the implementation yet, will do so after you changed the API.

Please also make sure to have a look at the checklist and tick those two boxes ;)

cneveux: @sjaeckel 
Sure I can change the function prototype as you suggested.
Concerning the function name, I'm not fan with it too but it was the most explicit one I found. Otherwise I can propose;
- `rsa_make_key2`
- `rsa_make_key_ex`

I'll add a test and check documentation. I just wanted to be sure first that idea was accepted. 
Do you want separated commit for the test?

sjaeckel: > Concerning the function name, I'm not fan with it too but it was the most explicit one I found.

yes, you've already chosen the best of those three proposals :)

> Do you want separated commit for the test?

I don't really care :)

It's nice if commits are separate and self-contained, but as long as it's all there it's fine.

karel-m: I would choose: `int rsa_make_key_hex_e(prng_state *prng, int wprng, int size, const char *hex_e, rsa_key *key);`

sjaeckel: I kind of like the idea, but I don't like the difference in API to the `rsa_set_*()` functions ... also I'm not sure whether that's really such a good idea as hex is maybe a good representation for "human interaction" but I'm pretty sure that "on the wire" this is going to be something else than a hex string ... even though in some use cases it's going to be `base16_decode() && rsa_make_key_bn_e()` ...

I've just looked into the use case of @cneveux and the op-tee integration would indeed be easiest with the initial approach where `e` is a "mpi `void*`", so I think we should take the more efficient variant, both runtime and memory wise and that's `int rsa_make_key_bn_e(prng_state *prng, int wprng, int size, const unsigned char *e, unsigned long elen, rsa_key *key);`

Fine @karel-m ?

sjaeckel: As op-tee is statically linking against ltc we could also provide a `int rsa_make_key_bn_e(prng_state *prng, int wprng, int size, void *e, rsa_key *key);` in `tomcrypt_private.h` so you don't have to convert back and forth.

@cneveux @jforissier @karel-m what do you think?

jforissier: @sjaeckel sounds good to me

karel-m: Hmm, well, at least leave a comment in `tomcrypt_private.h` like `/* used by op-tee */`

cneveux: ok,
* so I'll add and document this API, plus add a test.
`int rsa_make_key_hex_e(prng_state *prng, int wprng, int size, const char *hex_e, rsa_key *key);`

* and add but no document this API, no test needed too, as `rsa_make_key_hex_e()` and this `rsa_make_key_bn_e()` will call the same static function.  
`int rsa_make_key_bn_e(prng_state *prng, int wprng, int size, void *e, rsa_key *key); /* used by op-tee */` 

Do you agree?

src/headers/tomcrypt_pk.h
```diff
@@ -45,6 +45,10 @@ typedef struct Rsa_key {
 } rsa_key;
 
 int rsa_make_key(prng_state *prng, int wprng, int size, long e, rsa_key *key);
+int rsa_make_key_hex_e(prng_state *prng, int wprng, int size,
+                       const unsigned char *e, unsigned long elen, rsa_key *key);
+int rsa_make_key_bn_e(prng_state *prng, int wprng, int size, void *e,
```
sjaeckel: please put this in `tomcrypt_private.h`

src/headers/tomcrypt_pk.h
```diff
@@ -45,6 +45,10 @@ typedef struct Rsa_key {
 } rsa_key;
 
 int rsa_make_key(prng_state *prng, int wprng, int size, long e, rsa_key *key);
+int rsa_make_key_hex_e(prng_state *prng, int wprng, int size,
```
sjaeckel: I just had the idea that calling it `rsa_make_key_ubin_e()` would be more appropriate, what do you think?

`hex` would be reserved for a string-ingesting version as proposed by @karel-m  :)

src/pk/rsa/rsa_make_key.c
```diff
@@ -90,7 +76,92 @@ int rsa_make_key(prng_state *prng, int wprng, int size, long e, rsa_key *key)
...
+      return err;
+   }
+
+   if ((err = mp_read_unsigned_bin(tmp_e, (unsigned char *)e, elen)) != CRYPT_OK)
+     { goto error; }
+
+   err = rsa_make_key_bn_e(prng, wprng, size, tmp_e, key);
+
+error:
+   mp_clear(tmp_e);
```
sjaeckel: please use the same pattern as in `rsa_make_key()`
```suggestion
   if ((err = mp_read_unsigned_bin(tmp_e, (unsigned char *)e, elen)) == CRYPT_OK)
      err = rsa_make_key_bn_e(prng, wprng, size, tmp_e, key);

   mp_clear(tmp_e);
```

sjaeckel (changes requested):
LGTM

should be ready to merge after those nitpicks are resolved

src/headers/tomcrypt_pk.h
cneveux: sure

src/pk/rsa/rsa_make_key.c
cneveux: ok

src/headers/tomcrypt_pk.h
cneveux: For me, sounds better.

sjaeckel (approved):
Cool thanks

If you rebase now on top of develop I can merge this :)

cneveux:
> 
> 
> Cool thanks
> 
> If you rebase now on top of develop I can merge this :)

Done.

PR #575, opened 2021-12-03
Add SwiftPM support

cntrump:
### Checklist

Build library with libtommath:

```
swift build
```

Run Swift Unit tests:

```
swift test
```


cntrump:
Depend on https://github.com/libtom/libtommath/pull/518

modulemap/module.modulemap
```diff
@@ -0,0 +1,4 @@
+module libtomcrypt [extern_c] {
+    header "../src/headers/tomcrypt.h"
```
sjaeckel, line 2: If libtom/libtommath#518 is really the right way to do stuff, it would also mean that this should be changed to allow using ltc in other modules or not?

cntrump: > it would also mean that this should be changed to allow using ltc in other modules or not?

Yes, should be changed. if a C package depend on libtomcrypt. because C use `#include <tomcrypt.h>` in source.
No, if a Swift/ObjC package depend on libtomcyrpt. swift/objc use `import libtomcrypt` in source.

sjaeckel: but it would make sense to support both (C and Swift/ObjC packages depending on this) or not?

cntrump: > but it would make sense to support both (C and Swift/ObjC packages depending on this) or not?

Swift and ObjC can import package as module, module name defined in `module.modulemap`
But C is not support import module, so use `#include <xx>` syntax import symbols.

cntrump: Run `swift build -v`

header search path in log:
```
-I .build/checkouts/libtommath/modulemap
```

Open `Package.swift` in Xcode and build, header search path in log:

```
-I ~/Library/Developer/Xcode/DerivedData/libtomcrypt-bewymzccinmagxaywggcjblpnfww/SourcePackages/checkouts/libtommath/modulemap
```

sjaeckel: https://github.com/libtom/libtommath/blob/2cec6addaf4acbd8250611ec4cecadf1f4d655ee/Package.swift#L29

I'm not sure if this is correct or should instead state `publicHeadersPath: "."`

Like that you would also not need libtom/libtommath#518

cntrump: > I'm not sure if this is correct or should instead state publicHeadersPath: "."

It is challenged. 

Set `publicHeadersPath: "."` path same as `sources: ["."]` and use `-fmodule-map-file=modulemap/module.modulemap`.

It is not work:

```
$ swift test

....
./libtommath/mtest/mpi.h:128:8: error: conflicting types for 'mp_mul'
mp_err mp_mul(mp_int *a, mp_int *b, mp_int *c);
       ^
./libtommath/tommath.h:353:8: note: previous declaration is here
mp_err mp_mul(const mp_int *a, const mp_int *b, mp_int *c) MP_WUR;
       ^
<unknown>:0: error: too many errors emitted, stopping now
./libtommath/demo/tommath_tests.swift:2:8: error: could not build Objective-C module 'libtommath'
import libtommath
       ^

error: fatalError
```

sjaeckel: OK, but you shouldn't build `mtest` in the first place ... also let's move this discussion back to libtom/libtommath#518

PR #596, opened 2022-08-05
Amalgamation

sjaeckel:
### Checklist

-----

This adds support to create and build an amalgamated version of the library, which is maybe useful for some.


eduardsui:
Cool! Just tested it, it works great! Thank you! It still includes aes_tab.c, safe_tab.c, sober128tab.c and whirltab.c, but very easy to manage. I will run some tests on TLSe using the new amalgamated version, but from the first runs it looks absolutely great!

Thanks!

sjaeckel:
> It still includes aes_tab.c, safe_tab.c, sober128tab.c and whirltab.c, but very easy to manage.

better?

levitte (approved):

levitte:
Can't quite see the point with this sort of amalgamation, but if someone finds it useful, why not?

eduardsui (approved):

sjaeckel:
> Can't quite see the point with this sort of amalgamation, but if someone finds it useful, why not?

What do you mean by "can't see the point"? What would you prefer/do instead?

levitte:
> > Can't quite see the point with this sort of amalgamation, but if someone finds it useful, why not?
> 
> What do you mean by "can't see the point"? What would you prefer/do instead?

My remark wasn't specific to libtomcrypt.  Amalgamation showed up on my radar just a few months ago, and I've yet to understand what's the underlying need.  I can understand wanting to amalgamate a gazillion dependencies (I've looked at some rust projects and OMG the number of them!), but otherwise?...

But like I said, if someone does find this useful, yeah ok sure, why not?

sjaeckel:
> [...] I've yet to understand what's the underlying need. 

I guess the most important is "lazyness". You don't want to manage all those libraries, so you simply take the single C files and add them directly to your project.

eduardsui:
I confirm the "lazyness" :). Also, it is simpler to manage when writing wrappers for various interpreters or microcontrollers. A few years ago I've seen this idea on sqlite (which offers an amalgamated version) - I just add it to my project and it just worked. Very easy when checking out libraries, for example: I come across a new XML library and I want to try it by using it in the simplest mode possible. I just write a few lines of C code and just compile the whole thing without any need of makefiles or cmake. This doesn't mean that this is the best way, but it is "quick to try".

levitte:
Ok, I can understand that part.

I immediately find it a pity, then, that C has some unfortunate limitations that makes such things difficult, 'cause for example, static things that are meant for just one translation unit (source file) suddenly become "public" throughout the whole library. I saw a blog on this, by someone who tried to amalgamate OpenSSL.... oh boy.

Side note: I have just recently gotten interested in Zig, BTW, and think it's much more conducive to amalgamation. Nothing I really thought of before, but you got me thinking, so thank you for showing me your points

PR #597, opened 2022-08-16
contextualized extension of the Ed25519 scheme

sa-kib:
This PR adds Ed25519 context support and message pre-hashing support when signin/verifying.

### Checklist

* [x] documentation is updated
* [x] tests are added

### DIFF
--- a/doc/crypt.tex
+++ b/doc/crypt.tex
@@ -5909,25 +5909,49 @@ It has support for the following output formats:
 To sign and/or verify a message use the following functions:
 
 \index{ed25519\_sign}
+\index{ed25519ctx\_sign}
+\index{ed25519ph\_sign}
 \begin{verbatim}
-int ed25519_sign(const unsigned char  *msg, unsigned long msglen,
-                       unsigned char  *sig, unsigned long *siglen,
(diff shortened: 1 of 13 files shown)


sjaeckel:
87fbc596bf5a2a520718c5189154412ecb22fe74 is valid but I think we should keep the tweetnacl code as is (as much as possible). IMO a better fix would be to use `msglen + siglen` in the call of `zeromem()` inside `ed25519_verify_private()`.

sjaeckel:
@sa-kib thanks for the PR!

Please check the changes I did and confirm whether this is fine for you.

I'll rebase&squash then as necessary, force-push to your fork and will merge this PR afterwards.

sa-kib:
@sjaeckel thank you for taking a look!
I'm fine with the changes you've made (we weren't sure if we should touch tweetnacl), thank you for taking care.
This PR originated from OP-TEE Ed25519ph support, so I wanted to gather some feedback from optee community regarding usage of these API calls we introduced.

sjaeckel:
@jenswi-linaro @larperaxis you're the consumers of the API, are you fine with these changes as well?

Issue #603, opened 2022-09-16
Regression due to LTC_MDSA_MAX_GROUP change from 512 to 64


jforissier:
Hi,

Re. https://github.com/libtom/libtomcrypt/pull/546#issuecomment-817942961

The new value of `LTC_MDSA_MAX_GROUP` makes our test suite fail:
```
$ xtest 4006
[...]
o regression_4006.40 Asym Crypto case 157 algo 0x70004131 line 3096
regression_4000.c:668: ret_orig has an unexpected value: 0x3 = TEEC_ORIGIN_TEE, expected 0x4 = TEEC_ORIGIN_TRUSTED_APP
regression_4000.c:4043: ta_crypt_cmd_asymmetric_sign(c, &session, op, algo_params, num_algo_params, ptx_hash, ptx_hash_size, out, &out_size) has an unexpected value: 0xffff3024 = TEE_ERROR_TARGET_DEAD, expected 0x0 = TEEC_SUCCESS
  regression_4006.40 FAILED
```

https://github.com/OP-TEE/optee_test/blob/3.18.0/host/xtest/regression_4000.c#L3096

For the test vector, see https://raw.githubusercontent.com/OP-TEE/optee_test/3.18.0/host/xtest/nist/186-3dsatestvectors.h and search for `ac_dsa_vect106`.

sjaeckel:
I've added that testcase and I hope that the implicit re-licensing is OK! If not I won't merge the PR.

~I'm still waiting for the CI to finish, but my local build succeeds~ CI is green now ... :grimacing: ... maybe this is caused by something else?

jforissier:
Well, I can confirm that reverting the value to 512 makes the test pass on my side... But of course it doesn't mean there isn't something else (we do carry a few local patches on top of upstream libromcrypt in OP-TEE). I will definitely take a closer look on my side. As for the licensing of the test vector, I don't think there is any issue since it comes straight from the NIST test suite.

Thanks for your quick reply, I'll keep you posted.

jforissier:
Well the bug is actually in OP-TEE which passes a DSA key with a wrong `.qord`. @sjaeckel sorry for the noise :confused:

sjaeckel:
I hope this only uncovered a bug in your tests :)

PR #604, opened 2022-09-16
Add regression test for Issue #603

sjaeckel:
Regression test for #603


sjaeckel:
Won't merge since there was nothing to test for.

PR #607, opened 2022-11-14
Added gcm_reset() to gcm_memory() to avoid key leakage

Jinbosh8:
In current design of GCM encryption scheme, the memory space for gcm_state only gets freed at the end of gcm_memory(), which might have a potential risk of key leakage during a crash. A call of `gcm_reset()` was added to prevent this leakage.

### Checklist

* [ N/A] documentation is added or updated
* [ N/A] tests are added or updated

### DIFF
--- a/src/encauth/gcm/gcm_memory.c
+++ b/src/encauth/gcm/gcm_memory.c
@@ -106,6 +106,7 @@ int gcm_memory(      int           cipher,
        err = CRYPT_INVALID_ARG;
     }
 LTC_ERR:
+    gcm_reset(gcm);
     XFREE(orig);
     return err;
 }


src/encauth/gcm/gcm_memory.c
```diff
@@ -106,6 +106,7 @@ int gcm_memory(      int           cipher,
        err = CRYPT_INVALID_ARG;
     }
 LTC_ERR:
+    err = gcm_reset(gcm);
```
sjaeckel:
```suggestion
    gcm_reset(gcm);
```

You're overwriting `err` here with `CRYPT_OK`, which breaks some expectations and lets CI fail.

sjaeckel (changes requested):
:+1: thanks for reporting this

src/encauth/gcm/gcm_memory.c
Jinbosh8: Thank you very much for your suggestion. I've pushed the update.

sjaeckel (approved):

PR #624, opened 2023-06-21
Add support for custom constructed ASN.1 types

### LINKED ISSUE #622: How to encode [0] IMPLICIT SET OF?
### Prerequisites

### Description

How can I construct, using libtomcrypt release 1.18.2, a struct of type `ltc_asn1_list` that, when nested inside a SEQUENCE and encoded, will produce an encoded DER SET OF with an IMPLICIT CONTEXT SENSITIVE [0] tag (as in CMS SignedData certificates, for example)?

If I set `list.type = LTC_ASN1_SETOF`, I get a properly encoded SET but the tag is 0x31 (UNIVERSAL SET).  I can manually change the tag after encoding to 0xA0, except that the SET is nested inside layers of ASN.1 data.  If I use `list.type = LTC_ASN1_CUSTOM_TYPE`, I can get the tag I want, but the function `der_encode_custom_type()` encodes the items like a SEQUENCE, not in sorted order like a DER SET.  If I use the macro `LTC_SET_ASN1_CUSTOM_CONSTRUCTED`, I can nest a SET inside a custom typed SEQUENCE, creating an EXPLICIT tag.

Am I missing something?  Is there a solution in the latest developer version?  It seems that I must either change the tag manually after encoding, or else pre-sort the items manually  within the array `list.data`.  A fix would be to look during `der_encode_custom_type()` for `list.used=LTC_ASN1_SETOF`.

### Version
Release 1.18.2

sjaeckel:

This should fix #622. @stevemit can you please test it for your use case besides the toy example?

### DIFF
--- a/src/headers/tomcrypt_pk.h
+++ b/src/headers/tomcrypt_pk.h
@@ -572,6 +572,14 @@ typedef struct ltc_asn1_list_ {
       LTC_TMPVAR(SAI_list)[LTC_TMPVAR(SAI)].tag = (Tag);                   \
    } while (0)
 
+#define LTC_SET_ASN1_CUSTOM(list, index, Class, Structure, Tag, Type, Data, Size)   \
+   do {                                                                             \
+      int LTC_TMPVAR(SAC) = (index);                                                \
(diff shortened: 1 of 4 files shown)


levitte (approved):

levitte:
I just had a quick look, and ran the test too, and this does appear to answer #622.
I would merge this.

PR #629, opened 2023-08-03
Fixed wrong sign in docu of ECC Extended Key Generation, Fixes #628

### LINKED ISSUE #628: Wrong sign in documentation of 12.3.2 Extended Key Generation
### Prerequisites

* [ x ] Checked the developer manual
* [ x ] Checked that your issue isn't already filed: https://github.com/issues?utf8=✓&q=repo%3Alibtom%2Flibtomcrypt
* [ x ] Checked that your issue isn't due to the fact that you're using asymmetric cryptography and you forgot linking in and/or setting an MPI provider (usually this causes either random crashes or runtime errors like `LTC_ARGCHK 'ltc_mp.name != NULL' failure ...`). c.f. Ch. "Math Descriptors" of the developer manual.
* [ x ] Checked that your issue isn't related to TomsFastMath's limitation that PK operations can by default only be done with max. 2048bit keys

### Description

There seems to by a wrong sign in the elliptic curve formula in chapter 12.3.2 in the documentation, there is `- a` instead of `+ a`. (Probably a reminder of the old form `- 3`.)

### Version

v1.18.2-717-g1e629e6f

### Additional Information

Suggested patch:

```
diff --git a/doc/crypt.tex b/doc/crypt.tex
index 9a383346..eedf054a 100644
--- a/doc/crypt.tex
+++ b/doc/crypt.tex
@@ -5185,7 +5185,7 @@ typedef struct {
 } ltc_ecc_curve;
 \end{verbatim}

-The curve must be of the form $y^2 = x^3 - a \cdot x + b$, and all of the \textit{const char*} parameters have to be encoded in hexadecimal format.
+The curve must be of the form $y^2 = x^3 + a \cdot x + b$, and all of the \textit{const char*} parameters have to be encoded in hexadecimal format.

 Advanced example of creating an ECC key:
 \begin{small}
```

ulikos:

### DIFF
--- a/doc/crypt.tex
+++ b/doc/crypt.tex
@@ -5185,7 +5185,7 @@ typedef struct {
 } ltc_ecc_curve;
 \end{verbatim}
 
-The curve must be of the form $y^2 = x^3 - a \cdot x + b$, and all of the \textit{const char*} parameters have to be encoded in hexadecimal format.
+The curve must be of the form $y^2 = x^3 + a \cdot x + b$, and all of the \textit{const char*} parameters have to be encoded in hexadecimal format.
 
 Advanced example of creating an ECC key:
 \begin{small}


sjaeckel:
Thx for the PR!

PR #633, opened 2023-10-05
Fix AES-NI and other minor stuff

sjaeckel:


karel-m (approved):
looks good, passes all tests with my perl module (on all platforms in the test suite)

Issue #632, opened 2023-09-23
Tiger Hash - Question


NickBrighten:
I have a question about the tiger hash algorithm...

So, in this library the tiger hash algorithm is integrated as it is tiger128,3 / tiger160,3 and tiger192,3...
Is it possible to adopt or integrate the tiger128,4 / tiger160,4 and tiger192,4 algorithm??? Or is any option to get the tigerXXX,4 from existing code???

sjaeckel:
Wow PHP ... I don't know what else I should say :D 

The current code obviously doesn't support it, since it's a special PHP variant of that hash.

It looks like this could be added, one way would be to 

1. add a flag to the existing `struct tiger_state`
2. extend `tiger.c` by a new `const struct ltc_hash_descriptor tiger4_desc` with an unused hash ID and no OID
3. a new init function `tiger4_init()` that calls `tiger_init()` and sets the flag
4. extend the `[s]s_tiger_compress()` by the 4th pass if the flag is set
5. add testvectors (no idea where to get reliable ones from, since not even PHP has them FMU)

NickBrighten:
thx for your comment...
But is it really PHP only???

sjaeckel:
> thx for your comment...

NP - feel free to open a PR once you've implemented it

> But is it really PHP only???

There's https://crypto.stackexchange.com/a/28991 (and https://stackoverflow.com/a/33314933) which are the only online references I can find on it, so I'm pretty sure it's PHP specific.

TBH I'm not really certain whether we should even include it as an option, but since we're a cryptographic toolkit I wouldn't have a too big problem with it.

NickBrighten:
Hey sjaeckel - implementation of the Tiger2 (Tiger algorithm with 4 passes) are very simple - will push soon as possible to you :-)

PR #635, opened 2023-10-15
Add support for Tiger2 hash algorithm...

NickBrighten:
Add support for Tiger2 hash algorithm...


sjaeckel (changes requested):
Thanks for this PR.

I thought I made the way to go clear in #632. We definitely don't want an imaginary "tiger2" which doesn't exist. Please follow the steps in #632 on how to extend the existing implementation.

Also please don't commit your personal custom configuration.

MarekKnapek:
I would like to inform @sjaeckel that hash named "Tiger2" indeed does exist. It is described on author's web site at http://www.cs.technion.ac.il/~biham/Reports/Tiger/ you might want to use the web.archive.org service to obtain older copy of that web site. The only difference between Tiger and Tiger2 hashes is padding scheme, in Tiger2 it is changed in such a way to be the same as in MD and SHA family of hashes. There is plenty of on-line hash calculators which could be used to verify correctness of your implementation. I created one of such calculators at https://marekknapek.github.io/hash/

sjaeckel:
@MarekKnapek Thanks for pointing this out.

It shows an even better reason why the name "tiger2" as chosen by @NickBrighten would have been a very bad idea.

NickBrighten:
> @MarekKnapek Thanks for pointing this out.
> 
> It shows an even better reason why the name "tiger2" as chosen by @NickBrighten would have been a very bad idea.

don't know what's wrong with the name - it comes from the developer himself - you can read it at : http://www.cs.technion.ac.il/~biham/Reports/Tiger/ and https://en.wikipedia.org/wiki/Tiger_(hash_function)...
In both, there is a tiger2

It is irrelevant if the only change is the padding scheme or what else - if the developer of the source rename it, you will be adopt it or ignore it, but no discussion about the name...

sjaeckel:
> don't know what's wrong with the name - it comes from the developer himself [...] no discussion about the name...

I think there's a misunderstanding here. Indeed, there should be no discussion about the name.

You implemented the variation of `tiger` with more rounds and called it `tiger2`, even though it's still the `tiger` algorithm.

As you also stated there's a variation of the `tiger` algorithm, which the author called `tiger2`, which has a different padding scheme. This one differs from "what you called `tiger2`". How should we call the "real `tiger2` (with different padding scheme)" then if we'd ever integrate it? `tiger3`? `tiger2-but-the-real-one`?

Clear now what the problem is with you calling it `tiger2` instead of `tiger`?

PR #641, opened 2024-02-26
Simplify AES-NI detection logic

sjaeckel:
Based on the patch provided by @thesamesam via [0], but slighly modified.

@thesamesam could you maybe test whether this patch is fine as well?

[0] https://github.com/DCIT/perl-CryptX/pull/99


thesamesam:
Let me have a go. Thanks!

I don't remember if I managed to reproduce the original problem or if I was just able to reason it out, let's see ;)

@hannob Could you try it out on your machine which hit the original Gentoo bug too?

hannob:
FWIW as @thesamesam asked me to try to reproduce: I can confirm that with this patch I can build libtomcrypt. I can also build CryptX (patch needs a small adaption).

But I should also point out that I was unable to reproduce the error I got with CryptX in plain libtomcrypt.

levitte (approved):

PR #651, opened 2024-08-17
fix unused variable warning in hashes/tiger

karel-m:
Just cosmestics


sjaeckel (approved):
@karel-m is my second commit fine for you as well?

karel-m:
> @karel-m is my second commit fine for you as well?

It's OK

Issue #176, opened 2017-03-23
Polish API `unsigned long outlen` vs. `unsigned long *outlen`


karel-m:
based on discussion in #171 I have found the following places where we IMO handle incosistently `outlen`

```C
pelican_done(pelican_state *pelmac, unsigned char *out);  /* missing outlen*/
crc32_finish(.... void *hash, unsigned long size);        /* should be: unsigned long *size */
adler32_finish(.... void *hash, unsigned long size);      /* should be: unsigned long *size */
```

sjaeckel:
Looking at these it also became obvious to me that there's different paradigms used for different classes of algorithms... like the `XXhash_done()` functions always use the same signature as `pelican_done()` and TBH I'm not sure yet how much I (dis)like that type of signature...

rofl0r:
well i suppose in the case of crc32_finish it's clear that the result len will be 4 bytes, so there's no need to pass a pointer-to where the size is being written to by the callee.

karel-m:
> well i suppose in the case of crc32_finish it's clear that the result len will be 4 bytes, so there's no need to pass a pointer-to where the size is being written to by the callee.

But it is exactly the same case in many mac/authenc functions returning a tag of fixed length (where we utilize `unsigned long *outlen` parameter)

Issue #174, opened 2017-03-22
Polish API `unsigned char*` vs. `void*`


karel-m:
The discussion started in #171 

instead of:
```C
int fnc(const unsigned char *in, unsigned long inlen, unsigned char *out, unsigned long *outlen)
{
   /* some code */
}
```

use rather:
```C
int fnc(const void *inptr, unsigned long inlen, void *outptr, unsigned long *outlen)
{
   const unsigned char *in = inptr
   unsigned char *out = outptr;
   /* some code */
}
```

rofl0r:
also: in C the proper type for sizes is size_t, not unsigned long. (usually it's the same, but some odd platforms like windows have a 32bit long on 64bit).

sjaeckel:
@rofl0r ty :-)

PR #654, opened 2024-08-20
Constify unsigned_read in ltc_math_descriptor

### LINKED ISSUE #581: src of unsigned_read should be const
The `ltc_math_descriptor.unsigned_read` function pointer has a non-const `src` pointer argument, which the function is not supposed to modify

https://github.com/libtom/libtomcrypt/blob/673f5ce29015a9bba3c96792920a10601b5b0718/src/headers/tomcrypt_math.h#L179

This is error prone as one might accidentally swap dst and src.

The src pointer should get marked as `const`.

```c
int (*unsigned_read)(void *dst,
                     const unsigned char *src,
                     unsigned long  len);
```

levitte:
Drop the corresponding deconstying casts in code.

Fixes #581

### DIFF
--- a/doc/crypt.tex
+++ b/doc/crypt.tex
@@ -9671,9 +9671,9 @@ typedef struct {
      @param len   The number of octets
      @return CRYPT_OK on success
    */
-   int (*unsigned_read)(         void *dst,
-                        unsigned char *src,
-                        unsigned long  len);
+   int (*unsigned_read)(               void *dst,
+                        const unsigned char *src,
+                              unsigned long  len);
 
 /* ---- basic math ---- */
 
--- a/src/headers/tomcrypt_math.h
+++ b/src/headers/tomcrypt_math.h
@@ -175,9 +175,9 @@ typedef struct {
      @param len   The number of octets
      @return CRYPT_OK on success
    */
-   int (*unsigned_read)(         void *dst,
-                        unsigned char *src,
-                        unsigned long  len);
+   int (*unsigned_read)(               void *dst,
+                        const unsigned char *src,
+                              unsigned long  len);
 
 /* ---- basic math ---- */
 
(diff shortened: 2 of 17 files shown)


levitte:
There's oh so much more to do regarding constification, but a lot of it breaks against tomsfastmath, so that will need to change first.

sjaeckel:
Thanks for the PR! I also believe we should constify the API where possible.

This PR is slightly related to #174 #175 and #176

Regarding breaking tfm ... we could also constify the ltc API now and cast the const for tfm away. I checked tfm and it uses the arguments as if they were const anyway.

Maybe we should then introduce a macro instead of adding the cast explicitely? So we don't have to manually fix it again once tfm has been adapted resp. we can keep it as is, if someone has to build against an old tfm.

```
#if TFM_VERSION_something_something
/* oh yeah LTC_TFM_UNCOST is such a bad name */
#define LTC_TFM_UNCOST(type) (type)
#else
#define LTC_TFM_UNCOST(type)
#endif
```

We'd then also need to introduce `TFM_VERSION_something_something` in tfm, so we know what to compare against ;) I've opened https://github.com/libtom/tomsfastmath/pull/36 to tackle that.

For now we could just add the "old path" in ltc and add the version compatibility once it's done in tfm.

sjaeckel:
> I checked tfm and it uses the arguments as if they were const anyway.

haha, naive me :D looking at something and then doing it is sometimes very different ...

levitte:
I *hate* having to cast....  but your idea is sane, and I do like the `TFM_VERSION_3()` thing you added over in tomfastmath.  So, I'm game, and will have at least a draft PR submitted fairly soon, based on `TFM_VERSION_3()`.

levitte:
Have a look at #667, which is the full constification of the math stuff, which also subsumes this PR.

levitte:
Closing this in favor of #667

PR #655, opened 2024-08-20
In workflows, modernise the actions that are used

levitte:
actions/checkout@v2 → actions/checkout@v4
actions/upload-artifact@v3 → actions/upload-artifact@v4

This will reduce the amount of warnings in the workflow summary.


levitte:
Aaaahhhhh, how clean the resulting workflow result is 😃

PR #660, opened 2024-08-20
Make the definition of LTC_WIN32_BCRYPT private

levitte:
1. It's internal anyway, so doesn't quite need to be reflected in a
   public header file.
2. This makes it easy for someone build this library to choose not
   to use BCryptGenRandom() by simply definining LTC_NO_WIN32_BCRYPT,
   which was hard to reflect in an unmodified public header file.

Alternative to #653

### DIFF
--- a/src/headers/tomcrypt_private.h
+++ b/src/headers/tomcrypt_private.h
@@ -682,3 +682,14 @@ int which ## _export(unsigned char *out, unsigned long *outlen, prng_state *prng
 #else
    #define LTC_BYTE(x, n) (((x) >> (8 * (n))) & 255)
 #endif
+
+/*
+ * On Windows, choose whether to use CryptGenRandom() [older Windows versions]
+ * or BCryptGenRandom() [newer Windows versions].
+ * If CryptGenRandom() is desired, define LTC_NO_WIN32_BCRYPT when building.
+ */
+#if defined(_MSC_VER) && defined(_WIN32_WINNT) && _WIN32_WINNT >= 0x0600
+   #if !defined(LTC_NO_WIN32_BCRYPT)
+      #define LTC_WIN32_BCRYPT
+   #endif
+#endif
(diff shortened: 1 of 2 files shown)


src/headers/tomcrypt_private.h
```diff
@@ -682,3 +682,14 @@ int which ## _export(unsigned char *out, unsigned long *outlen, prng_state *prng
 #else
    #define LTC_BYTE(x, n) (((x) >> (8 * (n))) & 255)
 #endif
+
+/*
+ * on Windows, choose whether to use CryptGenRandom() [olded Windows versions]
+ * or BCryptGenRandom() [newer Windows versions.
+ * If CryptGenRandom() is desired, define LTC_NO_WIN32_BCRYPT when building
```
sjaeckel:
```suggestion
 * On Windows, choose whether to use CryptGenRandom() [older Windows versions]
 * or BCryptGenRandom() [newer Windows versions].
 * If CryptGenRandom() is desired, define LTC_NO_WIN32_BCRYPT when building.
```

sjaeckel (changes requested):
Thanks! That looks a lot better IMO.

After having a quick glance at `tomcrypt_cfg.h` there's most likely a few other things that could be made private.  ... and maybe also a discussion to open, whether it'd make sense to split this then up in two different private headers, one for headers and one for cfg ... or whether cfg should be private in the first place!? ... but let's merge this first to unblock @karel-m

levitte:
I don't see anything wrong with `tomcrypt_cfg.h` being public.

However, `tomcrypt_custom.h` could be a candidate to be made private, but I'm unsure, 'cause if a choice made there affects *other public headers* (such as what the applications get to see), then it isn't a candidate.

sjaeckel (approved):
Please squash & force-push, no need for that silly co-authored tag ;)

Btw. you're free to push feature branches directly to this repo, no need for your fork!

levitte:
> Btw. you're free to push feature branches directly to this repo, no need for your fork!

Oh!  Yeah, that might be practical.  Thank you.

levitte:
Squashed and ready to be merged :-)

Issue #663, opened 2024-09-01
Cygwin (gcc-11.4.0) fails to compile ciphers/sm4.c


karel-m:
Just FYI, I will try to create a pull request with a fix later.

```
gcc -Iltm -Iltc/headers -DLTC_SOURCE -DLTC_NO_TEST -DLTC_NO_PROTOTYPES -DLTM_DESC -DPERL_USE_SAFE_PUTENV -U__STRICT_ANSI__ -D_GNU_SOURCE -ggdb -O2 -pipe -Wall -Werror=format-security -D_FORTIFY_SOURCE=2 -fstack-protector-strong --param=ssp-buffer-size=4 -fdebug-prefix-map=/mnt/share/cygpkgs/perl/perl.x86_64/build=/usr/src/debug/perl-5.36.3-1 -fdebug-prefix-map=/mnt/share/cygpkgs/perl/perl.x86_64/src/perl-5.36.3=/usr/src/debug/perl-5.36.3-1 -fwrapv -fno-strict-aliasing   -O3  -msse4.1 -maes -Wall -Werror -Wextra -DARGTYPE=4 -c ltc/ciphers/sm4.c -o ltc/ciphers/sm4.o
In file included from ltc/headers/tomcrypt.h:12,
                 from ltc/headers/tomcrypt_private.h:4,
                 from ltc/ciphers/sm4.c:18:
ltc/ciphers/sm4.c:140:29: error: expected identifier or ‘(’ before numeric constant
  140 | LTC_INLINE static sm4_u32_t _L(sm4_u32_t B)
      |                             ^~
ltc/ciphers/sm4.c: In function ‘_T’:
ltc/ciphers/sm4.c:150:12: error: called object is not a function or function pointer
  150 |     return _L(t(Z));
      |            ^~
ltc/ciphers/sm4.c:151:1: error: control reaches end of non-void function [-Werror=return-type]
  151 | }
      | ^
At top level:
ltc/ciphers/sm4.c:287:12: error: ‘sm4_self_test_ltc’ defined but not used [-Werror=unused-function]
  287 | static int sm4_self_test_ltc(void)
      |            ^~~~~~~~~~~~~~~~~
cc1: all warnings being treated as errors
make[1]: *** [Makefile:199: ltc/ciphers/sm4.o] Error 1
```

```
$ gcc -v
Using built-in specs.
COLLECT_GCC=gcc
[...]

PR #664, opened 2024-09-01
Cygwin fix for SM4 (issue #663)

karel-m:
Fix for #663
* Static functions renamed (using `s_` prefix) so that they do not collide with cygwin stuff
* Added missing `#ifdef LTC_TEST`

### DIFF
--- a/src/ciphers/sm4.c
+++ b/src/ciphers/sm4.c
@@ -28,7 +28,7 @@ typedef ulong32 sm4_u32_t;
  * S-box defined in section 6.2
  * (1) Nonlinear transformation
  */
-static const sm4_u8_t sbox_table[16][16] = {
+static const sm4_u8_t sm4_sbox_table[16][16] = {
     {0xd6, 0x90, 0xe9, 0xfe, 0xcc, 0xe1, 0x3d, 0xb7,
         0x16, 0xb6, 0x14, 0xc2, 0x28, 0xfb, 0x2c, 0x05},
     {0x2b, 0x67, 0x9a, 0x76, 0x2a, 0xbe, 0x04, 0xc3,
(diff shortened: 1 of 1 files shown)


src/ciphers/sm4.c
```diff
@@ -67,7 +67,7 @@ static const sm4_u8_t sbox_table[16][16] = {
  * S-box
  * defined in section 2.6 S-box
  */
-LTC_INLINE static sm4_u8_t sbox(sm4_u8_t a)
+LTC_INLINE static sm4_u8_t s_sbox(sm4_u8_t a)
```
sjaeckel: I'd suggest to simply namespace all the functions, but leave their original name intact.

```suggestion
LTC_INLINE static sm4_u8_t s_sm4_sbox(sm4_u8_t a)
```

src/ciphers/sm4.c
```diff
@@ -80,35 +80,35 @@ LTC_INLINE static sm4_u8_t sbox(sm4_u8_t a)
  * But we just convert a 32bit word byte by byte.
  * So it's OK if we don't convert the endian order
  */
-LTC_INLINE static sm4_u32_t t(sm4_u32_t A)
+LTC_INLINE static sm4_u32_t s_trans(sm4_u32_t A)
```
sjaeckel:
```suggestion
LTC_INLINE static sm4_u32_t s_sm4_t(sm4_u32_t A)
```

and so on.

sjaeckel (changes requested):
Like that we also would prevent name clashes as seen when trying to do #596 and the names are still kind of the same from the original and not something entirely different.

src/ciphers/sm4.c
karel-m: done

src/ciphers/sm4.c
karel-m: done

levitte (approved):
LGTM

sjaeckel (approved):

PR #667, opened 2024-09-17
Constify math descriptors

### LINKED ISSUE #581: src of unsigned_read should be const

levitte:
Most of all, this adds proper constification to `ltc_math_descriptor`,
the rest is suffering the consequences.

For TomsFastMath specifically, we add unconsting casts so LTM can be compiled
to TFM versions before it gets properly constified.  This is done through a
version checking helper macro that is added in libtom/tomsfastmath#36.
Note that the current TFM release isn't properly constified, but hopefully,
0.14.0 will be.

### DIFF
--- a/demos/tv_gen.c
+++ b/demos/tv_gen.c
@@ -678,11 +678,11 @@ static void ecc_gen(void)
         fprintf(out, "%s\n", ltc_ecc_curves[x].OID);
         mp_set(k, 1);
 
-        mp_read_radix(order,   (char *)ltc_ecc_curves[x].order, 16);
-        mp_read_radix(modulus, (char *)ltc_ecc_curves[x].prime, 16);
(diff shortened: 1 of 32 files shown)


levitte (commented):
Saw your added changes, @sjaeckel, and approve

levitte:
This looks ready to merge

PR #675, opened 2024-11-08
Align demo/CMakeLists.txt a bit more with with makefile_include.mk

levitte:
Now, they build the same executables.


sjaeckel:
I have some more changes in this regard locally, I'll integrated what you did here next week and push to this PR if that's OK for you!?

levitte:
I just rebased it on top of a fresh `develop`.  Was that what you intended to do?

Either way, if you have more updates, it's OK to integrate them with this patch

levitte:
Closing in favor of #677, where this patch is incorporated

PR #676, opened 2024-11-15
Add a SBOM file in CycloneDX format

hughsie:
Hi,

My name is Richard Hughes and I'm a developer at Red Hat. I'm the maintainer of fwupd and LVFS, and am trying to improve software supply chain security by encouraging OEMs, ODMs and IBVs to ship Software Bill of Materials with each firmware binary blob (SBOMs).

I'm working alongside lots of other companies proactively trying to do the right thing. The reason I've opened this pull request is because your project is either used in the *build process* of a firmware we care about (e.g. EDK II, or coreboot) or is built *into the firmware binary* itself. Although my personal focus is on firmware, the SBOM file is in CycloneDX format (one of the most popular industry standards) which makes it also useful when building containers or OS images too.

I would like to contribute this template SBOM file into your project that gets included into source control with substituted values that get populated automatically. I'm not super familiar with your project, and so I've done my best populating the project values -- but please point out any that are incorrect and I'll fix them up. I've also put the `sbom.cdx.json` file in what I feel is the right place, but please say if you want me to put it somewhere different or name it a different thing; the directory and `sbom` prefix are unimportant.

I've written a bit more about this proposal here https://blogs.gnome.org/hughsie/2024/11/14/firmware-sboms-for-open-source-projects/ and there's also lot more information about firmware SBOMs here: https://lvfs.readthedocs.io/en/latest/sbom.html – many thanks for your time.

### DIFF
--- a/contrib/sbom.cdx.json
+++ b/contrib/sbom.cdx.json
@@ -1,0 +1,40 @@
+{
+  "bomFormat": "CycloneDX",
+  "specVersion": "1.6",
+  "version": 1,
+  "metadata": {
+    "authors": [
+      {
+        "name": "@VCS_SBOM_AUTHORS@"
+      }
+    ]
+  },
+  "components": [
+    {
+      "type": "library",
+      "bom-ref": "pkg:github/libtom/libtomcrypt@@VCS_TAG@",
+      "cpe": "cpe:2.3:a:libtom:libtomcrypt:@VCS_TAG@:*:*:*:*:*:*:*",
+      "name": "libtomcrypt",
+      "version": "@VCS_VERSION@",
+      "description": "Modular and portable cryptographic toolkit",
+      "authors": [
+        {
+          "name": "libtomcrypt developers"
+        }
+      ],
+      "licenses": [
+        {
+          "license": {
+            "id": "Unlicense"
+          }
+        }
+      ],
+      "externalReferences": [
+        {
+          "type": "vcs",
+          "url": "https://github.com/libtom/libtomcrypt"
+        }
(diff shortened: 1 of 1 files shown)


sbom.cdx.json
```diff
@@ -0,0 +1,38 @@
...
+    {
+      "type": "library",
+      "bom-ref": "pkg:github/libtom/libtomcrypt@@VCS_TAG@",
+      "cpe": "cpe:2.3:a:libtom:libtomcrypt:@VCS_TAG@:*:*:*:*:*:*:*",
+      "name": "libtomcrypt",
+      "version": "@VCS_VERSION@",
+      "description": "Modular and portable cryptographic toolkit",
+      "supplier": {
+        "name": "libtomcrypt developers"
+      },
```
levitte: @libtom/libtomcrypt, FYI: if "supplier" feels uncomfortable, the other option is to have this instead:

``` json
      "authors": {
        "name": "libtomcrypt developers"
      },
```

Alternatively, all the authors can be enumerated:

``` json
      "authors": {
        "name": "@VCS_AUTHORS@"
      },
```

hughsie: Yes, absolutely. I'm happy to fix up the PR with either option.

sjaeckel: I guess option 1 would be my preferred way to go.

And I like the idea of having this file in a `contrib` folder.

hughsie: Understood, thanks for the feedback. I've updated the PR with option 1 and moved it into a new contrib folder.

PR #680, opened 2025-01-21
Have src/headers/tomcrypt_private.h include stdarg.h

### LINKED ISSUE #679: Information about compilation failure under GitHub's CI
Hi @sjaeckel, I'm replying here because GitHub does not load comments under the commit anymore

Image

This is my setup:
Latest Ubuntu runner, which can also repro inside ubuntu noble image as well when using containers inside GitHub's CI. Can also repro with GCC. GCC error message is more useful as it suggests adding <stdarg.h> to the private header file (although I don't know if that's the correct solution).

https://github.com/diamante0018/master-tool/blob/master/.github/workflows/build.yml#L70 (I'm simply starting a build of a project that uses libtomcrypt with clang and compilation fails when building the library)
Image

levitte:
This ensures that `va_list`, which is used in that file, gets properly
declared.

As a consequence, the following pattern in other files can be reduced:

    #include "tomcrypt_private.h"
    #include <stdarg.h>

Fixes #679

### DIFF
--- a/src/encauth/siv/siv.c
+++ b/src/encauth/siv/siv.c
@@ -1,7 +1,6 @@
 /* LibTomCrypt, modular cryptographic library -- Tom St Denis */
 /* SPDX-License-Identifier: Unlicense */
 #include "tomcrypt_private.h"
-#include <stdarg.h>
 
 /**
   @file siv.c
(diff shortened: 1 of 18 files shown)


sjaeckel (approved):
Thanks!

PR #683, opened 2025-02-17
Some improvements

sjaeckel:
* Ubuntu 20.04 is not supported any longer as CI runner
* ILP32 is a valid ABI on AIX as well


levitte (approved):
For all parts I didn't add myself...

PR #689, opened 2025-05-03
Fix CMake configure on Windows

momo5502:
Calling `target_link_libraries` before a target has been added seems to cause errors during CMake configuration.

Since BCrypt was fixed in b508028b141a951aa9c41d0cddb815ed4ff950ed, this error seems to occur.

Moving the `target_link_libraries` call after the library has been added fixes this.


levitte (approved):

sjaeckel (approved):
I took the liberty to add a commit that runs the CMake builds on Appveyor.

https://ci.appveyor.com/project/libtom/libtomcrypt/builds/52001098 also shows that the CMake build on Windows was broken before your patch, so thanks a lot for that!

PR #692, opened 2025-06-04
Add TurboSHAKE XOF functions.

### LINKED ISSUE #691: Add TurboSHAKE XOF functions.
TurboSHAKE128 and TurboSHAKE256 as described in [https://keccak.team/files/TurboSHAKE.pdf](https://keccak.team/files/TurboSHAKE.pdf).

MarekKnapek:
TurboSHAKE128 and TurboSHAKE256 as described in https://keccak.team/files/TurboSHAKE.pdf.

https://github.com/libtom/libtomcrypt/issues/691

### Checklist


levitte:
That PDF refers to https://datatracker.ietf.org/doc/draft-irtf-cfrg-kangarootwelve/, which also describes the algo, *and provides test vectors*.  Might I suggest adding some tests, based on those test vectors?

sjaeckel:
And TBH I'd prefer to have this only merged once the RFC is finalized.
I've been burned too often by standards changing in some future version before release.

MarekKnapek:
I updated my branch, I added some unit tests. Regarding the standard being in draft stage. It seems to be draft since 2019-08-06 until 2025-02-21 and continuing to be draft until today, not sure what changed in the meantime. How would you like to continue? I have few ideas:

 - Keep the TurboSHAKE functions in the library, but behind some `#ifdef` that clearly communicates to the user that this feature is "beta" or "standard not finished yet" and that it might change in the future or something like that.
 - Do not add this functionality to your product until the RFC standardization is done. (You seem to like this approach.)
 - Add this functionality to your product and if it breaks in the future, say to your customers "bad luck" your problem using unfinished standard, not mine. (You seem to don't like this approach.)
 - Any more ideas?

levitte:
I could go with having this as an unstable feature, only to be enabled on demand...  especially if libtomcrypt already have infrastructure for that sort of thing (er, I forget if it has).  If that ends up too hard, I'd go with keeping this PR as is until the standard is finalized.

I was a bit involved in the addition of TLS 1.3 in OpenSSL, and we did hold off a release with that until the RFC was actually published... and we did see some changes very late in the RFC process.  So that is to say, there's quite some precedence to @sjaeckel's worries.

sjaeckel:
Thanks for adding the tests and thanks for your input!

So a long lived PR this will be, until cfrg comes to a conclusion ...

sjaeckel:
> So a long lived PR this will be, until cfrg comes to a conclusion ...

That was faster than expected - https://datatracker.ietf.org/doc/rfc9861/

MarekKnapek:
Nice. I just checked the test vectors with the RFC. They are the same. I updated my branch to reflect the RFC number.

MarekKnapek:
Hi @sjaeckel, do you like this feature and this pull request? Are you willing to consider including it in the LibTomCrypt project? Should I improve this code or commit message in some way? Should I add the KT12 hash also?

sjaeckel (changes requested):
Thanks again for the PR :)

> do you like this feature and this pull request? Are you willing to consider including it in the LibTomCrypt project?  Should I add the KT12 hash also?

yes to all those :) and thanks already in advance!

> Should I improve this code or commit message in some way?

Not sure about the code ... The newly added functions are 99% copies, would it maybe make sense to add generic implementations and call them? (or did I miss some other change in the copies?)

```c
static LTC_INLINE void s_keccak_f(ulong64 s[25], int max_rounds, int rc_offset)
{
   /* ... */
}

static void s_keccakf(ulong64 s[25])
{
   s_keccak_f(s, SHA3_KECCAK_ROUNDS, 0);
}
static void s_keccak_turbo_f(ulong64 s[25])
{
   s_keccak_f(s, SHA3_KECCAK_TURBO_ROUNDS, SHA3_KECCAK_TURBO_RC_OFFSET);
}

typedef void (*process_fn)(ulong64 s[25]);
static LTC_INLINE int s_sha3_process(hash_state *md, const unsigned char *in, unsigned long inlen, process_fn proc_f)
{
   /* ... */
}

int sha3_process(hash_state *md, const unsigned char *in, unsigned long inlen)
{
   return s_sha3_process(md, in, inlen, s_keccakf);
}
int sha3_turbo_process(hash_state *md, const unsigned char *in, unsigned long inlen)
{
   return s_sha3_process(md, in, inlen, s_keccak_turbo_f);
}
```

MarekKnapek:
Yes, the functions are almost identical. I'm not sure about your / the project's philosophy. I have two ideas. One is to make two separate "main" functions having all the meat-and-potatoes in them. Second option is to have single function with run-time parameter switching between Shake and TurboShake. There are pros and cons for both options.

Two separate functions. As it is today in my pull request. Pros: If the user wants only one functionality Shake or TurboShake, they are not bothered with the other one. Possibly making the code a bit faster (no need to do run-time if). Possibly making the code smaller (not including the other functionality). I think this is the philosophy of LibTomCrypt - include only what you use and don't pay for what you don't use. Implementation wise: There is copy & pasta so this could cause confusion for anybody reading the code. There is high maintainability cost of having duplicate code. I could use C++ templates to instantiate two versions of the function having only single source code and switching between them at compile-time (think of template dispatch or constexpr-if). But this project is not C++, in C I could use macros, x-macros or some similar shenanigans. But I guess this would not fly.

Single function with run-time parameter. Pros: Simpler code, better understandability, better maintainability, smaller code for those that need both. Cons: Bigger code, slower code for those that need only single one.

Should I consolidate the two functions with added boolean parameter switching between the two variants?

// Edit: Instead of boolean parameter, use number of rounds and round starting offset as you suggested.

sjaeckel:
I guess you're missing something here.

> But this project is not C++, in C I could use macros, x-macros or some similar shenanigans. But I guess this would not fly.

Why so complicated? :slightly_smiling_face: 

> Single function with run-time parameter. Pros: Simpler code, better understandability, better maintainability, smaller code for those that need both. Cons: Bigger code, slower code for those that need only single one.

Simply use an optionally inline function, as suggested. It combines the best of both worlds. The code is simple and the same, so we have less LOC. If someone cares about compiled code size they disable the inline function and get a shared implementation. If they want speed they keep inline enabled.

I took the liberty to do those changes and squashed it with the other two commits together at https://github.com/libtom/libtomcrypt/tree/turboshake (please check, maybe you want to keep some part of the other commit messages, I simply only kept the first one)

sjaeckel (changes requested):
Minor nitpick I just saw.

```
sha3_shake_turbo_process
sha3_shake_turbo_done
sha3_turbo_shake_test
```

I'd propose to rename them to:

```
turbo_shake_process
turbo_shake_done
turbo_shake_test
```
since the official name is TurboSHAKE.

Also there's
a) no need for a separate `sha3_turbo_process`
b) `turbo_shake_init` is missing
c) please add the option to separately enable it via `LTC_TURBO_SHAKE`, similar to `LTC_KECCAK`

And please squash everything together in a single commit :)

sjaeckel:
@MarekKnapek ping?

MarekKnapek:
Pong.
 - Changed the main `keccak` function to take number of rounds and round offset.
 - Changed the `sha3` and `turbo_shake` functions to pass a function pointer.
 - Incorporated the code from your branch.
 - Renamed the functions from `sha3_shake_turbo_*` to `turbo_shake_*`.
 - Added `LTC_TURBO_SHAKE` compile time option.

src/hashes/sha3_test.c
```diff
@@ -388,6 +388,183 @@ int sha3_shake_test(void)
...
+   turbo_shake_done(&c, hash, 64);
+   if (compare_testvector(hash, 64, turbo_shake_128_empty, sizeof(turbo_shake_128_empty), "TurboSHAKE128", 0)) {
+      return CRYPT_FAIL_TESTVECTOR;
+   }
+
+   /* TurboSHAKE128 on an empty buffer, digest length 10032 bytes, test last 32 bytes */
+   sha3_shake_init(&c, 128);
+   for(i = 0; i != 10000 / 10; ++i){ turbo_shake_done(&c, hash, 10); }
+   turbo_shake_done(&c, hash, 32);
+   if (compare_testvector(hash, 32, turbo_shake_128_empty_10032, sizeof(turbo_shake_128_empty_10032), "TurboSHAKE128", 0)) {
```
sjaeckel:
```suggestion
   if (compare_testvector(hash, 32, turbo_shake_128_empty_10032, sizeof(turbo_shake_128_empty_10032), "TurboSHAKE128", 1)) {
```

src/hashes/sha3_test.c
```diff
@@ -388,6 +388,183 @@ int sha3_shake_test(void)
...
+      0xe3, 0x5e, 0x7b, 0x83, 0xe8, 0xb7, 0xe6, 0xeb, 0x4b, 0x78, 0x60, 0x58, 0x80, 0x11, 0x63, 0x16,
+      0xfe, 0x2c, 0x07, 0x8a, 0x09, 0xb9, 0x4a, 0xd7, 0xb8, 0x21, 0x3c, 0x0a, 0x73, 0x8b, 0x65, 0xc0,
+   };
+
+   /* TurboSHAKE128 on an empty buffer */
+   sha3_shake_init(&c, 128);
+   turbo_shake_done(&c, hash, 64);
+   if (compare_testvector(hash, 64, turbo_shake_128_empty, sizeof(turbo_shake_128_empty), "TurboSHAKE128", 0)) {
+      return CRYPT_FAIL_TESTVECTOR;
+   }
```
sjaeckel:
```suggestion
   COMPARE_TESTVECTOR(hash, 64, turbo_shake_128_empty, sizeof(turbo_shake_128_empty), "TurboSHAKE128", 0);
```
We usually use the `COMPARE_TESTVECTOR()` macro nowadays. This didn't exist when this file was introduced, so it's a bit mixed style, but I'd prefer to use it.

src/hashes/sha3_test.c
```diff
@@ -388,6 +388,183 @@ int sha3_shake_test(void)
...
+
+   /* TurboSHAKE128 on an empty buffer, digest length 10032 bytes, test last 32 bytes */
+   sha3_shake_init(&c, 128);
+   for(i = 0; i != 10000 / 10; ++i){ turbo_shake_done(&c, hash, 10); }
+   turbo_shake_done(&c, hash, 32);
+   if (compare_testvector(hash, 32, turbo_shake_128_empty_10032, sizeof(turbo_shake_128_empty_10032), "TurboSHAKE128", 0)) {
+      return CRYPT_FAIL_TESTVECTOR;
+   }
```
sjaeckel:
```suggestion
   COMPARE_TESTVECTOR(hash, 32, turbo_shake_128_empty_10032, sizeof(turbo_shake_128_empty_10032), "TurboSHAKE128", 1);
```
And could you please increase the counter for each testcase, so one can easily find which one fails if it fails.

sjaeckel (changes requested):
Thanks for the update! Basically LGTM

Now only two things are left AFAICT
1. c.f. the comments
2. run `./helper.pl -a` until it spits no more errors :)

src/hashes/sha3_test.c
sjaeckel: Sorry, I only realized that my suggestion was nonsense, when I saw your implementation :see_no_evil: 

I've merged #707 to fix this, could you please take the two fixup commits I pushed on top of your branch, squash them into your commit, and rebase on top of develop?

sjaeckel (approved):
I'll merge after the rebase.

Thanks for your patience :)

sjaeckel:
Not sure what went wrong when you rebased and pushed, but you had dropped my changes instead of squashing them :grimacing:

I took the liberty of force-pushing the update to your branch and will merge once the build is done.

MarekKnapek:
Ah, I was confused by all the variants. Sometimes function sometimes macro, sometimes without prefix sometimes with prefix. Then the build failed, so I added the include and pushed again. Also, on the first glance, the `sha3_test.c` file might be in a wrong directory as there is a separate `tests` top-level directory. I also tried to edit the TeX docs and add an example usage, but I failed to do so.

MarekKnapek:
Correction: I failed to build something "reasonable" from the TeX source.

In meantime I implemented the KangarooTwelve XOF in 128bit and 256bit variants for my own little project. With unit tests and fuzz testing against the https://github.com/kerukuro/digestpp project. I will open new pull request here later.

sjaeckel:
> Ah, I was confused by all the variants. Sometimes function sometimes macro, sometimes without prefix sometimes with prefix. Then the build failed, so I added the include and pushed again.

Yeah, it's sometimes not obvious what is what and where :grimacing: 

> Also, on the first glance, the `sha3_test.c` file might be in a wrong directory as there is a separate `tests` top-level directory.

Yeah ... but no ... Each algorithm provides self-tests, so they're contained in the library sources. The top-level `tests` is more like integration tests.

> I also tried to edit the TeX docs and add an example usage, but I failed to do so.
> Correction: I failed to build something "reasonable" from the TeX source.

No worries, I just added that :)

But it should've been straightforward for you as well.

1. install texlive/latex via your package manager
2. `make docs`

Getting the LaTeX syntax right is another thing one has to get used to ...

MarekKnapek:
- Oh, I was running `make` in the `docs` directory. It generated HTML DoxyGen and PDF DoxyGen with all the internal functions, variables, struct and macros, I didn't consider that output "reasonable".
 - I installed TeX/LaTeX with additional fonts just fine.
 - I vaguely remember basic LaTeX syntax rules from technical university ~15 years ago.

sjaeckel:
> * Oh, I was running `make` in the `docs` directory. It generated HTML DoxyGen and PDF DoxyGen with all the internal functions, variables, struct and macros, I didn't consider that output "reasonable".

Thanks for the info, I will add some more help in the process of finishing the release.

sjaeckel:
Thanks a lot for this PR and the patience for the back-and-forth of review and so on :)

Looking forward to your K12 PR!

PR #697, opened 2025-09-08
X509 import

sjaeckel:
Start implemeting X.509 APIs.

In #693 this was requested and I was already at it and now we have the first part: Parsing of an X.509 certificate and the cryptographic validation of such a certificate.

NB: This does not yet contain the logic that is required to determine whether a CA is eligible to sign a certificate etc.

### Checklist


karel-m:
@sjaeckel Why was the CI pipeline with tests and builds not triggered after my last commit?

sjaeckel:
I guess because the target branch isn't develop!? Not sure though, just a guess.

sjaeckel:
I took the liberty to squash your adaptations into the main commit. I hope that's fine.

karel-m:
I've just reviewed this PR and pushed a few commits. Two issues I’m still not happy with:
* We are not rejecting certificates that contain unrecognized critical extensions, as required by RFC 5280 section 4.2. I've just added a comment to code about this.
* It is a bit surprising that x509_free takes ltc_x509_certificate ** as its argument. Other _free functions take a single pointer (rsa_free, ecc_free, pka_key_free ...) so x509_free is currently the odd one out.

I also managed to break the CI tests somehow; however, it is too late tonight for me to fix that now.

karel-m (changes requested):
see my commits

src/pk/asn1/x509/x509_extensions.c
```diff
@@ -0,0 +1,668 @@
...
+            break;
+         case LTC_X509_CE_EXT_KEY_USAGE:
+            extensions->ext_key_usage = &extensions_[cur_num];
+            break;
+         default:
+            /* TODO: RFC 5280 4.2 requires rejecting certs with unrecognized critical extensions
+               but currently it cannot be rejected here as some critical extensions are not decoded
+               like: Name Constraints, Inhibit Any Policy, No Revocation Available, OCSP No Check,
+                     Proxy Certificate Information, ac-auditEntity
+            */
```
sjaeckel, line 746: While adding some of them as dummy I came to the conclusion: this is just parsing and not validation. Should we maybe track this as "yes, it's a valid topic, but not now"?

I've collected the following infos on the extensions you mentioned which are tracked outside RFC 5280:

1. [No Revocation Available](https://www.rfc-editor.org/rfc/rfc9608) is `CRITICALITY { FALSE }`
2. [OCSP No Check](https://www.rfc-editor.org/rfc/rfc6960#section-4.2.2.2.1) The RFC says: `The CA does so by including the extension id-pkix-ocsp-nocheck.  This SHOULD be a non-critical extension.  The value of the extension SHALL be NULL.`
3. [Proxy Certificate Information](https://www.rfc-editor.org/rfc/rfc3820#section-3.8) is indeed critical
4. `ac-auditEntity` I couldn't find anything

sjaeckel (commented):
> * We are not rejecting certificates that contain unrecognized critical extensions, as required by RFC 5280 section 4.2. I've just added a comment to code about this.

Good point!

> * It is a bit surprising that x509_free takes ltc_x509_certificate ** as its argument. Other _free functions take a single pointer (rsa_free, ecc_free, pka_key_free ...) so x509_free is currently the odd one out.

Because I like the pattern more. I prefer to be on the safe side, see also the commit I just added which introduces the same pattern for `s_free()`.

karel-m:
~what is the origin of `tests/x509/gnutls/template-krb5name-full.pem` ?~

~cert's KeyUsage is a BIT STRING with all bits zero, which is IMO not valid according to RFC 5280 4.2.1.3~

~should it not be in `gnutls/bad/` ?~

openssl parses/accepts these

UPDATE: openssl rejects it later https://github.com/openssl/openssl/blob/e483d93b393bec71327c3b31779e8dcf37b3f42b/crypto/x509/v3_purp.c#L631

karel-m:
~tests/x509/openssl/embeddedSCTs1_issuer.pem = a self-signed CA root with BasicConstraints: CA:TRUE but not critical~

maybe it is fine as openssl accepts CA:TRUE without critical

PR #698, opened 2025-09-20
Fix build of aesni on msvc

enginelesscc:
This fixes building LTC_AES_NI on msvc, keeps compat with clang-cl

### Checklist

* [x] ran tests (locally)

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -35,7 +35,13 @@ build_script:
       ninja
       cd..
       nmake -f makefile.msvc all
+      cp test.exe test-stock.exe
+      cp timing.exe timing-stock.exe
       nmake -f makefile.msvc clean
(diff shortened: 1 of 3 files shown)


enginelesscc:
I see no [intended] path where I could extend the CI to run with LTC_AES_NI enabled.
What do we do there?
\- enable aesni by default on msvc's makefile?
\- or hardcode the flag in the appveyor.yml?

sjaeckel:
> I see no [intended] path where I could extend the CI to run with LTC_AES_NI enabled.
> What do we do there?
> - enable aesni by default on msvc's makefile?
> - or hardcode the flag in the appveyor.yml?

I see two options
* having another build&test step in the appveyor with AESNI enabled
* adding github actions on Windows support which also provides MSVC IIUC

I guess the first one would be the easiest!?

sjaeckel (approved):
Looks like the patch works as intended!

Thanks :)

Issue #568, opened 2021-04-19
New relese?


kloczek:
Is it possible to make new release soon?
I'm asking because lat latest perl CryptX seems is using featires which are not included in 1.18.2 and are only available in commits added after last release.
Here is build fail log on building perl CryptX against libtomcrypt 1.18.2.
```console
/usr/bin/gcc -c   -D_REENTRANT -D_GNU_SOURCE -O2 -g -grecord-gcc-switches -pipe -Wall -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -Wp,-D_GLIBCXX_ASSERTIONS -specs=/usr/lib/rpm/redhat/redhat-hardened-cc1 -fstack-protector-strong -specs=/usr/lib/rpm/redhat/redhat-annobin-cc1 -m64 -mtune=generic -fasynchronous-unwind-tables -fstack-clash-protection -fcf-protection -fdata-sections -ffunction-sections -flto=auto -flto-partition=none -fwrapv -fno-strict-aliasing -I/usr/local/include -D_LARGEFILE_SOURCE -D_FILE_OFFSET_BITS=64 -Wl,-z,relro -Wl,--as-needed -Wl,--gc-sections -Wl,-z,now -specs=/usr/lib/rpm/redhat/redhat-hardened-ld -flto=auto -flto-partition=none -fuse-linker-plugin   -DVERSION=\"0.071\" -DXS_VERSION=\"0.071\" -fPIC "-I/usr/lib64/perl5/CORE"   CryptX.c
CryptX.xs:35:9: error: unknown type name 'salsa20_state'
   35 | typedef salsa20_state           *Crypt__Stream__Salsa20;
      |         ^~~~~~~~~~~~~
CryptX.xs:36:9: error: unknown type name 'sosemanuk_state'
   36 | typedef sosemanuk_state         *Crypt__Stream__Sosemanuk;
      |         ^~~~~~~~~~~~~~~
CryptX.xs:37:9: error: unknown type name 'rabbit_state'
   37 | typedef rabbit_state            *Crypt__Stream__Rabbit;
      |         ^~~~~~~~~~~~
CryptX.xs:154:3: error: unknown type name 'curve25519_key'
  154 |   curve25519_key key;
      |   ^~~~~~~~~~~~~~
CryptX.xs:161:3: error: unknown type name 'curve25519_key'
  161 |   curve25519_key key;
      |   ^~~~~~~~~~~~~~
[...]

PR #612, opened 2023-03-11
add possibility to use different hash algorithms in RSAES-OAEP

sjaeckel:
The hash algorithms used in the MGF and to create the hash of the Label must not forcibly be the same. This change allows to use different algorithms.

Unfortunately this breaks API if you use `rsa_decrypt_key_ex()`. The `rsa_decrypt_key()` macro is still the same.

### Checklist

* [x] documentation is added or updated
* [x] tests are added or updated

PR #699, opened 2025-10-03
Some improvements

sjaeckel:
I've reordered and refactored some of the commits of #697 in order to limit its scope and simplify the merge and a potential faster release of 2.0

With these changes merged I currently don't see any planned ABI or API breakage (besides #515, but that's an entirely different discussion).

This should be the last PR with intentional changes before a v2.0.0-rc1, c.f. #568

### DIFF
--- a/libtomcrypt_VS2008.vcproj
+++ b/libtomcrypt_VS2008.vcproj
@@ -2281,6 +2281,14 @@
 						RelativePath="src\pk\asn1\x509\x509_encode_subject_public_key_info.c"
 						>
 					</File>
+					<File
+						RelativePath="src\pk\asn1\x509\x509_get_pka.c"
+						>
+					</File>
+					<File
+						RelativePath="src\pk\asn1\x509\x509_import_spki.c"
+						>
+					</File>
 				</Filter>
 			</Filter>
 			<Filter
--- a/makefile.mingw
+++ b/makefile.mingw
@@ -176,9 +176,10 @@ src/pk/asn1/oid/pk_get.o src/pk/asn1/oid/pk_oid_cmp.o src/pk/asn1/oid/pk_oid_str
 src/pk/asn1/pkcs8/pkcs8_decode_flexi.o src/pk/asn1/pkcs8/pkcs8_get.o \
 src/pk/asn1/x509/x509_decode_public_key_from_certificate.o src/pk/asn1/x509/x509_decode_spki.o \
 src/pk/asn1/x509/x509_decode_subject_public_key_info.o \
-src/pk/asn1/x509/x509_encode_subject_public_key_info.o src/pk/dh/dh.o src/pk/dh/dh_check_pubkey.o \
-src/pk/dh/dh_export.o src/pk/dh/dh_export_key.o src/pk/dh/dh_free.o src/pk/dh/dh_generate_key.o \
-src/pk/dh/dh_import.o src/pk/dh/dh_import_pkcs8.o src/pk/dh/dh_set.o src/pk/dh/dh_set_pg_dhparam.o \
+src/pk/asn1/x509/x509_encode_subject_public_key_info.o src/pk/asn1/x509/x509_get_pka.o \
+src/pk/asn1/x509/x509_import_spki.o src/pk/dh/dh.o src/pk/dh/dh_check_pubkey.o src/pk/dh/dh_export.o \
+src/pk/dh/dh_export_key.o src/pk/dh/dh_free.o src/pk/dh/dh_generate_key.o src/pk/dh/dh_import.o \
+src/pk/dh/dh_import_pkcs8.o src/pk/dh/dh_set.o src/pk/dh/dh_set_pg_dhparam.o \
 src/pk/dh/dh_shared_secret.o src/pk/dsa/dsa_decrypt_key.o src/pk/dsa/dsa_encrypt_key.o \
 src/pk/dsa/dsa_export.o src/pk/dsa/dsa_free.o src/pk/dsa/dsa_generate_key.o \
 src/pk/dsa/dsa_generate_pqg.o src/pk/dsa/dsa_import.o src/pk/dsa/dsa_import_pkcs8.o \
--- a/makefile.msvc
+++ b/makefile.msvc
@@ -169,9 +169,10 @@ src/pk/asn1/oid/pk_get.obj src/pk/asn1/oid/pk_oid_cmp.obj src/pk/asn1/oid/pk_oid
 src/pk/asn1/pkcs8/pkcs8_decode_flexi.obj src/pk/asn1/pkcs8/pkcs8_get.obj \
 src/pk/asn1/x509/x509_decode_public_key_from_certificate.obj src/pk/asn1/x509/x509_decode_spki.obj \
 src/pk/asn1/x509/x509_decode_subject_public_key_info.obj \
-src/pk/asn1/x509/x509_encode_subject_public_key_info.obj src/pk/dh/dh.obj src/pk/dh/dh_check_pubkey.obj \
-src/pk/dh/dh_export.obj src/pk/dh/dh_export_key.obj src/pk/dh/dh_free.obj src/pk/dh/dh_generate_key.obj \
-src/pk/dh/dh_import.obj src/pk/dh/dh_import_pkcs8.obj src/pk/dh/dh_set.obj src/pk/dh/dh_set_pg_dhparam.obj \
+src/pk/asn1/x509/x509_encode_subject_public_key_info.obj src/pk/asn1/x509/x509_get_pka.obj \
+src/pk/asn1/x509/x509_import_spki.obj src/pk/dh/dh.obj src/pk/dh/dh_check_pubkey.obj src/pk/dh/dh_export.obj \
+src/pk/dh/dh_export_key.obj src/pk/dh/dh_free.obj src/pk/dh/dh_generate_key.obj src/pk/dh/dh_import.obj \
+src/pk/dh/dh_import_pkcs8.obj src/pk/dh/dh_set.obj src/pk/dh/dh_set_pg_dhparam.obj \
 src/pk/dh/dh_shared_secret.obj src/pk/dsa/dsa_decrypt_key.obj src/pk/dsa/dsa_encrypt_key.obj \
 src/pk/dsa/dsa_export.obj src/pk/dsa/dsa_free.obj src/pk/dsa/dsa_generate_key.obj \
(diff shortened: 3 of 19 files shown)


karel-m:
I have just quickly changed the libtomcrypt in my CryptX perl bindings for the branch "some-improvements" from this PR.

It failed to build (which is probably expected) https://github.com/DCIT/perl-CryptX/actions/runs/18219248768/job/51875345151

I will try to adopt CryptX to the new changes (after that we will see whether CryptX test suite reveals something).

sjaeckel:
> I have just quickly changed the libtomcrypt in my CryptX perl bindings for the branch "some-improvements" from this PR.

Cool, thanks!

> It failed to build (which is probably expected) https://github.com/DCIT/perl-CryptX/actions/runs/18219248768/job/51875345151

If I'm not mistaken there are two errors

[148] ff. is already caused by #524 

[187] ff. is indeed caused by the API change of `rsa_verify_hash_ex()` in this PR

Both are expected, so thanks already in advance for fixing this in CryptX and testing those changes :relaxed:  

[148] https://github.com/DCIT/perl-CryptX/actions/runs/18219248768/job/51875345151#step:7:149
[187] https://github.com/DCIT/perl-CryptX/actions/runs/18219248768/job/51875345151#step:7:188

karel-m:
I have updated my perl bindings and it seems to work fine.

Just one minor suggestion: please add `@param mgf_hash_idx .....` to description of rsa_sign_hash_ex + rsa_verify_hash_ex

karel-m:
I am not sure about having: type = LTC_ECCSIG_RFC5656 (as I understand the "type" is rather a format specification not a flag defining whether to use deterministic signing) e.g. one might want LTC_ECCSIG_RFC5656 + LTC_ECCSIG_ETH27 at the same time.

karel-m:
ad `struct ltc_ecc_sig_opts` could we simplify `int *recid;` to `int recid;` ?

sjaeckel:
> ad `struct ltc_ecc_sig_opts` could we simplify `int *recid;` to `int recid;` ?

Yeah, we could, but that would mean that it must be initialized to -1 if one is not interested in it (which is nearly always the case). With the current implementation via a pointer, it is disabled by default, so it'd be easier to use for the majority of users. But if you think it'd be better as `int recid` feel free to say so! I don't really have a hard opinion on that.

> I am not sure about having: type = LTC_ECCSIG_RFC5656 (as I understand the "type" is rather a format specification not a flag defining whether to use deterministic signing) e.g. one might want LTC_ECCSIG_RFC5656 + LTC_ECCSIG_ETH27 at the same time.

I guess you're confusing RFC5656 and RFC6979 :)

karel-m:
> I guess you're confusing RFC5656 and RFC6979 :)

Oh, yes. Sorry for the noise.

karel-m:
FYI current branch `some-improvements` seems to fail on arm and cywin (will investigate later)

karel-m:
> FYI current branch `some-improvements` seems to fail on arm and cywin (will investigate later)

it is a segfault (memory corruption) related to `ecc_sign_hash_v2` (still not sure if in libtomcrypt or in C code of my perl bindings)

sjaeckel:
> > FYI current branch `some-improvements` seems to fail on arm and cywin (will investigate later)
> 
> it is a segfault (memory corruption) related to `ecc_sign_hash_v2` (still not sure if in libtomcrypt or in C code of my perl bindings)

hmm

```
$ ./test ecc
[...]
SUCCESS: passed=1 failed=0 nop=0 duration=41.5sec real=41.5sec
$ file test                                              
test: ELF 64-bit LSB executable, ARM aarch64, version 1 (GNU/Linux), statically linked, BuildID[sha1]=8aab34cfb2c11a97baefcf2787a078587a6795ea, for GNU/Linux 4.3.0, with debug_info, not stripped
```

At least the tests can't reproduce it for AARCH64 with Qemu ...

karel-m:
on cywin it fails inside `ecc_rfc6979_key` in the first `hmac_memory_multi` (which is strange)

karel-m:
The fix https://github.com/DCIT/perl-CryptX/commit/cccc52bb62e90900afd0317db7e4ffb19339a859

now passes tests also on cygwin + arm

sjaeckel:
> The fix [DCIT/perl-CryptX@cccc52b](https://github.com/DCIT/perl-CryptX/commit/cccc52bb62e90900afd0317db7e4ffb19339a859)

Thanks for triaging it and proposing the fix. Seeing it like that it's obvious now ...

> now passes tests also on cygwin + arm

The last patch should be an equivalent solution to your proposed fix, could you please check again?

karel-m:
> The last patch should be an equivalent solution to your proposed fix, could you please check again?

checked, it is good

src/headers/tomcrypt_pk.h
sjaeckel, line 1: Any comments on the new RSA API?

src/headers/tomcrypt_pk.h
```diff
@@ -95,12 +112,54 @@ int rsa_exptmod(const unsigned char *in,   unsigned long inlen,
...
+   int wprng;
+   prng_state *prng;
+   /* Operation-specific parameters */
+   union {
+      struct {
+         const unsigned char *lparam;
+               unsigned long  lparamlen;
+      } crypt;
+      /* let's make space for potential future extensions */
+      ulong64 dummy[8];
```
sjaeckel: This would allow us to extend functionality without breaking API&ABI in the future.

Is this really necessary for RSA?

Should some ECC struct get something like that?

src/headers/tomcrypt_pkcs.h
sjaeckel, line 1: I've searched through GH and couldn't find anyone using those APIs, so I thought we can simply remove them resp. make them private. People can still speak up and the deprecation can be reverted in the future.

Fine like that?

sjaeckel (commented):
If you're fine with those latest changes [0] I'll add the same macro trickery for the deprecated RSA APIs as well, properly deprecate all the remaining APIS, and then we'll only have to update the docs...

[0] https://github.com/libtom/libtomcrypt/pull/699/files/baf71ad0544d1b1f03b5b644203b6afbe43b9feb..b78da6e65607c23f852ce78583e90025cd2817d9

karel-m:
The old RSA style let you pass two values separately: `int mgf_hash` and `int lparam_hash`. But the new style has only one value: int `hash_idx` (see `rsa_args_to_op_params`).

sjaeckel:
> The old RSA style let you pass two values separately: `int mgf_hash` and `int lparam_hash`. But the new style has only one value: int `hash_idx` (see `rsa_args_to_op_params`).

True. That second parameter was only introduced in #612 and in order to make the API backwards compatible again I've limited the new functionality to the v2 API.

I thought if we already introduce new APIs, we can also do the full deprecation dance. We're not going to be ABI compatible, but the API's would stay the same.

karel-m:
After few hours I am still unable to switch my perl bindings to the new RSA API. I make it to compile but it fails during some operations (enc/dec, sign/verify) and it is not easy for me to debug why.

I have chosen a way like this:
```
          ltc_rsa_op_parameters rsa_oparams = {
            .prng = &self->pstate,
            .wprng = self->pindex,
            .padding = LTC_PKCS_1_OAEP,
            .u.crypt.lparam = lparam_ptr,
            .u.crypt.lparamlen = (unsigned long)lparam_len,
            .params.hash_alg = hash_descriptor[lparam_hash_id].name,
            .params.mgf1_hash_alg = hash_descriptor[mgf_hash_id].name,
            .params.saltlen = 0,
            .params.pss_oaep = 1,
          };
          
          rv = rsa_encrypt_key_v2(data_ptr, (unsigned long)data_len, buffer, &buffer_len, &rsa_oparams, &self->key);
```

but still getting errors like "Invalid PK key or key type specified for function" + on other places errors like "Invalid hash specified" etc.

Maybe I still did not fully understand the new concept (it has definitely much steeper learning curve than the old one).

sjaeckel:
> After few hours I am still unable to switch my perl bindings to the new RSA API. I make it to compile but it fails during some operations (enc/dec, sign/verify) and it is not easy for me to debug why.

ugh, sorry! The tests seem to work fine for me.

I've just pushed my WIP state with the updated macros etc.

sjaeckel:
> but still getting errors like "Invalid PK key or key type specified for function" + on other places errors like "Invalid hash specified" etc.

It could absolutely be that the generic validity checks don't handle everything correctly, I've only tested against the built-in set of tests... :grimacing:

karel-m:
shouldn't the `s_rsa_key_valid_pss_algs` be rather `s_rsa_key_valid_pss_oaep_algs` and allow also padding `LTC_PKCS_1_OAEP`?

karel-m:
The simple example I am not able to implement in the new RSA API (sorry for just a code fragment extracted from my perl bindings):
```
/* load the key in DER format */
rsa_import(data, (unsigned long)data_len, &self->key);

/* try to encrypt */
ltc_rsa_op_parameters rsa_oparams = {
  .prng = &self->pstate,
  .wprng = self->pindex,
  .padding = LTC_PKCS_1_OAEP,
  .u.crypt.lparam = lparam_ptr,
  .u.crypt.lparamlen = (unsigned long)lparam_len,
  .params.hash_alg = hash_descriptor[lparam_hash_id].name,
  .params.mgf1_hash_alg = hash_descriptor[mgf_hash_id].name,
  .params.saltlen = 0,
  .params.pss_oaep = 1,
};
rv = rsa_encrypt_key_v2(data_ptr, (unsigned long)data_len, buffer, &buffer_len, &rsa_oparams, &self->key);
```

sjaeckel:
I'll have a look into it later, but please remove `  .params.pss_oaep = 1,` and it works ... :see_no_evil:

karel-m:
> I'll have a look into it later, but please remove   .params.pss_oaep = 1, and it works ... 

It worked. But IMO a bit surprising to me (as the whole concept of key->params / check->params / ltc_rsa_op_parameters ...)

Anyway, still bunch of other failures (will investigate later).

sjaeckel:
> [...] a bit surprising to me (as the whole concept of key->params / check->params / ltc_rsa_op_parameters ...)

Yeah, a bit, but not really. The `key->params` exist because there's the possibility to restrict the PKCS#1 version via X.509 certificates. The re-use of `ltc_rsa_parameters` inside `ltc_rsa_op_parameters`, came out of laziness because it's kind of shared details (minus the `pss_oaep` field with isn't required there).

Regular RSA keys can be used with all PKCS#1 versions.

Restricted RSA keys ("PSS-keys") must only be used with PKCS#1 versions > 1.5, and come with a set of parameters. There's already `rsa_import_x509.c::rsa_decode_parameters()`, but that one isn't used yet, resp. I'll have to check whether it should be. I've other stuff to finish before continuing on this.

PR #706, opened 2025-11-18
New RSA API

sjaeckel:
Introduce the RSA API v2.

Follow-up of #699 which already contains some comments regarding the new API.

### Checklist

### DIFF
--- a/demos/timing.c
+++ b/demos/timing.c
@@ -736,8 +736,8 @@ static void time_rsa(void)
       .u.crypt.lparamlen = 8,
       .prng = &yarrow_prng,
       .wprng = find_prng("yarrow"),
-      .params.hash_alg = "sha1",
-      .params.mgf1_hash_alg = "sha1",
+      .params.hash_idx = find_hash("sha1"),
+      .params.mgf1_hash_idx = find_hash("sha1"),
       .params.saltlen = 8,
    };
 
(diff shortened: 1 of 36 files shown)


karel-m (commented):
I have committed my suggestions as separate commits.

karel-m:
The updated version with my changes not only passes all libtomcrypt tests, but also my perl CryptX test suite.

sjaeckel (commented):
Thanks a lot for picking this up and working on it! Very much appreciated :)

For now just one question regarding your latest changes.

Why the switch back to the resolved index instead of using the hash name? I like the string-based "way of working" a bit better nowadays, even though it complicates the internal implementation a bit. (Which could certainly be improved by some nifty helpers.)

I still hope we get rid of the descriptors at one point, so I thought we could already future-proof the implementation, but maybe there's another reason I'm not aware of?

karel-m:
ad hash idx vs hash name - it's just that I didn't like those extra round-trips name>>idx>>name (I have no problem with reverting it)

sjaeckel:
> ad hash idx vs hash name - it's just that I didn't like those extra round-trips name>>idx>>name (I have no problem with reverting it)

hmm, I guess we should think in light of "how would resp. could it work if there was no registry" ... which will be "interesting" ...

In that context I just had a look at #515 again and I'm still in favor of merging that one, but now I'm not so sure anymore whether we should go forward and remove all registries, since as @levitte put it "[...] it does cut away the possibility for more complex systems, where there's a part that needs to be able to find implementations by name [...]", which we currently see here.

Therefore let's keep this as it is now and we will see how it evolves.

PR #708, opened 2025-11-27
KangarooTwelve

MarekKnapek:
### Checklist


sjaeckel (changes requested):
I guess you're missing something there.

`hash_state` is a `union`, i.e. it would be valid to do the following:

```c
struct sha3_state st;
sha3_shake_init((hash_state*)&st, 128);
```

https://godbolt.org/z/Wq3oaETcE

Please revert [this change], I'd prefer to keep the codebase of all the algorithms similar.

[this change]: https://github.com/libtom/libtomcrypt/pull/708/commits/2afd1344c90743e4141a60c0ab6c9c04644940b8

src/hashes/sha3_test.c
```diff
@@ -388,161 +388,256 @@ int sha3_shake_test(void)
...
+   LTC_ARGCHK(amount >= 0);
+
+   for(i = 0; i != amount; ++i)
+   {
+      bin[i] =
+         (s_turbo_shake_hex_to_nibble(hex[i * 2 + 0]) << 4) |
+         (s_turbo_shake_hex_to_nibble(hex[i * 2 + 1]) << 0);
+   }
+   return CRYPT_OK;
+}
```
sjaeckel: Could this maybe be done with the `base16_decode()` API?

MarekKnapek:
Yes, that works, but. I need to keep the two SHAKE states inside a KT state between invocations of the init/process/done functions. And because of the union thing, the KT state would be unnecessary big. Keeping state for other hashes that we don't care about inside KT logic. Do you prefer me reverting this or changing every other algorithm to use its own state instead?

// Edit: Oh, I misunderstood. Now I get it. I will do the change you propose.

src/hashes/sha3_test.c
```diff
@@ -388,161 +388,256 @@ int sha3_shake_test(void)
...
+#endif
+
+#ifdef LTC_TURBO_SHAKE
+static LTC_INLINE int s_turbo_shake_test_one(int bits_count, long input_bytes_count, long skip_digest_bytes, long digest_bytes_count, int counter, char const *expected_digest_hex)
```
sjaeckel: I very much like the idea where this is going to! 👍 💯 

How about:


```suggestion
typedef struct turbo_shake_test {
   int bits_count;
   unsigned long input_bytes_count, skip_digest_bytes, digest_bytes_count;
   const char *expected_digest_hex;
} turbo_shake_test;


static LTC_INLINE int s_turbo_shake_test_one(const turbo_shake_test *testcase, int counter)
```

src/hashes/sha3_test.c
```diff
@@ -388,161 +388,256 @@ int sha3_shake_test(void)
...
+   }
+   return CRYPT_OK;
+}
+static LTC_INLINE unsigned char s_turbo_shake_hex_to_nibble(char hex)
+{
+   LTC_ARGCHK((hex >= '0' && hex <= '9') || (hex >= 'a' && hex <= 'f') || (hex >= 'A' && hex <= 'F'));
+
+   return (hex >= '0' && hex <= '9') ? (hex - '0') : ((hex >= 'a' && hex <= 'f') ? (10 + hex - 'a') : (10 + hex - 'A'));
+}
+static LTC_INLINE int s_turbo_shake_hex_to_bin(char const* hex, unsigned char* bin, long amount)
```
sjaeckel: The usual way in C is to use `const char*` as type. They're equivalent but I'd prefer to keep it like that.

src/hashes/sha3_test.c
```diff
@@ -388,161 +388,256 @@ int sha3_shake_test(void)
...
+   if ((err = s_turbo_shake_test_one(256,          17*17*17,     0, 64, counter++, "c74ebc919a5b3b0dd1228185ba02d29ef442d69d3d4276a93efe0bf9a16a7dc0cd4eabadab8cd7a5edd96695f5d360abe09e2c6511a3ec397da3b76b9e1674fb")) != CRYPT_OK) return err;
+   if ((err = s_turbo_shake_test_one(256,       17*17*17*17,     0, 64, counter++, "02cc3a8897e6f4f6ccb6fd46631b1f5207b66c6de9c7b55b2d1a23134a170afdac234eaba9a77cff88c1f020b73724618c5687b362c430b248cd38647f848a1d")) != CRYPT_OK) return err;
+   if ((err = s_turbo_shake_test_one(256,    17*17*17*17*17,     0, 64, counter++, "add53b06543e584b5823f626996aee50fe45ed15f20243a7165485acb4aa76b4ffda75cedf6d8cdc95c332bd56f4b986b58bb17d1778bfc1b1a97545cdf4ec9f")) != CRYPT_OK) return err;
+   if ((err = s_turbo_shake_test_one(256, 17*17*17*17*17*17,     0, 64, counter++, "9e11bc59c24e73993c1484ec66358ef71db74aefd84e123f7800ba9c4853e02cfe701d9e6bb765a304f0dc34a4ee3ba82c410f0da70e86bfbd90ea877c2d6104")) != CRYPT_OK) return err;
```
sjaeckel: And then instead of those n calls

```c
const turbo_shake_test testcases[] = {
   { 128, 0, 0, 32, "1e415f1c5983aff2169217277d17bb538cd945a397ddec541f1ce41af2c1b74c" },
   { 128, 0, 0, 64, "1e415f1c5983aff2169217277d17bb538cd945a397ddec541f1ce41af2c1b74c3e8ccae2a4dae56c84a04c2385c03c15e8193bdf58737363321691c05462c8df" },
   ...
};
for (counter = 0; counter < LTC_ARRAY_SIZE(testcases); counter++) {
   if ((err = s_turbo_shake_test_one(&testcases[counter], counter)) != CRYPT_OK) {
      return err;
   }
}
return CRYPT_OK;
```

sjaeckel (changes requested):
Thanks a lot!

Very good PR :)

Thanks for updating the docs  ❤️ 

What do you think of my suggestions?

src/headers/tomcrypt_custom.h
sjaeckel, line 708:
```suggestion
#if defined(LTC_TURBO_SHAKE) && !defined(LTC_SHA3)
   #error LTC_TURBO_SHAKE requires LTC_SHA3
#endif

#if defined(LTC_KANGAROO_TWELVE) && !defined(LTC_TURBO_SHAKE) 
   #error LTC_KANGAROO_TWELVE requires LTC_TURBO_SHAKE
#endif

```

I already forgot that in the TurboShake PR ... we should check that our dependencies are fulfilled and highlight what's maybe missing.

src/headers/tomcrypt_custom.h
```diff
@@ -259,6 +259,7 @@
 #define LTC_SHA3
 #define LTC_KECCAK
 #define LTC_TURBO_SHAKE
+#define LTC_KT
```
sjaeckel:
```suggestion
#define LTC_KANGAROO_TWELVE
```

Even though I'm a huge fan of short names, `KT` is even too short for me 😜 

And the API should also be renamed accordingly please `s/kt_/kangaroo_twelve_/g`.

sjaeckel (changes requested):
I overlooked that at the first review.

Thanks for reverting the first commit, the changes look a lot better now.

doc/crypt.tex
```diff
@@ -3143,6 +3143,21 @@ \subsection{TurboSHAKE}
 The init function \code{turbo\_shake\_init()} is implemented as a macro which calls \code{sha3\_shake\_init()}.
 
 
+\subsection{KangarooTwelve}
+Another variation of SHA3 SHAKE is KangarooTwelve, which has been specified in \href{https://datatracker.ietf.org/doc/rfc9861/}{\texttt{RFC 9861}}.
+
+The API works equivalent to the one of SHA3 SHAKE, where the APIs only have a different name. Additionally, KangarooTwelve supports customization. You can append any or none customization bytes after all input bytes and before squeezing any output digest bytes.
```
sjaeckel, line 3149: Can one also append customization bytes in multiple calls?

I'm just asking.

And is the result of `kt_customization("foobar");` the same as of `kt_customization("foo"); kt_customization("bar");`?

sjaeckel: OK, so the RFC specifies that we **MAY** propose an incremental interface, for either `M` or `C`, so we should also document that somewhere what we support.

I just had a look at the RFC for the first time ... the `D` isn't fixed to `0x1F`, maybe it'd make sense to add support for variable `D` in a future PR. But now we focus on KT.

src/hashes/sha3_test.c
MarekKnapek: Yes, fixed.

src/hashes/sha3_test.c
MarekKnapek: Yes, fixed.

src/hashes/sha3_test.c
MarekKnapek: Removed in favor of base16_decode.

src/hashes/sha3_test.c
MarekKnapek: Yes, fixed.

src/headers/tomcrypt_custom.h
MarekKnapek: Yes, fixed.

src/headers/tomcrypt_custom.h
MarekKnapek: Yes, fixed. What about the function names?

src/hashes/sha3_test.c
```diff
@@ -388,161 +388,256 @@ int sha3_shake_test(void)
...
+   do
+   {
+      count = rem < ((long)(sizeof(digest))) ? rem : ((long)(sizeof(digest)));
+      if ((err = s_turbo_shake_hex_to_bin(&expected_hex[0], &expected_digest_bin[0], count)) != CRYPT_OK) return err;
+      if ((err = turbo_shake_done(&md, &digest[0], count)) != CRYPT_OK) return err;
+      LTC_COMPARE_TESTVECTOR(digest, count, expected_digest_bin, count, "TurboSHAKE", counter);
+      rem -= count;
+      expected_digest_hex += count * 2;
+   }while(rem != 0);
+   return CRYPT_OK;
```
sjaeckel, line 465:
```suggestion
   int err;
   hash_state md;
   unsigned long offset, rem, count;
   unsigned char input[1024];
   unsigned char digest[64];
   char const *expected_hex;
   unsigned char expected_digest_bin[sizeof(digest)];

   LTC_ARGCHK(bits_count == 128 || bits_count == 256);
   LTC_ARGCHK(input_bytes_count >= 0);
   LTC_ARGCHK(skip_digest_bytes >= 0);
   LTC_ARGCHK(digest_bytes_count >= 1);
   LTC_ARGCHK(counter >= 0);
   LTC_ARGCHK(expected_digest_hex && expected_digest_hex[0] != '\0');

   if ((err = turbo_shake_init(&md, bits_count)) != CRYPT_OK) return err;
   offset = 0;
   rem = input_bytes_count;
   do {
      count = rem < sizeof(input) ? rem : sizeof(input);
      if ((err = s_turbo_shake_generate_ptn(input, offset, count)) != CRYPT_OK) return err;
      if ((err = turbo_shake_process(&md, input, count)) != CRYPT_OK) return err;
      offset += count;
      rem -= count;
   } while(rem != 0);
   rem = skip_digest_bytes;
   do {
      count = rem < sizeof(digest) ? rem : sizeof(digest);
      if ((err = turbo_shake_done(&md, digest, count)) != CRYPT_OK) return err;
      rem -= count;
   } while(rem != 0);
   rem = digest_bytes_count;
   expected_hex = expected_digest_hex;
   do {
      count = rem < sizeof(digest) ? rem : sizeof(digest);
      if ((err = s_turbo_shake_hex_to_bin(expected_hex, expected_digest_bin, count)) != CRYPT_OK) return err;
      if ((err = turbo_shake_done(&md, digest, count)) != CRYPT_OK) return err;
      LTC_COMPARE_TESTVECTOR(digest, count, expected_digest_bin, count, "TurboSHAKE", counter);
      rem -= count;
      expected_digest_hex += count * 2;
   } while(rem != 0);
   return CRYPT_OK;
```

Three more things I just saw, which I also missed in the first two reviews 😬 (I should focus more, but it's a lot of changes, so it's hard to do that while also doing my regular work).
1. we usually use the nearest correct type, so for sizes it should be `unsigned long` most of the time. (The correct type would be `size_t`, but that move didn't happen yet...)
2. we try to use the implicit meaning of statements. If you use `&foo[0]` in an API call, where `foo` is an array of something and the API expects a pointer, simply use `foo` there. Keep the usage of `&foo[n]` for the cases where you explicitly index into the array. It makes reading the code easier.
3. Try to mimick the code style used in the recent changes. I know, we should have an auto-formatter, but that's something for the future.

doc/crypt.tex
MarekKnapek: Yes, the customization could be appended multiple times. And yes, about the foobar==foo+bar. You can append unlimited amount of customization bytes, up to unsigned long count, if you want to append more, we should change the variable type holding the number of customization bytes added. To 128 bit number or even larger one. The project https://github.com/kerukuro/digestpp is quite limited in this regard, it holds the entire customization string in memory at once. My implementation holds only the SHAKE state. They also hold a 8 kB buffer, I hold a SHAKE state instead.

src/headers/tomcrypt_custom.h
sjaeckel: > Yes, fixed. What about the function names?

The API should also be renamed accordingly please s/kt_/kangaroo_twelve_/g.

src/hashes/sha3_test.c
MarekKnapek: Yes, fixed.

src/headers/tomcrypt_custom.h
MarekKnapek: Yes, fixed.

src/hashes/sha3_test.c
```diff
@@ -388,161 +388,261 @@ int sha3_shake_test(void)
...
-      0x10, 0x88, 0x2f, 0xde, 0x31, 0xc6, 0xdf, 0x75, 0xbe, 0xb9, 0xd4, 0xcd, 0x93, 0x05, 0xcf, 0xca,
-      0xe3, 0x5e, 0x7b, 0x83, 0xe8, 0xb7, 0xe6, 0xeb, 0x4b, 0x78, 0x60, 0x58, 0x80, 0x11, 0x63, 0x16,
-      0xfe, 0x2c, 0x07, 0x8a, 0x09, 0xb9, 0x4a, 0xd7, 0xb8, 0x21, 0x3c, 0x0a, 0x73, 0x8b, 0x65, 0xc0,
-   };
+#ifdef LTC_KANGAROO_TWELVE
+typedef struct kangaroo_twelve_test_s {
+   int bits_count, is_ptn;
+   unsigned long input_bytes_count, customization_bytes_count, skip_digest_bytes, digest_bytes_count;
+   const char *expected_digest_hex;
+} kangaroo_twelve_test_t;
```
sjaeckel:
```suggestion
typedef struct kangaroo_twelve_test {
   int bits_count, is_ptn;
   unsigned long input_bytes_count, customization_bytes_count, skip_digest_bytes, digest_bytes_count;
   const char *expected_digest_hex;
} kangaroo_twelve_test;
```

If you typedef, always call the `struct` and the `typedef` the same, and please don't suffix if there's no need for it.

We've started to prefix with `ltc_` if they're declared in a header and if they're new infrastructure, e.g. the X.509 implementation that will be merged at one point, but obviously this doesn't apply here.

src/hashes/sha3_test.c
```diff
@@ -388,161 +388,261 @@ int sha3_shake_test(void)
...
+static LTC_INLINE int s_turbo_shake_test_one(const turbo_shake_test_t *testcase, int counter)
+{
+   int err;
+   hash_state md;
+   unsigned long offset;
+   unsigned long rem;
+   unsigned long count;
+   unsigned char input[1024];
+   unsigned char digest[64];
+   char const *expected_hex;
```
sjaeckel:
```suggestion
   const char *expected_hex;
```

src/hashes/sha3_test.c
```diff
@@ -388,161 +388,261 @@ int sha3_shake_test(void)
...
+static LTC_INLINE int s_kangaroo_twelve_test_one(const kangaroo_twelve_test_t *testcase, int counter)
+{
+   int err;
+   hash_state md;
+   unsigned long offset;
+   unsigned long rem;
+   unsigned long count;
+   unsigned char input[1024];
+   unsigned char digest[64];
+   char const *expected_hex;
```
sjaeckel:
```suggestion
   const char *expected_hex;
```

src/headers/tomcrypt_hash.h
```diff
@@ -14,6 +14,18 @@ struct sha3_state {
 };
 #endif
 
+#ifdef LTC_KANGAROO_TWELVE
+struct kangaroo_twelve_state {
+    struct sha3_state outer;
+    struct sha3_state inner;
+    unsigned long blocks_count;
+    unsigned long customization_len;
```
sjaeckel: This is a follow-up of https://github.com/libtom/libtomcrypt/pull/708/files#r2569540830

I guess it would make sense to make this a `ulong64`!?

So it's at least the same size for all architectures.

I doubt this will ever be required to be bigger than 64bit, even if it's technically feasible and also supported by the standard. If you really really believe that we should support more than 64bits I'd suggest to either make it use the MPI provider or ideally make it dynamically sized, since if 64bits isn't enough 128bits also won't be ;)

As of the RFC Ch. 3.3 it should be possible to hold "a non-negative integer x < 256**255", which is a 2040bit number, so this would require 32 64bit numbers.

This should then look as follows

```c
    unsigned char clen_len;
    ulong64 *customization_len;
```

Then grow the array as needed... but this is all very theoretical ... If I look at the [XKCP project], which seems to be kind of the reference implementation, they don't even support incremental updates of the customization IIUC and take a `size_t` ...

[XKCP project]: https://github.com/XKCP/K12

src/hashes/sha3_test.c
```diff
@@ -388,161 +388,261 @@ int sha3_shake_test(void)
...
+   return CRYPT_OK;
+}
+#endif
+
+#ifdef LTC_TURBO_SHAKE
+typedef struct turbo_shake_test_s {
+   int bits_count;
+   unsigned long input_bytes_count, skip_digest_bytes, digest_bytes_count;
+   const char *expected_digest_hex;
+} turbo_shake_test_t;
```
sjaeckel:
```suggestion
typedef struct turbo_shake_test {
   int bits_count;
   unsigned long input_bytes_count, skip_digest_bytes, digest_bytes_count;
   const char *expected_digest_hex;
} turbo_shake_test;
```

same

sjaeckel (commented):
Sorry for those nitpicks and thanks for your patience.

src/hashes/sha3_test.c
MarekKnapek: Can not do because `src/hashes/sha3_test.c:517:3: error: ‘kangaroo_twelve_test’ redeclared as different kind of symbol`, changed to `kangaroo_twelve_test_case` instead.

src/hashes/sha3_test.c
MarekKnapek: Yes, fixed.

src/hashes/sha3_test.c
MarekKnapek: Yes, fixed.

src/headers/tomcrypt_hash.h
MarekKnapek: Changed to u64.

src/hashes/sha3_test.c
MarekKnapek: Again, struct and function could not have the same name.

MarekKnapek:
> Sorry for those nitpicks and thanks for your patience.

No need to apologize, it is your project, that means your rules and standards. I'm not familiar with your rules, by telling me what I did wrong I'm learning them.

sjaeckel:
That was quick! Thanks for the PR and the reactive way of handling the reviews :)

PR #714, opened 2026-04-02
Optimize gcm_gf_mult using PCLMULQDQ and PMULL

sjlombardo:
The current gcm_gf_mult table optimization is great for large payloads, but it is less well-suited for small ones. For example, encrypting the same total volume of data in 4 KB chunks vs. a single 1 MB message is 3-10x slower with the table-optimized `gcm_gf_mult` implementation. This creates a bottleneck for programs that process many small messages with different keys.

Unfortunately disabling LTC_GCM_TABLES introduces the reverse problem. It speeds things up considerably for small messages but degrades performance for larger ones.

Using hardware carry-less multiply solves these problems for both cases. This PR implements multiplication for GCM using Intel PCLMULQDQ on x86/x86_64 and PMULL on ARM AArch64. Benchmarking shows it is consistently 3-10x faster across messages between 1K and 4K. Hardware acceleration scales better under load, performs predictably regardless of payload size, and makes key setup faster. On balance, this eliminates the performance tradeoff. If there is interest, additional benchmark details can be provided.

Hardware support for this approach is extensive. The required instructions for x86/x86_64 have been around for over 10 years. Similarly, AArch64 support in ARMv8+crypto is available in all reasonably modern devices including most phones, desktops, and server hardware.

The core implementation is based on well-understood and referenced papers:

- From [Intel Carry-Less Multiplication Instruction and its Usage for Computing the GCM Mode (Gueron & Kounavis)](https://cdrdv2-public.intel.com/836172/clmul-wp-rev-2-02-2014-04-20.pdf), algorithms 1 and 5 (the implementation follows the paper directly and is attributed in the source code comments)
- From [Implementing GCM on ARMv8 (Gouvea & Lopez)](https://conradoplg.modp.net/files/2010/12/gcm14.pdf), algorithms 3 and 5

In order to minimize changes, the existing `gcm_gf_mult` is renamed to `gcm_gf_mult_sw` internally and a very simple `gcm_gf_mult` dispatcher sits in front of it. It establishes whether there is hardware support and routes to the appropriate implementation or the software fallback. There are no public API changes to LibTomCrypt.

CPU detection follows the same pattern as the recent AES-NI work on `develop` using `CPUID` on x86/x86_64, `getauxval` on ARM Linux, and `sysctlbyname` on ARM Apple.

The new functions use `__attribute__((target(...)))` under GCC and Clang to avoid global `-mpclmul` or `+crypto` compiler flag changes to the makefile(s). Compile-time opt-out is possible with `LTC_NO_GCM_PCLMUL` or `LTC_NO_GCM_PMULL`. Both are disabled under `LTC_NO_ASM`. If neither of the previous macros is set then `LTC_GCM_TABLES` is automatically disabled since the intrinsic path is preferable.

It wasn't necessary to add tests because the existing GCM test suite covers this adequately and there is no API change. The test suite was run and passed on Linux x86_64, Linux AArch64, Windows x86_64, Windows ARM64, macOS x86_64, and macOS AArch64.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -139,6 +139,7 @@ jobs:
         os: [ ubuntu-22.04, ubuntu-24.04, ubuntu-22.04-arm, ubuntu-24.04-arm ]
         build_type: [ '', -DCMAKE_BUILD_TYPE=Debug, -DCMAKE_BUILD_TYPE=Release, -DCMAKE_BUILD_TYPE=RelWithDebInfo, -DCMAKE_BUILD_TYPE=MinSizeRel ]
         cc: [ clang, gcc ]
+        cflags: [ '', '-DLTC_CFLAGS="-DLTC_NO_ASM"']
         config:
           # Static library build
           - { CMAKEOPTIONS: '-DBUILD_SHARED_LIBS=Off' }
@@ -161,12 +162,12 @@ jobs:
         run: |
           mkdir build
           cd build
-          CC=${{ matrix.cc }} cmake ${{ matrix.config.CMAKEOPTIONS }} ${{ matrix.build_type }} ..
+          CC=${{ matrix.cc }} cmake ${{ matrix.config.CMAKEOPTIONS }} ${{ matrix.build_type }} ${{ matrix.cflags }} ..
           make -j$(nproc)
       - name: test
         run: |
           cd build
-          CC=${{ matrix.cc }} cmake ${{ matrix.config.CMAKEOPTIONS }} ${{ matrix.build_type }} -DBUILD_TESTING=On ..
+          CC=${{ matrix.cc }} cmake ${{ matrix.config.CMAKEOPTIONS }} ${{ matrix.build_type }} ${{ matrix.cflags }} -DBUILD_TESTING=On ..
           make -j$(nproc)
           ctest
       - name: error logs
--- a/src/encauth/gcm/gcm_gf_mult.c
+++ b/src/encauth/gcm/gcm_gf_mult.c
@@ -7,6 +7,229 @@
 */
 #include "tomcrypt_private.h"
 
+#if defined(LTC_GCM_MODE) || defined(LTC_LRW_MODE)
+#if defined(LTC_GCM_PCLMUL)
+#if defined(_MSC_VER)
+#include <intrin.h>
+#else
+#include <cpuid.h>
+#endif
+#include <wmmintrin.h>
+#include <smmintrin.h>
+#include <emmintrin.h>
+
+static LTC_INLINE int s_pclmul_is_supported(void)
+{
+   static int initialized = 0, is_supported = 0;
+
+   if (initialized == 0) {
+      /* Test CPUID.1.0.ECX[1]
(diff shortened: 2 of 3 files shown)


sjaeckel (approved):
Thanks a lot for the contribution! Pretty impressive improvements.

I took the liberty to rebase the commit and do the necessary modifications to pass `./helper.pl -a` and added a CI job for CMake.

I ran the timing demo on some variations with the following results on x86_64:

```
$ make timing CFLAGS="-DLTC_NO_GCM_PCLMUL -DLTC_NO_TABLES"
[...]
$ ./timing encmac |& grep -e GCM -e Timing
ENC+MAC Timings (zero byte AAD, 16 byte IV, cycles/byte on 1KB blocks):
GCM (no-precomp)               19
GCM (precomp)                  19
ENC+MAC Timings (zero byte AAD, 16 byte IV, cycles/byte on 4KB blocks):
GCM (no-precomp)               19
GCM (precomp)                  18
ENC+MAC Timings (zero byte AAD, 16 byte IV, cycles/byte on 32KB blocks):
GCM (no-precomp)               18
GCM (precomp)                  18
$ make timing CFLAGS="-DLTC_NO_GCM_PCLMUL"
[...]
$ ./timing encmac |& grep -e GCM -e Timing
ENC+MAC Timings (zero byte AAD, 16 byte IV, cycles/byte on 1KB blocks):
GCM (no-precomp)               96
GCM (precomp)                   7
ENC+MAC Timings (zero byte AAD, 16 byte IV, cycles/byte on 4KB blocks):
GCM (no-precomp)               27
GCM (precomp)                   6
ENC+MAC Timings (zero byte AAD, 16 byte IV, cycles/byte on 32KB blocks):
GCM (no-precomp)                8
GCM (precomp)                   5
$ make timing
[...]
$ ./timing encmac |& grep -e GCM -e Timing
ENC+MAC Timings (zero byte AAD, 16 byte IV, cycles/byte on 1KB blocks):
GCM (no-precomp)                6
GCM (precomp)                   5
ENC+MAC Timings (zero byte AAD, 16 byte IV, cycles/byte on 4KB blocks):
GCM (no-precomp)                5
GCM (precomp)                   5
ENC+MAC Timings (zero byte AAD, 16 byte IV, cycles/byte on 32KB blocks):
GCM (no-precomp)                5
GCM (precomp)                   5
```

sjaeckel:
While looking into the failing build logs on ARM of #717 I saw the following warning we might have ignored previously:

https://github.com/libtom/libtomcrypt/actions/runs/24385662530/job/71218807533?pr=717#step:4:83

```
In file included from /home/runner/work/libtomcrypt/libtomcrypt/src/encauth/gcm/gcm_gf_mult.c:124:
/usr/lib/llvm-18/lib/clang/18/include/arm_neon.h:32317:21: warning: cast from 'bfloat16_t *' (aka '__bf16 *') to 'int32_t *' (aka 'int *') increases required alignment from 2 to 4 [-Wcast-align]
 32317 | int32_t __reint1 = *(int32_t *) &__reint << 16;
       |                     ^~~~~~~~~~~~~~~~~~~~
```

PR #717, opened 2026-04-10
SHA-1, SHA-224 & SHA-256 with x86 instructions

### LINKED ISSUE #716: SHA-1 x86
Hi @sjaeckel , not sure where to open a discussion. Issue or pull request? I implemented SHA-1 accelerated by x86 intrinsics.

But, for that I would need some support from LibTomCrypt architecture:

 - How to detect at compile-time what processor architecture are we compiling for? Is it x86 or x64 or some other one?
 - How to detect compiler used? Is it GCC? Or clang? Or MSVC? Or something else which I do not know how to write intrinsics for?
 - How to detect SHA-1 (and SSE2 and SSSE3 and SSE4.1) instruction set presence at run-time? Use the `cpuid` instruction and cache its result. But where? Is there a precedent already in the lib?
 - How to detect the ability to use `alignas` or `pragma pack(n)` or something to force variable alignment to 16 bytes?
 - How to add a new file into all those make, cmake and other build systems? Because I have no idea.

My idea is to have two implementations of SHA-1 ("sha-1-portable" and "sha-1-x86") in parallel at once in the lib (if the user wants to). Then have third, generic "sha-1" one, which "just" dispatches its logic into one of the previous ones, depending on CPUID results.

My code is at: https://github.com/MarekKnapek/libtomcrypt/commits/SHA-1x86/.

MarekKnapek:
WIP - there are still things that need to be resolved, see https://github.com/libtom/libtomcrypt/issues/716.

### DIFF
--- a/src/hashes/sha1_desc.c
+++ b/src/hashes/sha1_desc.c
@@ -1,0 +1,175 @@
+/* LibTomCrypt, modular cryptographic library -- Tom St Denis */
+/* SPDX-License-Identifier: Unlicense */
+#include "tomcrypt_private.h"
+
+#ifdef LTC_SHA1
+
+const struct ltc_hash_descriptor sha1_desc =
+{
+    "sha1",
+    2,
+    20,
+    64,
+
+    /* OID */
+   { 1, 3, 14, 3, 2, 26,  },
+   6,
+
+    &sha1_init,
+    &sha1_process,
+    &sha1_done,
+    &sha1_test,
+    NULL
+};
+
+#if defined LTC_SHA1_X86
+
+#if !defined (LTC_S_X86_CPUID)
+#define LTC_S_X86_CPUID
+static LTC_INLINE void s_x86_cpuid(int* regs, int leaf)
+{
+#if defined _MSC_VER
+   __cpuid(regs, leaf);
+#else
+    int a, b, c, d;
+
+    a = leaf;
+    b = c = d = 0;
+    asm volatile ("cpuid"
+        :"=a"(a), "=b"(b), "=c"(c), "=d"(d)
+        :"a"(a), "c"(c)
+    );
+    regs[0] = a;
+    regs[1] = b;
+    regs[2] = c;
+    regs[3] = d;
+#endif
+}
+#endif /* LTC_S_X86_CPUID */
+
+static LTC_INLINE int s_sha1_x86_is_supported(void)
+{
+    static int initialized = 0;
+    static int is_supported = 0;
+
+    if (initialized == 0) {
+        int regs[4];
+        int sse2, ssse3, sse41, sha;
+        /* Leaf 0, Reg 0 contains the number of leafs available */
+        s_x86_cpuid(regs, 0);
+        if(regs[0] >= 7) {
+           s_x86_cpuid(regs, 1);
+           sse2  = ((((unsigned int)(regs[3])) >> 26) & 1u) != 0; /* SSE2,   leaf 1, edx, bit 26 */
+           ssse3 = ((((unsigned int)(regs[2])) >>  9) & 1u) != 0; /* SSSE3,  leaf 1, ecx, bit  9 */
+           sse41 = ((((unsigned int)(regs[2])) >> 19) & 1u) != 0; /* SSE4.1, leaf 1, ecx, bit 19 */
+           s_x86_cpuid(regs, 7);
+           sha = ((((unsigned int)(regs[1])) >> 29) & 1u) != 0; /* SHA, leaf 7, ebx, bit 29 */
+           is_supported = sse2 && ssse3 && sse41 && sha;
+        }
+        initialized = 1;
+    }
+    return is_supported;
+}
+#endif /* LTC_SHA1_X86 */
+
+/**
+   Initialize the hash state
+   @param md   The hash state you wish to initialize
+   @return CRYPT_OK if successful
+*/
+int sha1_init(hash_state * md)
+{
+#if defined LTC_SHA1_X86
+    if(s_sha1_x86_is_supported()) {
+        return sha1_x86_init(md);
+    }
+#endif
+    return sha1_c_init(md);
+}
+
+/**
+   Process a block of memory though the hash
+   @param md     The hash state
+   @param in     The data to hash
+   @param inlen  The length of the data (octets)
+   @return CRYPT_OK if successful
+*/
+int sha1_process(hash_state * md, const unsigned char *in, unsigned long inlen)
+{
+#if defined LTC_SHA1_X86
+    if(s_sha1_x86_is_supported()) {
+        return sha1_x86_process(md, in, inlen);
+    }
+#endif
+    return sha1_c_process(md, in, inlen);
+}
+
+/**
+   Terminate the hash to get the digest
+   @param md  The hash state
+   @param out [out] The destination of the hash (20 bytes)
+   @return CRYPT_OK if successful
+*/
+int sha1_done(hash_state * md, unsigned char *out)
+{
+#if defined LTC_SHA1_X86
+    if(s_sha1_x86_is_supported()) {
+        return sha1_x86_done(md, out);
+    }
+#endif
+    return sha1_c_done(md, out);
+}
+
+/**
+  Self-test the hash
+  @return CRYPT_OK if successful, CRYPT_NOP if self-tests have been disabled
+*/
+int sha1_test(void)
+{
+   return sha1_test_desc(&sha1_desc, "SHA1");
+}
+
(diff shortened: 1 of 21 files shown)


src/hashes/sha1_desc.c
```diff
@@ -0,0 +1,182 @@
...
+/**
+  Self-test the hash
+  @return CRYPT_OK if successful, CRYPT_NOP if self-tests have been disabled
+*/
+int sha1_test(void)
+{
+   return sha1_test_desc(&sha1_desc, "SHA1");
+}
+
+int sha1_test_desc(const struct ltc_hash_descriptor *desc, const char *name)
```
sjaeckel, line 132: @karel-m could you please have a look at `helper.pl`, this now always complains about the following, and I'm not sure how to fix this best.

```
*desc, missing in ./src/misc/crypt/crypt_register_all_hashes.c
*desc, missing in ./tests/test.c
```

src/hashes/sha2/sha256_x86.c
```diff
@@ -0,0 +1,362 @@
...
+    _mm_store_si128(((__m128i*)(&md->sha256_x86.state[4])), state_1);
+    return CRYPT_OK;
+}
+
+#ifdef LTC_CLEAN_STACK
+static int s_sha256_compress(hash_state * md, const unsigned char *buf)
+{
+    int err;
+    err = ss_sha256_compress(md, buf);
+    burn_stack(sizeof(ulong32) * 74);
```
MarekKnapek: Todo: Measure stack usage.

src/hashes/sha1_desc.c
karel-m:
```patch
   my @descriptors;
   find({ wanted => sub { push @src, $_ if $_ =~ /\.c$/ }, no_chdir=>1 }, "./src/${which}/");
   for my $f (@src) {
-    my @n = map { my $x = $_; $x =~ s/^.*?ltc_${what}_descriptor\s+(\S+).*$/$1/; $x } grep { $_ =~ /ltc_${what}_descriptor/ } split /\n/, read_file($f);
+    my @n = map { my $x = $_; $x =~ s/^.*?ltc_${what}_descriptor\s+(\S+).*$/$1/; $x } grep { $_ =~ /^[^()]*ltc_${what}_descriptor/ } split /\n/, read_file($f);
     push @descriptors, @n if @n;
   }
```

but still returns:
```
check-source:    PASS
check-defines:   PASS
check-ciphers:   PASS
sha256_c_desc missing in ./src/misc/crypt/crypt_register_all_hashes.c
sha224_c_desc missing in ./src/misc/crypt/crypt_register_all_hashes.c
sha256_c_desc missing in ./tests/test.c
sha224_c_desc missing in ./tests/test.c
check-hashes:    FAIL 4
check-prngs:     PASS
```

sjaeckel:
FYI: Please feel free to read the commit messages of the commits I added, they contain the reasoning behind most of the changes.

sjaeckel:
Looking at the results in appveyor it seems like this somehow doesn't result in any improvements on MSVC ...

sjaeckel:
> Looking at the results in appveyor it seems like this somehow doesn't result in any improvements on MSVC ...

So that's a little surprising, but it seems the CPUs of Appveyor don't support the SHA instructions!? https://ci.appveyor.com/project/libtom/libtomcrypt/builds/53881218

https://github.com/libtom/libtomcrypt/blob/c1752846d5456d6e7808635b00cc38c6737d2ae0/src/hashes/sha1_desc.c#L68

```
CPU says: 1 && 1 && 1 && 0
```

I'd suggest we simply ignore this and someone using Windows will maybe pop up if something else is broken. I'll rebase and squash this all together, OK @MarekKnapek ?

MarekKnapek:
- Looks great. Much neater than my attempts!
 - Not sure why you removed the `LTC_ARGCHK` sanity checks in 6a9e083a6e67d27a6e14a2f49cfcb842b8e69ed7 and in 1b68c950d072a628a74a1898199fea3e0f3276e9.
 - Regarding MSVC, I prepared a Visual Studio 2026 project files. They are using the MSVC compiler and also the clang-cl compiler. They are building with warning level 3 (MSVC) and warning level4 (clang-cl). And x86 dispatching and execution works fine. (I do have SHA-capable Windows machine and SHA-not-capable Linux machine.) Edit: They are at https://github.com/MarekKnapek/libtomcrypt/commits/msvs/ and at https://github.com/MarekKnapek/libtomcrypt/commits/msvs-sha-x86/

sjaeckel:
> * Not sure why you removed the `LTC_ARGCHK` sanity checks in [6a9e083](https://github.com/libtom/libtomcrypt/commit/6a9e083a6e67d27a6e14a2f49cfcb842b8e69ed7) and in [1b68c95](https://github.com/libtom/libtomcrypt/commit/1b68c950d072a628a74a1898199fea3e0f3276e9).

Because
1. they were in the compress functions, so they would get executed each round, but they need only to be tested once.
2. they are redundant
```
    LTC_ARGCHK(md != NULL);
    LTC_ARGCHK(buf != NULL);
``` 
are already verified here
https://github.com/libtom/libtomcrypt/blob/68aae28a9d67d5ca9c1105b0087096e50d6c85a4/src/headers/tomcrypt_private.h#L150-L151

```
    LTC_ARGCHK(((ltc_uintptr)(&md->sha{1,256}.state[0])) % 16 == 0);
```
is aligned by us in the `init()` function, so doesn't need checking again

```
    LTC_ARGCHK(((ltc_uintptr)(&K[0])) % 16 == 0);
```
is aligned by the compiler, c.f. its declaration

```
    LTC_ARGCHK(sizeof(int) == 4);
```
I'm not sure why that's even checked, and if it were required it should be a compile-time check via `LTC_STATIC_ASSERT(sizeof(int) == 4);` in some other place.

> * Regarding MSVC, I prepared a Visual Studio 2026 project files.

We provide VS2008 project files which should be sufficient to be updated to a newer VS version, or not? Also you should most likely simply use the CMake file instead of a custom VS solution.

MarekKnapek:
.... aaand it is building on ci https://github.com/MarekKnapek/libtomcrypt/actions/runs/24456588598

sjaeckel:
> .... aaand it is building on ci https://github.com/MarekKnapek/libtomcrypt/actions/runs/24456588598

Could you please check whether you can simply build and run the CMake project on a Windows GitHub runner?

MarekKnapek:
Yes, I can run CMake -> generate Visual Studio 2026 project. But the build fails because `libtommath` and because the `s_mp_rand_platform` function. I think it is illegal to use code such as `if(false){ non_existing_function(); }` and relying on the optimizer to remove never used functions - but there is nothing to remove, the compilation fails before the optimizer realizes nobody calls the functions. It should be an `#if` instead of an `if`. https://github.com/MarekKnapek/libtomcrypt/actions/runs/24461800098/job/71477784630#step:3:16613 and https://github.com/MarekKnapek/libtomcrypt/actions/runs/24461800098/job/71477784630#step:3:33223 If I fix / comment out the problematic function calls, everything builds just fine.

sjaeckel:
> Yes, I can run CMake -> generate Visual Studio 2026 project. But the build fails because `libtommath` and because the `s_mp_rand_platform` function. I think it is illegal to use code such as `if(false){ non_existing_function(); }` and relying on the optimizer to remove never used functions - but there is nothing to remove, the compilation fails before the optimizer realizes nobody calls the functions. It should be an `#if` instead of an `if`. https://github.com/MarekKnapek/libtomcrypt/actions/runs/24461800098/job/71477784630#step:3:16613 and https://github.com/MarekKnapek/libtomcrypt/actions/runs/24461800098/job/71477784630#step:3:33223 If I fix / comment out the problematic function calls, everything builds just fine.

Did you read the comment in `s_mp_rand_platform`?

sjaeckel (approved):

MarekKnapek:
Yes, I read it. And I explained to you here few moments ago why I think you are wrong there.

> We intentionally don't fix this issue in order to have a single point of failure for misconfigured compilers.

This is nonsense, according to me. This is not misconfigured compiler, this is bending the rules of the C language and relying on it.

Similarly as `pragma align(x)` or `__decspec(align(x))` or `__attribute__(aligned(x))` are hacks and proper way of doing things is `alignas(x)`. But that is not available in ancient C standard. Only in recent standard.

I would suggest to convert those `if`s from run-time ones (even if evaluating constants, still run-time) into compile-time ones instead. Via using the `#if` preprocessor.

sjaeckel:
your-opinion

sjaeckel:
> Looks great. Much neater than my attempts!

Thank you.

And thank **you** for the contribution and most of the implementation and integration work!

PR #719, opened 2026-04-11
The scrypt Password-Based Key Derivation Function (RFC 7914)

karel-m:
Implementation of scrypt (in the end, quite a short piece of code, as I was able to reuse existing parts of libtomcrypt).

Follows RFC 7914 https://www.rfc-editor.org/rfc/rfc7914.html

### Checklist


karel-m:
Sorry, my pr/scrypt-rfc7914 branch was not based on develop, but on pr/argon2-rfc9106.

I will leave it as it is for now so that we have CI build/test results, but it must be rebased before merging.

karel-m:
Not sure whether it would not be better to name the function `scrypt_pbkdf(..)` instead of just `scrypt(..)`

sjaeckel:
> Not sure whether it would not be better to name the function `scrypt_pbkdf(..)` instead of just `scrypt(..)`

I like that, I think this should be renamed.

src/misc/scrypt/scrypt.c
```diff
@@ -0,0 +1,190 @@
...
+   for (i = 0; i < r; ++i) {
+      XMEMCPY(B + i * 64, Y + (2 * i) * 64, 64);
+   }
+   for (i = 0; i < r; ++i) {
+      XMEMCPY(B + (i + r) * 64, Y + (2 * i + 1) * 64, 64);
+   }
+}
+
+/* Integerify: interpret last 64-byte block as little-endian and return low 64 bits */
+static ulong64 s_integerify(const unsigned char *B, unsigned long r)
```
sjaeckel: I just realized, should those small helpers maybe be declared as `LTC_INLINE`? The same applies to argon2, but I only realized that now, after merging it already ...

sjaeckel (approved):
Thanks a lot!

Basically LGTM, besides that one question, but I'd also be happy to take this as is :)

Please also squash all those commits into one.

sjaeckel (approved):

PR #721, opened 2026-04-14
Fix s_pclmul_is_supported() cache flag on MSVC

cyanogilvie:
Move initialized=1 out of the #else branch so that on MSVC it doesn't call __cpuid every time.


sjaeckel (approved):
Thanks for reporting this! LGTM

PR #726, opened 2026-04-15
Cleanup and fixes

sjaeckel:


karel-m:
@sjaeckel could you please add this small nitpick to this patch? (not worth a separate PR and burning the energy of running the whole test suite)
```
diff --git a/tests/ecc_test.c b/tests/ecc_test.c
index f2c5ee1b..7d0bfb97 100644
--- a/tests/ecc_test.c
+++ b/tests/ecc_test.c
@@ -994,9 +994,7 @@ static int s_ecc_rfc6979(void)
                   }
                  }
                 },
-                {
-                 NULL
-                }
+                {0}
    };

    const void *inputs[] = {
```

It fixes this clang warning:
```
   * clang tests/multi_test.o
tests/ecc_test.c:999:17: warning: missing field 'x' initializer [-Wmissing-field-initializers]
  999 |                 }
      |                 ^
```

It seems that `{NULL}` only sets the first field to NULL; `{0}` is the idiomatic way to zero-initialise the whole object.

sjaeckel:
> @sjaeckel could you please add this small nitpick to this patch?

Uhm, that change is already in this PR in the `Replace int* recid hack in ECC.` commit!? Or did I misunderstand your comment?

https://github.com/libtom/libtomcrypt/blob/5f18a34dc1607a6053003c82bdf3ee024b13cd84/tests/ecc_test.c#L997-L999

karel-m:
Oh, sorry, I did not check it.

In that case, my vote to merge this into develop :)

PR #730, opened 2026-04-16
BLAKE3

karel-m:
Implemented BLAKE3 (hash function) support. Everything (code, headers, tests, docs) is implemented exactly like all other hash algorithms in libtomcrypt.

The code is adapted from a public domain implementation: https://github.com/michaelforney/blake3-tiny (I preferred simplicity over performance; volunteers with a more performant implementation are welcome).

Test vectors are taken from the official BLAKE3 repository: https://github.com/BLAKE3-team/BLAKE3/blob/master/test_vectors/test_vectors.json

The only issue is that BLAKE3 is not currently standardised, and I usually tend to reject similar PRs with non-standardised algorithms :)

### Checklist


sjaeckel (approved):
> The only issue is that BLAKE3 is not currently standardised, and I usually tend to reject similar PRs with non-standardised algorithms :)

Kind of true, but then it's also widely used and specified at the C2SP https://c2sp.org/BLAKE3@v1.0.0

PR #731, opened 2026-04-17
SHA-512 accelerated by x86

MarekKnapek:
SHA-512 accelerated by x86.

### DIFF
--- a/src/hashes/sha2/sha384_desc.c
+++ b/src/hashes/sha2/sha384_desc.c
@@ -1,0 +1,191 @@
+/* LibTomCrypt, modular cryptographic library -- Tom St Denis */
+/* SPDX-License-Identifier: Unlicense */
+/**
+   @param sha384.c
+   LTC_SHA384 hash included in sha512.c, Tom St Denis
+*/
+
+#include "tomcrypt_private.h"
+
+#if defined(LTC_SHA384) && defined(LTC_SHA512)
+
+const struct ltc_hash_descriptor sha384_desc =
+{
+    "sha384",
+    4,
+    48,
+    128,
+
+    /* OID */
+   { 2, 16, 840, 1, 101, 3, 4, 2, 2,  },
+   9,
+
+    &sha384_init,
+    &sha512_process,
+    &sha384_done,
+    &sha384_test,
+    NULL
+};
+
+#if defined LTC_SHA384_X86
+
+#if !defined (LTC_S_X86_CPUID)
+#define LTC_S_X86_CPUID
+#if defined _MSC_VER
+#include <intrin.h>
+#endif
+static LTC_INLINE void s_x86_cpuid(int* regs, int leaf)
+{
+#if defined _MSC_VER
+   __cpuid(regs, leaf);
+#else
+    int a, b, c, d;
+
+    a = leaf;
+    b = c = d = 0;
+    asm volatile ("cpuid"
+        :"=a"(a), "=b"(b), "=c"(c), "=d"(d)
+        :"a"(a), "c"(c)
+    );
+    regs[0] = a;
+    regs[1] = b;
+    regs[2] = c;
+    regs[3] = d;
+#endif
+}
+static LTC_INLINE void s_x86_cpuidex(int* regs, int eax, int ecx)
+{
+#if defined _MSC_VER
+   __cpuidex(regs, eax, ecx);
+#else
+    int a, b, c, d;
+
+    a = eax;
+    c = ecx;
+    b = d = 0;
+    asm volatile ("cpuid"
+        :"=a"(a), "=b"(b), "=c"(c), "=d"(d)
+        :"0"(a), "2"(c)
+    );
+    regs[0] = a;
+    regs[1] = b;
+    regs[2] = c;
+    regs[3] = d;
+#endif
+}
+#endif /* LTC_S_X86_CPUID */
+
+static LTC_INLINE int s_sha384_x86_is_supported(void)
+{
+    static int initialized = 0;
+    static int is_supported = 0;
+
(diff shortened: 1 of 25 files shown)


MarekKnapek:
I have no idea what is the problem on ci with the amalgamed build. How do I run this myself?

src/hashes/sha2/sha384_desc.c
```diff
@@ -0,0 +1,197 @@
...
+*/
+int sha384_test(void)
+{
+   return sha384_test_desc(&sha384_desc, "SHA384");
+}
+
+int sha384_test_desc(const struct ltc_hash_descriptor *desc, const char *name)
+{
+ #ifndef LTC_TEST
+    return CRYPT_NOP;
```
karel-m, line 159: before this line there should be:
```
LTC_UNUSED_PARAM(desc);
LTC_UNUSED_PARAM(name);
```

src/hashes/sha2/sha512_224_desc.c
```diff
@@ -0,0 +1,193 @@
...
+*/
+int  sha512_224_test(void)
+{
+   return sha512_224_test_desc(&sha512_224_desc, "SHA512-224");
+}
+
+int sha512_224_test_desc(const struct ltc_hash_descriptor *desc, const char *name)
+{
+ #ifndef LTC_TEST
+    return CRYPT_NOP;
```
karel-m, line 159: before this line there should be:
```
LTC_UNUSED_PARAM(desc);
LTC_UNUSED_PARAM(name);
```

src/hashes/sha2/sha512_256_desc.c
```diff
@@ -0,0 +1,193 @@
...
+*/
+int  sha512_256_test(void)
+{
+   return sha512_256_test_desc(&sha512_256_desc, "SHA512-256");
+}
+
+int sha512_256_test_desc(const struct ltc_hash_descriptor *desc, const char *name)
+{
+ #ifndef LTC_TEST
+    return CRYPT_NOP;
```
karel-m, line 159: before this line there should be:
```
LTC_UNUSED_PARAM(desc);
LTC_UNUSED_PARAM(name);
```

src/hashes/sha2/sha512_desc.c
```diff
@@ -0,0 +1,218 @@
...
+*/
+int sha512_test(void)
+{
+   return sha512_test_desc(&sha512_desc, "SHA512");
+}
+
+int sha512_test_desc(const struct ltc_hash_descriptor *desc, const char *name)
+{
+ #ifndef LTC_TEST
+    return CRYPT_NOP;
```
karel-m, line 187: before this line there should be:
```
LTC_UNUSED_PARAM(desc);
LTC_UNUSED_PARAM(name);
```

src/hashes/sha2/sha512_224_desc.c
```diff
@@ -0,0 +1,195 @@
...
+   9,
+
+    &sha512_224_init,
+    &sha512_process,
+    &sha512_224_done,
+    &sha512_224_test,
+    NULL
+};
+
+#if defined LTC_SHA224_X86
```
karel-m: or `#if defined LTC_SHA512_224_X86`?

src/hashes/sha2/sha512_256_desc.c
```diff
@@ -0,0 +1,195 @@
...
+   9,
+
+    &sha512_256_init,
+    &sha512_process,
+    &sha512_256_done,
+    &sha512_256_test,
+    NULL
+};
+
+#if defined LTC_SHA256_X86
```
karel-m: or `#if defined LTC_SHA512_256_X86` ?

src/hashes/sha2/sha512_x86.c
```diff
@@ -0,0 +1,465 @@
...
+    &sha512_x86_init,
+    &sha512_x86_process,
+    &sha512_x86_done,
+    &sha512_x86_test,
+    NULL
+};
+
+/* the K array */
+#define K sha512_x86_k
+static const ulong64 sha512_x86_k[80] LTC_ALIGN(32) = {
```
sjaeckel:
```suggestion
#define K sha512_x86_k
LTC_ALIGN_MSVC(32)
static const ulong64 K[80] LTC_ALIGN(32) = {
```

Please note that's two changes.

Regarding the `LTC_ALIGN_MSVC()` macro addition: or isn't that required because this feature is only enabled for newer MSVC versions, which already have support for `attribute((aligned()))`?

sjaeckel (changes requested):
Basically LGTM, besides this one comment for now. I didn't look in further detail yet, but will do that tomorrow.

By merging #726 there was a merge conflict, which I've just resolved while rebasing this branch and therefore I've force-pushed.

src/hashes/sha2/sha512_x86.c
MarekKnapek: I don't think newer MSVC versions have support for GCC-style `__attribute__`, they do have support for MSVC-style `__declspec` and C-standard style `alignas`.

I have put the same alignment specifiers into the generic portable version of the SHA-512 table. In hopes that the compiler will recognize the two constants are exactly the same and de-duplicate them.

sjaeckel (changes requested):
That last commit doesn't belong in this PR and requires further checks before applying it.

We started using inline functions declared in header files some time ago, but then @karel-m had some platforms where this doesn't work so we reverted back to not using them. We'd have to check again today. Please remove that commit and we'll do that in a separate PR.

karel-m:
Since this PR was started many things have changed on the develop branch.

So I have opened a new #787 based on the sha512_x86.c code from this PR.

Issue #738, opened 2026-04-23
sha1_x86.c related failures with old compilers (Linux, FreeBSD)


karel-m:
FreeBSD 9.2 / amd64 / gcc 4.2.1 20070831 patched [FreeBSD]

https://www.cpantesters.org/cpan/report/5b495d5a-3f2b-11f1-bba7-a1476e8775ea
```
cc -Iltm -Iltc/headers -DLTC_SOURCE -DLTC_NO_TEST -DLTC_NO_PROTOTYPES -DLTM_DESC -DHAS_FPSETMASK -DHAS_FLOATINGPOINT_H -fno-strict-aliasing -pipe -I/usr/local/include -DPIC -fPIC -O  -DLTC_NO_AES_NI -DLTC_NO_GCM_PCLMUL -DARGTYPE=4 -c ltc/hashes/sha1_x86.c -o ltc/hashes/sha1_x86.o
ltc/hashes/sha1_x86.c:14: warning: expected [error|warning|ignored] after '#pragma GCC diagnostic'
In file included from ltc/hashes/sha1_x86.c:22:
/usr/include/gcc/4.2/tmmintrin.h:34:3: error: #error "SSSE3 instruction set not enabled"
ltc/hashes/sha1_x86.c:23:54: error: smmintrin.h: No such file or directory
ltc/hashes/sha1_x86.c:24:111: error: immintrin.h: No such file or directory
ltc/hashes/sha1_x86.c:27: warning: expected [error|warning|ignored] after '#pragma GCC diagnostic'
ltc/hashes/sha1_x86.c:53: warning: '__target__' attribute directive ignored
ltc/hashes/sha1_x86.c: In function 's_sha1_x86_compress':
ltc/hashes/sha1_x86.c:75: error: incompatible types in assignment
ltc/hashes/sha1_x86.c:77: error: incompatible types in assignment
ltc/hashes/sha1_x86.c:79: error: incompatible types in assignment
```

FreeBSD 10.3 / amd64 / gcc 4.2.1 Compatible FreeBSD Clang 3.4.1 (tags/RELEASE_34/dot1-final 208032)

https://www.cpantesters.org/cpan/report/55835eb6-3f2b-11f1-bba7-a1476e8775ea
```
[...]

PR #740, opened 2026-04-24
Fix for issue #738.

MarekKnapek:
https://github.com/libtom/libtomcrypt/issues/738

### Checklist

### DIFF
--- a/src/headers/tomcrypt_cfg.h
+++ b/src/headers/tomcrypt_cfg.h
@@ -324,14 +324,18 @@ typedef unsigned long ltc_mp_digit;
       #define LTC_GCM_PCLMUL
       #undef LTC_GCM_TABLES
    #endif
-   #if !defined(LTC_NO_SHA1_X86)
-      #define LTC_SHA1_X86
-   #endif
-   #if !defined(LTC_NO_SHA224_X86)
-      #define LTC_SHA224_X86
-   #endif
-   #if !defined(LTC_NO_SHA256_X86)
-      #define LTC_SHA256_X86
+   #if (defined __GNUC__ && (__GNUC__ * 100 + __GNUC_MINOR__ >= 409)) || \
+       (defined __clang__ && (__clang_major__ * 100 + __clang_minor__ >= 308)) || \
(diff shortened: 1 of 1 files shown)


src/headers/tomcrypt_cfg.h
```diff
@@ -324,14 +324,19 @@ typedef unsigned long ltc_mp_digit;
...
-   #endif
-   #if !defined(LTC_NO_SHA224_X86)
-      #define LTC_SHA224_X86
-   #endif
-   #if !defined(LTC_NO_SHA256_X86)
-      #define LTC_SHA256_X86
+   #if \
+      (defined __GNUC__ && defined __GNUC_MINOR__ && (((__GNUC__) > 4) || ((__GNUC__) == 4 && (__GNUC_MINOR__) >= 9))) || \
+      (defined __clang__ && defined __clang_major__ && defined __clang_minor__ & (((__clang_major__) > 3) || ((__clang_major__) == 3 && (__clang_minor__) >= 8))) || \
+      (defined _MSC_VER && defined _MSC_FULL_VER && (_MSC_VER) >= 1900)
```
sjaeckel: Let's assume some preconditions to make this simpler 
* `__GNUC_MINOR__` is defined since nearly forever (predef mentions 2.7 having it), so we don't have to check that.
* `__clang_{major,minor}__` are also defined since forever

and also use the established pattern to compare against `(maj * 100 + min)`
```suggestion
   #if (defined __GNUC__ && (__GNUC__ * 100 + __GNUC_MINOR__ >= 409)) || \
       (defined __clang__ && (__clang_major__ * 100 + __clang_minor__ >= 308)) || \
       (defined _MSC_VER && defined _MSC_FULL_VER && (_MSC_VER) >= 1900)
```

sjaeckel (changes requested):
Thanks for the godbolting and finding out which are the minimum versions!

Regarding the link in the comment of the Git commit: https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests

sjaeckel (approved):

PR #751, opened 2026-04-30
More fixes and improvements

sjaeckel:
### Checklist

* [x] if this fixes something: added a `Fixes:` tag to the commit message


src/hashes/sha3.c
```diff
@@ -263,45 +263,51 @@ int sha3_512_init(hash_state *md)
...
+   LTC_ARGCHK(sha3 != NULL);
    if (num != 128 && num != 256) return CRYPT_INVALID_ARG;
-   XMEMSET(&md->sha3, 0, sizeof(md->sha3));
-   md->sha3.capacity_words = (unsigned short)(2 * num / (8 * sizeof(ulong64)));
+   XMEMSET(sha3, 0, sizeof(*sha3));
+   sha3->capacity_words = (unsigned short)(2 * num / (8 * sizeof(ulong64)));
    return CRYPT_OK;
 }
+
+int sha3_shake_init(hash_state *md, int num)
```
karel-m, line 275: what about also adding:
```
int sha3_shake128_init(hash_state *md)
{
   return s_sha3_shake_init(&md->sha3, 16);
}

int sha3_shake256_init(hash_state *md)
{
   return s_sha3_shake_init(&md->sha3, 32);
}
```

so that we can have:
```
const struct ltc_hash_descriptor shake128_desc = {..}

const struct ltc_hash_descriptor shake256_desc = {..}
```

which would be handy for RFC 8702: RSA-PSS-SHAKE128/256 and ECDSA-SHAKE128/256

karel-m:
I have added two small commits. Otherwise, this PR looks good.

The only exception is the `LTC_NO_STRUCT_PADDING part`, which I would suggest moving into a separate PR and perhars discuss pros/cons there. I have doubts whether it will work as expected, especially the change from `ulong64 dummy[8]` to `void *padding[8]`. In my opinion, this goes in the wrong direction as the reserved area becomes platform-dependent: 64 bytes vs. 32 bytes, depending on the pointer size.

karel-m:
The commit related to replacing `LTC_FAST_TYPE_PTR_CAST` might be seen as slightly controversial. If you find it inappropriate here feel free to revert it and/or move it to a separate PR.

karel-m:
@sjaeckel that is all from me for this weekend

PR #754, opened 2026-05-03
RFC 8702: RSA-PSS-SHAKE128/256 and ECDSA-SHAKE128/256

karel-m:
Implemented RFC 8702 support for:
- RSA-PSS-SHAKE128 / RSA-PSS-SHAKE256
- ECDSA-SHAKE128 / ECDSA-SHAKE256

At first this looked like a small change, but the RSA-PSS part turned out to be less clean than expected.

Added SHAKE128/256 support through:
- `sha3_shake128_init` / `sha3_shake128_process` / `sha3_shake128_done`
- `sha3_shake256_init` / `sha3_shake256_process` / `sha3_shake256_done`

Also added the corresponding descriptors: `shake128_desc`, `shake256_desc` (these descriptors have fixed digest lengths set according to RFC 8702 - shake128 outputs 32 bytes, shake256 outputs 64 bytes).

The ECDSA-SHAKE128/256 implementation was fairly straightforward.

The RSA-PSS-SHAKE128/256 implementation required a few workarounds because SHAKE is also used directly for MGF [not like `MGF1(shakeNNN)`]. Therefore I added a new `direct_shake_mgf` flag to `struct ltc_rsa_parameters` and a few IFs in the code.

The result works but the implementation is not as clean as I would have liked.

### Checklist


src/headers/tomcrypt_pk.h
```diff
@@ -72,6 +72,8 @@ typedef struct ltc_rsa_parameters {
    int hash_idx;
    /** MGF1 hash algorithm index, -1 if unset */
    int mgf1_hash_idx;
+   /** Direct SHAKE MGF strength for RFC 8702 PSS-SHAKE: 0, 128 or 256 */
+   int direct_shake_mgf;
```
sjaeckel: I don't like the fact that it's now ambiguous and one could also do `mgf1_hash_idx = find_hash("shakeXXX");` and the result of the operation would be different. I believe it'd be better if this were resolved internally, even if it makes the implementation more complicated than via this approach.

sjaeckel (commented):
Thanks a lot! Bascially LGTM, besides that one remark.

src/headers/tomcrypt_pk.h
karel-m: I agree that the whole logic around mgf1_hash_idx and direct_shake_mgf is hard to follow and full of unexpected surprises.

I expected this PR to be a simple/easy addition. As I mentioned, I am not satisfied with the result either.

karel-m:
**UPDATE:** removed `direct_shake_mgf`, all "magic" now takes place inside `ltc_pkcs_1_mgf1(..)` - which is now slightly misleadingly named as it is not always MGF1 (but it is only an internal function, not part of the public API)

src/pk/pkcs1/pkcs_1_mgf1.c
```diff
@@ -36,6 +39,15 @@ int ltc_pkcs_1_mgf1(int                  hash_idx,
       return err;
    }
 
+#ifdef LTC_SHA3
+   if (XSTRCMP(hash_descriptor[hash_idx].name, "shake128") == 0) {
+      return sha3_shake_memory(128, seed, seedlen, mask, &masklen);
+   }
+   if (XSTRCMP(hash_descriptor[hash_idx].name, "shake256") == 0) {
+      return sha3_shake_memory(256, seed, seedlen, mask, &masklen);
+   }
```
sjaeckel:
```suggestion
   if (hash_descriptor[hash_idx].ID == 34
         || hash_descriptor[hash_idx].ID == 35) {
      return sha3_shake_memory(hash_descriptor[hash_idx].hashsize * 8, seed, seedlen, mask, &masklen);
   }
```

WDYT?

karel-m: IMO shake128 has hashsize 32 / shake256 has hashsize 64 (which migh be surprising but it is how RFC 8702 uses it)

perhaps:
```c
   if (hash_descriptor[hash_idx].ID == 34) { /* shake128 */
      return sha3_shake_memory(128, seed, seedlen, mask, &masklen);
   }
   if (hash_descriptor[hash_idx].ID == 35) { /* shake256 */
      return sha3_shake_memory(256, seed, seedlen, mask, &masklen);
   }
```

sjaeckel: I've found another (better?) approach, WDYT?

PR #755, opened 2026-05-04
Fix warnings

sjaeckel:
Pulled out from #751

### DIFF
--- a/src/misc/pem/pem_read.c
+++ b/src/misc/pem/pem_read.c
@@ -56,8 +56,11 @@ static LTC_INLINE int s_bufp_fits(struct bufp *buf, unsigned long to_write)
 {
    char *d = buf->work;
    char *e = buf->end;
-   char *w = d + to_write;
-   if (d == NULL || w < d || w > e)
+   char *w;
+   if (d == NULL || e == NULL)
+      return 0;
+   w = d + to_write;
+   if (w < d || w > e)
       return 0;
    return 1;
 }
--- a/src/misc/pem/pem_read.c
+++ b/src/misc/pem/pem_read.c
@@ -9,14 +9,6 @@
 
 #ifdef LTC_PEM
 
-extern const struct str pem_proc_type_encrypted;
-#ifdef LTC_SSH
-extern const struct str pem_ssh_comment;
-#endif
-extern const struct str pem_dek_info_start;
-extern const struct blockcipher_info pem_dek_infos[];
-extern const unsigned long pem_dek_infos_num;
-
 static LTC_INLINE unsigned long s_bufp_alloc_len(struct bufp *buf)
 {
    if (buf->start == NULL || buf->end == NULL)
--- a/src/misc/pem/pem_ssh.c
+++ b/src/misc/pem/pem_ssh.c
@@ -555,6 +555,7 @@ static int s_read_authorized_keys(const void *buf, unsigned long len, ssh_author
    }
    XMEMCPY(cpy, buf, len);
    s = cpy;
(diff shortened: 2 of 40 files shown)


src/misc/pem/pem_ssh.c
```diff
@@ -555,6 +555,7 @@ static int s_read_authorized_keys(const void *buf, unsigned long len, ssh_author
    }
    XMEMCPY(cpy, buf, len);
    s = cpy;
+   err = CRYPT_ERROR;
```
sjaeckel, line 558: Is this correct? Shouldn't this instead return `CRYPT_OK` if `len==0`?

src/misc/pem/pem_read.c
```diff
@@ -56,8 +48,11 @@ static LTC_INLINE int s_bufp_fits(struct bufp *buf, unsigned long to_write)
 {
    char *d = buf->work;
    char *e = buf->end;
-   char *w = d + to_write;
-   if (d == NULL || w < d || w > e)
+   char *w;
+   if (d == NULL || e == NULL)
+      return 0;
+   w = d + to_write;
+   if (w < d || w > e)
```
sjaeckel, line 55: FYI: I re-wrote this to the same form as before, since it is clearer and easier to understand IMO.

sjaeckel (commented):
IMO we should modify our builds to catch the parts that we want to as well.

We already run ubsan in the meta builds, why aren't those things caught there?

src/misc/pem/pem_ssh.c
karel-m: not sure as we do not accept zero `len` in `ssh_read_authorized_keys`

src/misc/pem/pem_read.c
karel-m: I am fine with that

karel-m:
Just FYI some options/flags I have used:

Sanitizers:
```
make clean
make -j$(nproc) CC=clang CFLAGS="-Wall -Wextra -I../libtommath -DUSE_LTM -DLTM_DESC -O1 -g3 -fno-omit-frame-pointer -fno-common -fsanitize=address,undefined -fsanitize-address-use-after-scope" LDFLAGS="-fsanitize=address,undefined" EXTRALIBS="-L../libtommath -ltommath" all
ASAN_OPTIONS=detect_leaks=1:abort_on_error=0:symbolize=1 UBSAN_OPTIONS=print_stacktrace=1:halt_on_error=0 ./test
```

Max compiler warnings:
```
make clean
make -j$(nproc) CC=clang CFLAGS="-Wall -Wextra -Wstrict-prototypes -Wmissing-prototypes -Wmissing-variable-declarations -Wshadow -Wformat=2 -Wno-format-nonliteral -Wunreachable-code -Wbad-function-cast -Wwrite-strings -Wpointer-arith -Wimplicit-fallthrough -Wformat-security -Wreturn-type -Wincompatible-pointer-types -Wimplicit-function-declaration -Wint-conversion -Wfloat-conversion -Wstrict-overflow -Wvla -Walloca -Wnull-dereference -Wredundant-decls -Wnested-externs -Wdeclaration-after-statement -Wmissing-field-initializers -Wold-style-definition -I../libtommath -DUSE_LTM -DLTM_DESC -O2" EXTRALIBS="-L../libtommath -ltommath" all
```

src/misc/pem/pem_ssh.c
sjaeckel: hmm, good question. Yet another case of `Operation failed successfully.` ...

PR #758, opened 2026-05-05
More fixes and improvements

### LINKED ISSUE #742: SHA1/SHA224/SHA256 state layouts are not memcpy-safe
In my Perl bindings for libtomcrypt (CryptX), I support cloning digest/MAC contexts.

For a long time, I treated hash_state / hmac_state as effectively copyable by value, but with the current SHA1 and SHA224/SHA256 state layouts this is no longer that simple.

The issue is that these states now contain an internal pointer (state) which points into aligned storage inside the same struct (state_buf). After a raw struct copy, the cloned object still points into the original object’s buffer.

One "solution" would be to remove the clone() method from my Perl bindings, but before doing that I wanted to ask whether there is any chance that these state structs will become memcpy-friendly again.

sjaeckel:
### Checklist

### DIFF
--- a/src/hashes/sha1.c
+++ b/src/hashes/sha1.c
@@ -180,25 +180,6 @@ static int s_sha1_c_compress(hash_state *md, const unsigned char *buf)
 #endif
 
 /**
-   Initialize the hash state
-   @param md   The hash state you wish to initialize
-   @return CRYPT_OK if successful
-*/
-int sha1_c_init(hash_state * md)
-{
-   LTC_ARGCHK(md != NULL);
-
-   md->sha1.state[0] = 0x67452301UL;
-   md->sha1.state[1] = 0xefcdab89UL;
-   md->sha1.state[2] = 0x98badcfeUL;
-   md->sha1.state[3] = 0x10325476UL;
-   md->sha1.state[4] = 0xc3d2e1f0UL;
-   md->sha1.curlen = 0;
-   md->sha1.length = 0;
-   return CRYPT_OK;
-}
-
-/**
    Process a block of memory though the hash
    @param md     The hash state
    @param in     The data to hash
--- a/src/hashes/sha1_x86.c
+++ b/src/hashes/sha1_x86.c
@@ -193,25 +193,6 @@ static int s_sha1_x86_compress(hash_state *md, const unsigned char *buf)
 #endif
 
 /**
-   Initialize the hash state
-   @param md   The hash state you wish to initialize
-   @return CRYPT_OK if successful
-*/
-int sha1_x86_init(hash_state * md)
-{
-   LTC_ARGCHK(md != NULL);
-
-   md->sha1.state[0] = 0x67452301UL;
-   md->sha1.state[1] = 0xefcdab89UL;
-   md->sha1.state[2] = 0x98badcfeUL;
-   md->sha1.state[3] = 0x10325476UL;
-   md->sha1.state[4] = 0xc3d2e1f0UL;
-   md->sha1.curlen = 0;
-   md->sha1.length = 0;
-   return CRYPT_OK;
-}
-
-/**
    Process a block of memory though the hash
    @param md     The hash state
    @param in     The data to hash
--- a/src/hashes/sha2/sha256.c
+++ b/src/hashes/sha2/sha256.c
@@ -20,7 +20,7 @@ const struct ltc_hash_descriptor sha256_portable_desc =
    { 2, 16, 840, 1, 101, 3, 4, 2, 1,  },
    9,
 
-    &sha256_c_init,
+    &sha256_init,
     &sha256_c_process,
     &sha256_c_done,
     &sha256_c_test,
@@ -225,28 +225,6 @@ static int s_sha256_compress(hash_state * md, const unsigned char *buf)
     return err;
 }
 #endif
-
-/**
-   Initialize the hash state
-   @param md   The hash state you wish to initialize
-   @return CRYPT_OK if successful
-*/
-int sha256_c_init(hash_state * md)
-{
-    LTC_ARGCHK(md != NULL);
-
-    md->sha256.curlen = 0;
-    md->sha256.length = 0;
-    md->sha256.state[0] = 0x6A09E667UL;
-    md->sha256.state[1] = 0xBB67AE85UL;
-    md->sha256.state[2] = 0x3C6EF372UL;
-    md->sha256.state[3] = 0xA54FF53AUL;
-    md->sha256.state[4] = 0x510E527FUL;
-    md->sha256.state[5] = 0x9B05688CUL;
-    md->sha256.state[6] = 0x1F83D9ABUL;
-    md->sha256.state[7] = 0x5BE0CD19UL;
-    return CRYPT_OK;
-}
 
 /**
    Process a block of memory though the hash
--- a/src/hashes/sha2/sha256_x86.c
+++ b/src/hashes/sha2/sha256_x86.c
@@ -250,10 +250,10 @@ static int LTC_SHA_TARGET s_sha256_x86_compress(hash_state * md, const unsigned 
 #undef K
 
 #ifdef LTC_CLEAN_STACK
-static int s_sha256_compress(hash_state * md, const unsigned char *buf)
+static int s_sha256_x86_compress(hash_state * md, const unsigned char *buf)
 {
     int err;
-    err = ss_sha256_compress(md, buf);
+    err = ss_sha256_x86_compress(md, buf);
     burn_stack(sizeof(ulong32) * 74);
     return err;
 }
--- a/src/hashes/sha2/sha256_x86.c
+++ b/src/hashes/sha2/sha256_x86.c
@@ -38,7 +38,7 @@ const struct ltc_hash_descriptor sha256_x86_desc =
    { 2, 16, 840, 1, 101, 3, 4, 2, 1,  },
    9,
 
-    &sha256_x86_init,
+    &sha256_init,
     &sha256_x86_process,
     &sha256_x86_done,
     &sha256_x86_test,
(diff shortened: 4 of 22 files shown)


karel-m:
Do we actually still need sha1_x86_init and sha1_c_init when they are both equal to sha1_init?

sjaeckel:
> Do we actually still need sha1_x86_init and sha1_c_init when they are both equal to sha1_init?

Nope, they could be defined to `sha1_init`! Will do that :) Thanks!

karel-m:
similarly sha224_x86_init / sha256_x86_init vs sha224_c_init / sha256_c_init

src/hashes/sha2/sha256.c
```diff
@@ -226,38 +226,29 @@ static int s_sha256_compress(hash_state * md, const unsigned char *buf)
...
    @return CRYPT_OK if successful
 */
-HASH_PROCESS(sha256_c_process,s_sha256_compress, sha256, 64)
+static HASH_PROCESS(s_sha256_c_process,s_sha256_compress, sha256, 64)
+
+static LTC_INLINE void s_sha256_c_fixup_state(hash_state * md)
+{
+   ulong32 *state = LTC_ALIGN_BUF(md->sha256.state_buf, 16);
+   if (state != md->sha256.state) {
+      XMEMMOVE(state, md->sha256.state, 8 * sizeof(ulong32));
```
MarekKnapek: I'm not sure this is safe. What if the user  memcpy the state to a new place. Then they destroys the old state. Think of free(). Then they call process or finish function. The memmove would access already freed memory.

sjaeckel: I've pushed an update.

MarekKnapek (commented):
Maybe also unify the `*_done` functions. As there is no `x86`specific logic inside them. They do call the `*_process` functions which might get dispatched to portable or `x86` version, but themselves do not contain any added smart-ness over the portable version.

tests/cipher_hash_test.c
```diff
@@ -4,6 +4,54 @@
...
+   return CRYPT_OK;
+}
+
+static int unaligned_hash_md_tests(void)
+{
+   unsigned char md_buf[sizeof(hash_state) + 16];
+   unsigned long n;
+   hash_state md;
+   for (n = 0; n < 16; ++n) {
+      hash_state *md2 = (void*)&md_buf[n];
```
MarekKnapek: This smells like undefined behavior to me. The `struct hash_state` or more precisely `union hash_state` has some data members (such as `int` or `ulong32` or `ulong64` or `void*`) that have alignment requirement other than `1`. By this cast you break this alignment contract. This is not allowed in C and is considered undefined behavior.

This is the reason why I created a new data type `ugly_struct` in my https://github.com/libtom/libtomcrypt/pull/761. I could create two variants, one is:

```
struct ugly_struct
{
    char padding[x];
    hash_state state;
};
```

the other is:

```
struct ugly_struct
{
    char padding[x];
    sha1_state state;
};
```

I choose the second one. Because other members of the `union` could unintentionally increase the alignment.

The difference between my approach and your approach is that in your pursuit of un-alignment-ness of hash state, you also break the alignment of fundamental data types. In my pursuit of un-alignment-ness of hash state, I do not break the alignment of fundamental data types, only the hash state as whole.

src/hashes/sha1.c
```diff
@@ -186,7 +186,23 @@ static int s_sha1_c_compress(hash_state *md, const unsigned char *buf)
...
-HASH_PROCESS(sha1_c_process, s_sha1_c_compress, sha1, 64)
+static HASH_PROCESS(s_sha1_c_process, s_sha1_c_compress, sha1, 64)
+
+static LTC_INLINE void s_sha1_c_fixup_state(hash_state * md)
+{
+   ulong32 *state = LTC_ALIGN_BUF(md->sha1.state_buf, 16);
+   if (state != md->sha1.state) {
+      XMEMMOVE(state, md->sha1.state_buf + md->sha1.align, 5 * sizeof(ulong32));
+      md->sha1.state = state;
+      md->sha1.align = md->sha1.state_buf - (unsigned char*)state;
```
MarekKnapek: This subtraction seems to be backwards. The `state` variable should point to higher memory address than the `state_buf` variable.

src/hashes/sha1_x86.c
```diff
@@ -198,7 +198,23 @@ static int s_sha1_x86_compress(hash_state *md, const unsigned char *buf)
...
-HASH_PROCESS(sha1_x86_process, s_sha1_x86_compress, sha1, 64)
+static HASH_PROCESS(s_sha1_x86_process, s_sha1_x86_compress, sha1, 64)
+
+static LTC_INLINE void s_sha1_x86_fixup_state(hash_state * md)
+{
+   ulong32 *state = LTC_ALIGN_BUF(md->sha1.state_buf, 16);
+   if (state != md->sha1.state) {
+      XMEMMOVE(state, md->sha1.state_buf + md->sha1.align, 5 * sizeof(ulong32));
+      md->sha1.state = state;
+      md->sha1.align = md->sha1.state_buf - (unsigned char*)state;
```
MarekKnapek: This subtraction seems to be backwards. The `state` variable should point to higher memory address than the `state_buf` variable.

src/hashes/sha2/sha256.c
```diff
@@ -233,7 +233,23 @@ static int s_sha256_compress(hash_state * md, const unsigned char *buf)
...
-HASH_PROCESS(sha256_c_process,s_sha256_compress, sha256, 64)
+static HASH_PROCESS(s_sha256_c_process,s_sha256_compress, sha256, 64)
+
+static LTC_INLINE void s_sha256_c_fixup_state(hash_state * md)
+{
+   ulong32 *state = LTC_ALIGN_BUF(md->sha256.state_buf, 16);
+   if (state != md->sha256.state) {
+      XMEMMOVE(state, md->sha256.state_buf + md->sha256.align, 8 * sizeof(ulong32));
+      md->sha256.state = state;
+      md->sha256.align = md->sha256.state_buf - (unsigned char*)state;
```
MarekKnapek: This subtraction seems to be backwards. The `state` variable should point to higher memory address than the `state_buf` variable.

src/hashes/sha2/sha256_x86.c
```diff
@@ -265,7 +265,23 @@ static int s_sha256_compress(hash_state * md, const unsigned char *buf)
...
-HASH_PROCESS(sha256_x86_process,s_sha256_x86_compress, sha256, 64)
+static HASH_PROCESS(s_sha256_x86_process,s_sha256_x86_compress, sha256, 64)
+
+static LTC_INLINE void s_sha256_x86_fixup_state(hash_state * md)
+{
+   ulong32 *state = LTC_ALIGN_BUF(md->sha256.state_buf, 16);
+   if (state != md->sha256.state) {
+      XMEMMOVE(state, md->sha256.state_buf + md->sha256.align, 8 * sizeof(ulong32));
+      md->sha256.state = state;
+      md->sha256.align = md->sha256.state_buf - (unsigned char*)state;
```
MarekKnapek: This subtraction seems to be backwards. The `state` variable should point to higher memory address than the `state_buf` variable.

tests/cipher_hash_test.c
MarekKnapek: Build with UBSan and get this result. `tests/cipher_hash_test.c:17:12: runtime error: store to misaligned address 0x7ffcec3d0b71 for type 'hash_state *' (aka 'union Hash_state *'), which requires 8 byte alignment`

sjaeckel: > This smells like undefined behavior to me.

Absolutely.

> This is the reason why I created a new data type `ugly_struct`

Unfortunately this won't work to demo the fixup of misaligned buffers, that's why I implemented it as I did.

```
$ pahole -C ugly_struct demos/small.o                                                                           
struct ugly_struct {
        char                       padding[1];           /*     0     1 */

        /* XXX 7 bytes hole, try to pack */

        hash_state                 state;                /*     8   856 */

        /* size: 864, cachelines: 14, members: 2 */
        /* sum members: 857, holes: 1, sum holes: 7 */
        /* last cacheline: 32 bytes */
};
```

We could pack the struct

```
$ pahole -C ugly_struct2 demos/small.o                                   
struct ugly_struct2 {                                                                                                             
        char                       padding[1];           /*     0     1 */         
        hash_state                 state;                /*     1   856 */                                                        
                                                                 
        /* size: 857, cachelines: 14, members: 2 */                                                                               
        /* last cacheline: 25 bytes */                 
} __attribute__((__packed__));      
```

but this would then also lead to

```
warning: taking address of packed member 'state' of class or structure 'ugly_struct2' may result in an unaligned pointer value [-Waddress-of-packed-member]
```

There was https://github.com/DCIT/perl-CryptX/issues/95 which lead to introducing the dynamic aligning of buffers, so even though we could in theory even handle completely unaligned accesses we should maybe only test multiples of 8 which should in turn not produce UB. Right?

src/hashes/sha2/sha256.c
sjaeckel: 🤦 I really shouldn't do multiple things at once ...

sjaeckel: I've also included this in the testcases.

tests/cipher_hash_test.c
MarekKnapek: > Unfortunately this won't work to demo the fixup of misaligned buffers

I do not understand. The `ugly_struct` will definitely demo the misalignment of the **state** variables. If the **state** needs to be aligned more strictly than any other fundamental type. Such as to 16 bytes (for SHA-1 and SHA-256), and largest fundamental type is `ulong64` which is not enough to be aligned to 16 bytes.

The `padding[1]` will shift the structure by 8 bytes, or maybe by 4 bytes on 32bi platforms. There was never an intend to mis-align the structure by 1 byte by me. As this is impossible without invoking undefined behavior.

If you are talking about aligning **buffers** instead of **state**, this is different. I never intended to align the data buffers and always used the unaligned access in my intrinsics (except briefly, but  I force pushed the change back).

The alignment requirement seems to bring more troubles than benefit. There is always an option to not require any alignment (besides the fundamental C one) and use different intrinsics to read from and write to the state variables.

karel-m:
Can we move the changes related to "... state `memcpy()`'able again" out of this PR?

They are IMO not easy, whereas the rest of this PR are mostly non-controversial cleanups/improvements.

sjaeckel (commented):
Looks like this can be merged!?

karel-m:
> Looks like this can be merged!?

Yes, I have removed stuff related to "... state memcpy()'able again"

karel-m (approved):

PR #760, opened 2026-05-16
ARIA block cipher (RFC 5794)

karel-m:
Added ARIA support as defined in [RFC 5794](https://www.rfc-editor.org/rfc/rfc5794.html).

Unoptimized implementation based on the pseudocode from RFC (using libtomcrypt macros and other pieces where possible).

### Checklist


sjaeckel (approved):
Should we maybe also add it to the list of supported PKCS#8 algos?

PR #761, opened 2026-05-18
Hash state test

MarekKnapek:
### Checklist


MarekKnapek:
Anyway, there is already better PR on its way. It uses a byte data type to track the align offset instead of int. It might be better. This PR is for scrap.

This PR might serve as an inspiration for testing un-aligned hash state.

karel-m:
solved in #785

PR #762, opened 2026-05-18
Hash memcpy

MarekKnapek:
### Checklist


src/hashes/sha2/sha256_x86.c
karel-m, line 361: this needs to be changed to:
```c
STORE32H(state[i], out+(4*i));
```

MarekKnapek: Wow, this should not even compile as I removed the state pointer from the struct. It is redundant now, because it could be computed at any time by using the byte buffer and align variable.

Anyway, there is already better PR on its way. It uses a `byte` data type to track the align offset instead of `int`. It might be better. This PR is for scrap.

BUT the state pointer remains in that PR. Should it? If it is possible to compute it at any time? Isn't it redundant now?

karel-m:
solved in #785

PR #763, opened 2026-05-20
ubsan

MarekKnapek:
### Checklist

### DIFF
--- a/src/encauth/ccm/ccm_memory.c
+++ b/src/encauth/ccm/ccm_memory.c
@@ -253,8 +253,8 @@ int ccm_memory(int cipher,
 
                 /* xor the PT against the pad first */
                 for (z = 0; z < 16; z += sizeof(LTC_FAST_TYPE)) {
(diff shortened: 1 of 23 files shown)


karel-m:
@MarekKnapek how is this PR different from https://github.com/libtom/libtomcrypt/pull/756 ?

MarekKnapek:
Eh, when talking about undefined behavior in https://github.com/libtom/libtomcrypt/pull/758 I decided to compile all of libtomcrypt with UBSan and fix all places it triggered during unit tests. Both PRs seem to be created roughly at the same time. This PR might contain more changes as it not only fixes the unaligned integer casts, but also some other things UBSan detected. Such as memcpy on null pointer and zero length. Bit-shift on singed negative variable.

sjaeckel (approved):
Thanks a lot for the PR.

I took the liberty to squash the commits together.

I also took your macro based version instead of the inline functions proposed in #756 since we had problems with inline functions on some compilers ages ago. @karel-m IIRC some IBM compiler chugged on this in the past, can you maybe check?

Also I kept the ed{25519,448} modifications from @karel-m since I liked the style better.

PR #769, opened 2026-06-05
Fix x86 ClangCL and GCC/Clang `-std=c##` compiles

### LINKED ISSUE #767: x86 CPUID compile errors on Windows with Clang and GCC compilers
### Prerequisites

* [X] Checked the developer manual

### Description

There are some compilation errors relating to recently added x86 CPUID detection code.

#### ClangCL (Clang 20.1.8):
- Looks like `<intrin.h>` should be included in files that call `__cpuid`. (I guess MSVC magically recognizes `__cpuid` without a header, but ClangCL needs the header.)

```
<...>\cpp\lib\libtomcrypt\src\hashes\sha2\sha256_desc.c(12,4): error : call to undeclared library function '__cpuid' with type 'void (int *, int)'; <...>
<...>\cpp\lib\libtomcrypt\src\hashes\sha2\sha256_desc.c(12,4): message : include the header <intrin.h> or explicitly provide a declaration for '__cpuid' <...>
```

#### MinGW-UCRT64 (GCC 15.2.0) / MinGW-Clang64 (Clang 21.1.7):
- Looks like `asm` should be changed to `__asm__`, to allow building without GNU C extensions.
- I wonder if this issue also manifests on other x86 targets (e.g. Linux) that compile with `-std=c##`.

```
<...>/cpp/lib/libtomcrypt/src/hashes/sha2/sha256_desc.c:18:5: error: use of undeclared identifier 'asm'
   18 |     asm volatile ("cpuid"
      |     ^~~
```

I am open to submitting a fix if desired, or an existing contributor could do so. Thanks for the recent work on adding x86 acceleration to the library.

### Version

a68fa19bc2b532f66a6f18ca457daec53054a312 (top of develop tree May 19)

p0lymeric:
### Checklist

* [x] if this fixes something: added a `Fixes:` tag to the commit message

Fixes compile for some x86 compiler configurations, including ClangCL on Windows, and GCC/Clang in Standard C mode (-std=c##).

Fixes https://github.com/libtom/libtomcrypt/issues/767

Notes:
- Reintroduced some nasty-looking parts from this past commit (aes_desc.c:L66-78):
  - https://github.com/libtom/libtomcrypt/commit/ec3804c452acfa794306b092d2209ab7e1d90ba5#diff-4b49c40dd5787d702d14e1859464a5941919a8dd8288dce824014f07939e321eL66-L78
- ClangCL is handled by falling through `defined(_MSC_VER) && !defined(__clang__)` into the GNU-style inline asm block
- `__asm__` instead of `asm` so that gcc/clang accept the inline asm block in Standard C mode

### DIFF
--- a/src/ciphers/aes/aes_desc.c
+++ b/src/ciphers/aes/aes_desc.c
@@ -15,7 +15,7 @@
 #define LTC_S_X86_CPUID
 static LTC_INLINE void s_x86_cpuid(int* regs, int leaf)
 {
-#if defined _MSC_VER
+#if defined(_MSC_VER) && !defined(__clang__) // MSVC but not ClangCL
    __cpuid(regs, leaf);
 #else
    int a, b, c, d;
--- a/src/encauth/gcm/gcm_gf_mult.c
+++ b/src/encauth/gcm/gcm_gf_mult.c
@@ -33,7 +33,7 @@
 #define LTC_S_X86_CPUID
 static LTC_INLINE void s_x86_cpuid(int* regs, int leaf)
 {
-#if defined _MSC_VER
+#if defined(_MSC_VER) && !defined(__clang__) // MSVC but not ClangCL
    __cpuid(regs, leaf);
 #else
    int a, b, c, d;
--- a/src/hashes/sha1_desc.c
+++ b/src/hashes/sha1_desc.c
@@ -28,7 +28,7 @@ const struct ltc_hash_descriptor sha1_desc =
 #define LTC_S_X86_CPUID
 static LTC_INLINE void s_x86_cpuid(int* regs, int leaf)
 {
-#if defined _MSC_VER
+#if defined(_MSC_VER) && !defined(__clang__) // MSVC but not ClangCL
    __cpuid(regs, leaf);
 #else
     int a, b, c, d;
--- a/src/hashes/sha2/sha224_desc.c
+++ b/src/hashes/sha2/sha224_desc.c
@@ -33,7 +33,7 @@ const struct ltc_hash_descriptor sha224_desc =
 #define LTC_S_X86_CPUID
 static LTC_INLINE void s_x86_cpuid(int* regs, int leaf)
 {
-#if defined _MSC_VER
+#if defined(_MSC_VER) && !defined(__clang__) // MSVC but not ClangCL
    __cpuid(regs, leaf);
 #else
     int a, b, c, d;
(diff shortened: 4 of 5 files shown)


karel-m:
there already exists a similar patch https://github.com/libtom/libtomcrypt/pull/755/changes/e9a7e949c6dea527ad20f46fa42cdbd53c071b0e in #755

p0lymeric:
I agree, https://github.com/libtom/libtomcrypt/commit/e9a7e949c6dea527ad20f46fa42cdbd53c071b0e covers the -std=c## case. Thank you, did not see that patch.

This PR also has a fix for ClangCL compilation. A previous cleanup commit (https://github.com/libtom/libtomcrypt/commit/ec3804c452acfa794306b092d2209ab7e1d90ba5) removed the ```&& !defined(__clang__)``` factor from the ```#if defined(_MSC_VER)``` check, in the function that became ```s_x86_cpuid```.

karel-m:
@p0lymeric FYI the `__asm__` part is now fixed in develop

I am not sure about the ugly hack `#if defined(_MSC_VER) && !defined(__clang__) ...`

Isn't it about missing `#include <intrin.h>` ?

I am not using ClangCL but if it tries to pretend to be GNUC and MSVC at once then this might be a problem:
```
#if defined(__GNUC__)
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wcast-align"
#pragma GCC diagnostic ignored "-Wdeclaration-after-statement"
#pragma GCC diagnostic ignored "-Wuninitialized"
#pragma GCC diagnostic ignored "-Wunused-function"
#elif defined(_MSC_VER)
#include <intrin.h>
#endif
```

karel-m:
I am still not sure (I do not have clang-cl) whether the correct fix should instead look like this:
```diff
  #if defined(__GNUC__)
  #pragma GCC diagnostic push
  #pragma GCC diagnostic ignored "-Wcast-align"
  #pragma GCC diagnostic ignored "-Wdeclaration-after-statement"
  #pragma GCC diagnostic ignored "-Wuninitialized"
  #pragma GCC diagnostic ignored "-Wunused-function"
-  #elif defined(_MSC_VER)
+ #endif
+ #if defined(_MSC_VER)
  #include <intrin.h>
  #endif
```

sjaeckel:
> I do not have clang-cl

Maybe someone can adapt the appveyor CI config to also build against clang-cl?

https://clang.llvm.org/docs/UsersManual.html#clang-cl

karel-m:
see #786

sjaeckel:
Superseded by #786 

Please feel free to re-open if there's something missing.

p0lymeric:
Thank you! Yes the <intrin.h> include looks good to me.

PR #777, opened 2026-07-18
SIV properly handle zero AD components

karel-m:
This PR fixes an SIV corner case that occurs when zero AD components are used.

### Checklist

### DIFF
--- a/doc/crypt.tex
+++ b/doc/crypt.tex
@@ -2556,6 +2556,7 @@ To encrypt and create a tag resp. decrypt and check the tag, the following API f
 \begin{verbatim}
 int siv_encrypt_memory(                int  cipher,
                        const unsigned char *key,    unsigned long  keylen,
+                             unsigned long adnum,
                        const unsigned char *ad[],   unsigned long  adlen[],
                        const unsigned char *pt,     unsigned long  ptlen,
                              unsigned char *ct,     unsigned long *ctlen);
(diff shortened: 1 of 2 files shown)


sjaeckel (commented):
This has been discussed on the IETF ML [0] and FMU it was unconclusive.

Now I'm not sure if merging this would make our implementation worse, since we already allow a zero-length nonce...

[0] https://mailarchive.ietf.org/arch/msg/cfrg/Gu5cyorHPLrQ_CGYe775Xc6SV1c/

karel-m:
I did not look into it that deeply. I mainly wanted to be compatible with OpenSSL SIV implementation.

You are right, it may need more investigation.

karel-m:
I investigated this further and the main problem seems to be that with SIV there are two kinds of "empty associated data":
* zero AD components (no AD at all)
* one empty AD component

I found that empty AD was also an issue in OpenSSL before versions 3.0.9/3.1.1 - see: https://openssl-library.org/news/secadv/20230714.txt

I updated `siv_test()` so that it now checks both scenarios "zero AD components" + "one empty AD component" and verifies that libtomcrypt behaves in the same way as OpenSSL 3.5.6.

So the main idea of this PR is: let us stick with what the "big boys" in the OpenSSL project are doing.

sjaeckel (approved):

PR #785, opened 2026-07-30
Another attempt to make sha1/sha256 states memcpyable

karel-m:
The main idea:
* using `_mm_loadu_si128` instead of `_mm_load_si128`
* using `_mm_storeu_si128` instead of `_mm_store_si128`

However, it looks soooo simple that I am not sure whether I have overlooked something.

### DIFF
--- a/src/hashes/sha1.c
+++ b/src/hashes/sha1.c
@@ -188,8 +188,6 @@ int sha1_c_init(hash_state * md)
 {
    LTC_ARGCHK(md != NULL);
 
-   md->sha1.state = LTC_ALIGN_BUF(md->sha1.state_buf, 16);
-
    md->sha1.state[0] = 0x67452301UL;
    md->sha1.state[1] = 0xefcdab89UL;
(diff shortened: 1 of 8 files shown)


karel-m:
ping @MarekKnapek

MarekKnapek:
Yes, looks good, as I mentioned previously in https://github.com/libtom/libtomcrypt/issues/742#issuecomment-4391360479 replacing the aligned mem-loads with unaligned mem-loads should be enough.

Currently, there are still aligned mem-loads present. Not when reading or storing the hash state, but when reading the constants table in SHA-256. The constants table is heavily decorated with macros forcing it to be aligned to 16 bytes. In old C, this is not possible to achieve via any standards means, only with compiler specific means. In newer C, it is possible to achieve this via standards means. So maybe the hash state could be macro decorated also? I believe that if you have a compiler capable of using intrinsics, the same compiler should be able to align a struct stricter than largest fundamental data type. I suggested this before, but it was rejected.

sjaeckel (approved):

PR #786, opened 2026-07-31
appveyor + clang-cl

karel-m:
Just testing


sjaeckel:
You can also push to a separate branch appveyor/foo so GH actions won't be triggered.

karel-m:
the `clang-cl` build now passes

